PR #192, opened 2013-07-15
Preserve GUI state across page loads

daprice:
This saves the current tab and the open/closed state of each accordion section to localStorage so they remain the same when the page is reloaded.

Along with the single column mobile layout I'm working on, this could be especially helpful to mobile users since iOS often reloads the page when the browser is in the background; keeping the accordions closed (when the user closes them) saves lots of vertical space.


foosel:
I know, it's ridiculous to come up with this now... I think at this point this would be better done by a plugin. My concerns with a general inclusion of something like this still apply. People already get confused/bitten by the implementation of local storage for the filter/sorting settings on the file lists.

Sorry for closing & even more so for letting this collect dust for so long.

PR #300, opened 2013-11-02
octoprint/util/comm.py: make line number detection more robust

srwalter:
When the printer requests a resend, we have to determine which line
number it wants.  The old logic would be confused if there was more
than one number in the line.  Some version of Teacup firmware
request a resend with the format "rs N25 Expected checksum 109".
The new logic detects this format correctly.


foosel:
Mentioned regex based approach implemented in the above commit.

Issue #31, opened 2013-01-29
Folders in GUI


hughht5:
Any chance you can implement folders for files?

Use case: Printing a prusa i2 takes 6 full plates for me. I'd like to upload prusa i2 folder easily then choose which plate to print.

dragondgold:
Any updates on this? :)

foosel:
In fact yes, @Salandora implemented folder support and it has been merged on devel since last week I think. He's also turning the file manager into a plugin. So I guess this can be closed :)

foosel:
And some screenshots (only tablet atm, sorry)

![screenshot_2015-11-27-20-15-34](https://cloud.githubusercontent.com/assets/83657/11447273/db0c841a-9543-11e5-994f-40531645ab26.png)

![screenshot_2015-11-27-20-16-00](https://cloud.githubusercontent.com/assets/83657/11447282/e3b26512-9543-11e5-9c27-089955287204.png)

And now a round of applause for @Salandora please, that's an awesome job well done :)

dragondgold:
Great! Thank you so much and thank you @Salandora it's working great and a very useful feature!

Salandora:
@dragondgold I guess you will even more like the fileManager plugin when released.
Because you will have the ability to copy, move and rename files and folders then. And also a litte nice feature "upload to Sd" from an already uploaded file ;-)
I hope to get it released this weekend.

dragondgold:
@Salandora that sounds greeeat! Looking forward for your plugin!

PR #387, opened 2014-02-20
File manager

Salandora:
Request number: #31

Added folder support and a little filemanager to better organize files and folders
had to do many changes to api/files.py and gcodefiles.py

Still have todo changes to filtering, sorting and pagination.
Status: 
   Filtering: Only SD/local but that's also given through folders
   Sorting: Only by directory Name
   Pagination: No


Salandora:
Closing because becomes an plugin

Issue #324, opened 2013-12-05
Add handling of @pause in gcode files


Cylindric:
Hi, it would be useful for us single-extruder people to be able to pause the print at a precise position. Would it be possible to implement the @pause command similar to how Repetier does?

Basically if Octoprint sees @pause in the gcode, it enters the same pause state as if the pause button was pressed.

thanks

guysoft:
Is this still the case?
Doesn't this make this plugin useless?
https://github.com/MoonshineSG/OctoPrint-MultiColors

foosel:
Just noticed that I forgot to close this when 1.3.7 was released, which added native support for `@pause` [and some other commands](http://docs.octoprint.org/en/master/features/atcommands.html).

PR #574, opened 2014-09-04
Add handling of @pause in gcode files #324

dattas:
This pull request is to implement a basic version of #324 . It will automatically pause the print within Octoprint if @pause is encountered as a command in the gcode file.


foosel:
I guess we can close this now that you are doing this via a plugin, right?

PR #591, opened 2014-09-24
Generic Settings Loader

cdaringe:
# Generic Settings Loader
## Background

I have two features in development, separate from this PR, that involve accessing OctoPrint settings.  I feel that the settings protocol is a little clunky in current state, and disliked that the settings must be explicitly and redundantly called out in the code.  This PR serves to make getting/setting of settings simpler exclusively for the developer.  I foolishly branched from foosel/master instead of foosel/devel, so I'd like to request to do the PR now, then after issues are settled, perform the conflict resolution.  This is a **large** PR, so I would be happy to annotate it heavily to help the review process OR schedule time to review together online.
### Server Side
1. Made the get/set methods **generic** in the Settings class and the Settings API.  I.e., just call `get()` vs. `getInt/getBool/etc`.  Uses new function `coercePrimative()` to infer native type when passed js stringified IO.  Additionally this permits us to reduce the number of times all of the settings are explicity laid out in each file (from 4x => 2x [1x for server 1x for client].  Perhaps a package build process can reduce this need further as Python Dictionary and js Object syntax are ~equivalent.
2. Server ajax set/get "all" functions now use a generic setting **iterator**, vs. explicity calling out each setting.
3. To facilitate the above, the default settings are intially broken into two dictionaries. 1) settings usable and exposed to the client, then 2) another dictionary that is not exposed to the client.  They are zipped back into a single default_settings variable on init.  The ajax getters/setters determine which properties to send by looking at the client dictionary.
### Client Side
1. Similair to server, I have listed settings for client only once in settings.js, then **auto-generate ko observables** (commonly listed as `koos`).
   _This means you must only add your new setting to settings.py and settings.js a single time per file, and your new setting is accessible on server and client!_
### Helpers

`eachDeep()` and `dict_merge()` were introduced in collectionUtils.py as a means to recurse through the settings and to zip dictionaries together, respectively.
## Updated folder setting loading scheme

Because folder get/set functions are reliant on getBaseFolder/setBaseFolder (file IO), the the batch get/set of the ajax requests couldn't be handled without special cases.
To permit using the generic functions and still update the folders properly:
1. Auto create the default folders on load via setUserDirs()
2. After the ajax set is executed, run setUserDirs() again to build the new dir immediately.

```
Note--the *pre-existing* upload feature has a bug wherein if you change the upload dir, you must restart the server for the change to take place.  I will have a follow up PR to resolve this.
```
## General
1. Linted js -- fixed syntax errors (missing ;'s, == vs. ===, etc)
2. Alphabetized all settings dictionaries (server) & objects (client) for consistency
3. Alphabetized some file's member functions for readability (settings.js)
4. **Unified setting names between client and server**


mic159:
I think this is quite overdue.
Currently the settings are quite cumbersome when it shouldn't need to be.
What's the holdup on it?

foosel:
@mic159 Priorities. I wanted to get the new version out of the door (RC1 released yesterday) that with the plugin system already includes a lot of big changes to the source and didn't have the resources to properly think about let alone merge yet another major refactoring.

foosel:
I admittedly completely dropped the ball on that one 😕 It simply never felt the right time to tackle this given it's huge size and the reliance of third party plugins on the existing also makes changes extremely tricky now. Considering the repository is done now and the sources have diverged faaaar I'm closing this.

I want to revamp the settings some time down the road, but I have a ton of other more pressing things to handle first so consider this something that's not going to be done until at least 1.5.0 (or whatever comes after 1.4.0).

PR #663, opened 2014-12-05
Relative path for webcam stream

Bevaz:
If Octoprint server and webcam stream are hosted on the same server there could be a problem
when access is possible from different networks: for example from local network and from internet.

Let assume that octoprint address from local net is **http://192.168.111.22:5000** and from internet **http://octoprint.somedomain.org**. Webcam stream is hosted on the same server on port **4321** and path **/video**.
In this case when you connect to **http://192.168.111.22:5000** the address of webcam stream should be **http://192.168.111.22:4321/video**. When connect to **http://octoprint.somedomain.org** the address of webcam stream should be **http://octoprint.somedomain.org:4321/video**

Now you can specify **:port/path** as a webcam stream address (**:4321/video** in example above) and it will be added to server address depending on how you access to the server. Port with path are removed from octoprint address and port with path for webcam stream are added.

### DIFF
--- a/src/octoprint/server/__init__.py
+++ b/src/octoprint/server/__init__.py
@@ -13,6 +13,7 @@ from flask.ext.principal import Principal, Permission, RoleNeed, identity_loaded
 from flask.ext.babel import Babel
 from babel import Locale
 from watchdog.observers import Observer
+from urlparse import urlparse
 
 import os
 import logging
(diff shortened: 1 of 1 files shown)


foosel:
Closing. As outlined above, only works in a very limited scenario that would only fit a very small fraction of the user base. And that specific use case would probably be way better covered by a plugin basically replacing the webcam image or even just the URL at runtime.

PR #673, opened 2014-12-10
BugFix to Reconnect on Debian

fjriosp:
Solves the problems on reconnect using pyserial on Debian (I don't know if also in another distros).

Source:
http://forum.arduino.cc/index.php?topic=91291.0


guysoft:
Hey,
I tried:
1. apply this to the source at OctoPi 0.10.0
2. rebooted
3. conneted to the printer
4. Turned the printer off and got serial exception
5. turn the printer on
6. press connect button>

Expected:
Should reconnect and show as "operational"

Actual:
Still hangs on "Connecting".

Does this mean it does not work?

PR #712, opened 2015-01-10
Daemon cleans up pidfile when receiving a SIGTERM

### LINKED ISSUE #711: Daemonized OctoPrint doesn't clean up its pidfile when receiving a TERM signal
1. What were you doing?

  Writing a rc.d script for FreeBSD.

2. What did you expect to happen?

  OctoPrint would clean up its pidfile after a graceful exit.

3. What happened instead?

  The pidfile remained in place. This caused an "already running" error to occur when I attempted to run the daemon again.

4. Branch & Commit or Version:

  devel 3d5fdf2a917833808f132212b76ad3c6c5768419

5. Printer model & used firmware incl. version

  Not relevant.

6. Link to octoprint.log on gist.github.com or pastebin.com:

  Not relevant.

7. Link to contents of terminal tab on gist.github.com or pastebin.com:

  https://gist.github.com/rspeed/d37acd6fc447eac1cb3e

# Notes

The base issue is that `daemon.py` uses `atexit`, which states in [its documentation](https://docs.python.org/2/library/atexit.html):

> The functions registered via this module are not called when the program is killed by a signal not handled by Python

Since SIGTERM is the standard POSIX way to ask a daemon to quit, and since daemons are supposed to clean up after themselves, this design requires unusual and fragile behavior in init scripts. Specifically, I noticed that the init script has this line:

> rm -f $PIDFILE

That's generally considered to be a bad practice.

rspeed:
Changes the behavior of a daemonized OctoPrint to delete its pidfile when receiving a TERM signal. Current behavior would only delete the file upon exit of the separate process that was started to kill the daemon process (using the `--daemon stop` flag).

Compatibility with the existing command line arguments is retained because it was already sending SIGTERM to the daemon process.

Should fix #711.

Note: I didn't touch `scripts/octoprint.init` because I'm not very familiar with the behavior of Debian's init.d, and have limited capability to test any changes.

### DIFF
--- a/src/octoprint/daemon.py
+++ b/src/octoprint/daemon.py
@@ -53,13 +53,14 @@ class Daemon:
 		os.dup2(se.fileno(), sys.stderr.fileno())
 	
 		# write pidfile
-		signal.signal(signal.SIGTERM,self.delpid)
-
 		pid = str(os.getpid())
 		with open(self.pidfile,'w+') as f:
 			f.write(pid + '\n')
-	
-	def delpid(self, _signo, _stack_frame):
+
+		# register listener for SIGTERM
+		signal.signal(signal.SIGTERM, self.term)
+
+	def term(self, _signo, _stack_frame):
 		os.remove(self.pidfile)
 		sys.exit(0)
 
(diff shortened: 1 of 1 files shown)


foosel:
The init script actually uses `start-stop-daemon` and not the built-in daemonization (which I btw just forgot and caused me scratching my head quite a bit when testing this), so everything green, thanks!

rspeed:
I have to assume `start-stop-daemon` sends SIGTERM when shutting down the process, so the init script probably needs the `rm` line removed (and possibly other changes).

Issue #710, opened 2015-01-09
Invalid cross-device link on Linux


dattas:
1. What were you doing?
Uploading a .stl file on a linux filesystem which has a separate /tmp filesystem, this appears to only occur in Centos and Archlinux and not the *buntu and *bian systems
2. What did you expect to happen?
File would show up in the file manager so that it can be edited later
3. What happened instead?
The slicing UI comes up, but if you dismiss it or go to try to slice the .stl file, it doesn't show up in the file manager. File does exist on filesystem in the correct location.
4. Branch & Commit or Version:
Version: 1.2.0-dev-383-g3d5fdf2 (devel branch)
5. Printer model & used firmware incl. version
N/A
6. Link to octoprint.log on gist.github.com or pastebin.com:
https://gist.github.com/dattas/2d5ec98b0ef936604036
7. Link to contents of terminal tab on gist.github.com or pastebin.com:
N/A

I love cookies and so I think perhaps there should be a tmp directory in the .octoprint directory instead of writing the temp files in /tmp. This should prevent most of the Cross-device link errors

Renha:
maybe shutil.move must be used instead of os.rename
os.rename should be ok for ones who had /tmp on same FS

derpicknicker1:
Pls close

Issue #716, opened 2015-01-13
[Support] os.rename throws error (/devel)


derpicknicker1:
In file `/src/octoprint/util/__init__.py` in function _safeRename_ line 185
```python
os.rename(old, new)
```

throwns the following Error:

```python
2015-01-13 12:25:31,831 - octoprint.filemanager.storage - ERROR - Error while writing .metadata.yaml to /home/root/.octoprint/uploads
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/OctoPrint-1.2.0_dev_395_g4e4ad0b-py2.7.egg/octoprint/filemanager/storage.py", line 881, in _save_metadata
    safeRename(metadata_temporary_path, metadata_path, throw_error=True)
  File "/usr/lib/python2.7/site-packages/OctoPrint-1.2.0_dev_395_g4e4ad0b-py2.7.egg/octoprint/util/__init__.py", line 185, in safeRename
    os.rename(old, new)
OSError: [Errno 18] Invalid cross-device link
```
This didn't crash octoprint but .metadata.yaml never was written.

Found something about this issue:
https://bugs.launchpad.net/python-keystoneclient/+bug/1265190

 I build and run on a Beaglebone Black, Angström Linux 2014.06 3.14 kernel (thing-printer.com image), OctoVersion: 1.2.0-dev-395-g4e4ad0b-dirty (devel branch)

So changing line 185 to
```python
shutil.move(old, new)
```
will do the trick. Don't forget to import shutil.

GitIssueBot:
Hi @derpicknicker1,

It looks like there is some information missing from your ticket that will be needed in order to process it properly. Please take a look at the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) and the page [How to file a bug report](https://github.com/foosel/OctoPrint/wiki/How-to-file-a-bug-report) on the project wiki, which will tell you **exactly** what your ticket has to contain in order to be processable.

If you did not intend to report a bug, please take special note of the title format to use as described in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md).

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next two weeks (until 2015-01-27 13:20) I'll close this ticket so it doesn't clutter the bug tracker.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

PR #717, opened 2015-01-13
Bugfix for Invalid cross-device link on Linux when /tmp is on another fs

Renha:
Hello. There are fix for #710 and #716 

When octoprint going to move a file from one directory to another on linux, if directories is located on different filesystems, os.rename wouldn't work, raising error "OSError: [Errno 18] Invalid cross-device link". In that pull request it is fixed with changing os.rename into shutil.move

Another effect of changing: shutil.move is high-level operation, which uses os.rename if it is acceptable, so maybe win32-specific code isn't needed with shutil (investigations still needed though).

### DIFF
--- a/src/octoprint/util/__init__.py
+++ b/src/octoprint/util/__init__.py
@@ -9,6 +9,7 @@ import time
 import re
 import tempfile
 import logging
+import shutil
 from flask import make_response
 
 from octoprint.settings import settings, default_settings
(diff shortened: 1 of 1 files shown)


dattas:
This appears to completely fix issue #710 for me.

foosel:
Just FYI, that's also the fix I had in mind, I just have to make sure that
there are no locations left which I didn't come around to yesterday, hence
no merge yet. Plus I wanted to experiment with win32 behaviour, looks like
that ugly workaround is still necessary though, since it's against win's
reluctance to overwrite existing files which still causes trouble here.

PR #732, opened 2015-01-19
Inherit clean command so that it supports the usual flags.

DanLipsitt:
Debian packaging tools expect setup.py to take the `--all` flag (and its abbreviation, `-a`). This PR meets those needs by subclassing the standard setuptools `clean` command, which supports those flags.

This PR changes the semantics of setup.py slightly compared to previous versions of Octoprint. '--all' now needs to be passed in order to delete everything that is deleted by previous versions. Some additional build files may also be deleted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ versioneer.tag_prefix = ''
 versioneer.parentdir_prefix = ''
 
 from setuptools import setup, find_packages, Command
+from distutils.command.clean import clean as Clean
 import os
 import shutil
 import glob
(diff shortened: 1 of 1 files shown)


DanLipsitt:
Oops, I might have based this on the wrong branch. I'll take another look.

foosel:
The branch is just fine, but it doesn't work just like that, since ``self.build_lib`` is not available upon running that due to not being invoked by the overriding implementation.

Wanna fix this or should I? :)

DanLipsitt:
The branch seems okay, the problem is that it doesn't merge against your devel branch right now.

However, I don't understand the issue with build_temp, and I can't replicate it locally in a fresh virtualenv and checkout:

```
$ git reset --hard feature/inherit-clean 
HEAD is now at 3261ef9 Inherit clean command so that it supports the usual flags.
$ python ./setup.py clean
/Users/dan/.virtualenvs/tmp-c849bbe4eda935e6/lib/python2.7/site-packages/setuptools/dist.py:293: UserWarning: The version specified ('1.1.0-dev-232-g3261ef9') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running clean
```

This works after `setup.py build` as well. I'm happy to work on it more but I don't know how at the moment.

foosel:
I've (finally) come around to take another look at this and I still could reproduce the above error. However I've now fixed things, and without changing semantics of the command compared to current behaviour.

I overrode the command option finalisation so that `--all` is always set, unless `--orig` is provided (in which case the command behaves exactly like stock, no egg clean up, no pyc cleanup). Additionally I've added `--nopyc` and `--noegg` to prevent those steps from happening. I've also removed redundancy, the `build` folder no longer needs to be cleaned manually since the stock procedure already takes care of that when `--all` is provided and the folder is empty.

PR #738, opened 2015-01-22
Fixed wrapping of icons at Access Control settings page with some translations

Renha:
Hello. Working on Russian translation, I have found out that in some reason when table headers are longer than English ones, Action icons heve wrapped to two lines, which looks awful. That pull request should fix that for any translation


foosel:
You are right that this style should be applied to the action cell, how don't you think it would be better to do this in ``octoprint.less`` and make it part of the general style sheet that way? I'm not a big fan of mixing style and markup and usually try to avoid it where possible.

Renha:
You are right, of course, I shouldn't do things I don't exactly know how to do, I'll make another solution. Now i've inspected octoprint.less file and found out reason why Russian text doesn't fit, it's because of fixed width. I should do some experiments here and will make another pull request with more reliable solution. Thank you.

foosel:
> I shouldn't do things I don't exactly know how to do

Oh, no, you should! This is the way people learn, doing things, then learning how to do them better :)

PR #742, opened 2015-01-26
Fixed wrapping of Icons at Action column in Access Controll settings page

Renha:
Icons at Action column in Access Controll settings page are now wrapped correctly with any translation language
It's done by editing of octoprint.less file, octoprint.css file was regenerated also.


Renha:
Is something wrong? I need to improve something?

foosel:
No, I just sometimes need some time not in front of the computer and put up the feet a bit, like everyone else ;)

PR #752, opened 2015-02-01
Fix logger call for gcode multi commands events

Diaoul:
Missing `_` causes the whole thing to fail.

      File "/home/pi/OctoPrint/src/octoprint/events.py", line 110, in _work
        listener(event, payload)
      File "/home/pi/OctoPrint/src/octoprint/events.py", line 267, in eventCallback
        self.executeCommand(processedCommand, commandType)
      File "/home/pi/OctoPrint/src/octoprint/events.py", line 275, in executeCommand
        self._executeGcodeCommand(command)
      File "/home/pi/OctoPrint/src/octoprint/events.py", line 296, in _executeGcodeCommand
        self.logger.debug("Executing GCode commands: %r" % command)
    AttributeError: 'CommandTrigger' object has no attribute 'logger'

### DIFF
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -293,7 +293,7 @@ class CommandTrigger(GenericEventListener):
 	def _executeGcodeCommand(self, command):
 		commands = [command]
 		if isinstance(command, (list, tuple, set)):
-			self.logger.debug("Executing GCode commands: %r" % command)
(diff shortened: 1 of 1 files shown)


foosel:
Apparently I forgot to backport f5eef06e88e11f2b528b09a228d8acd1861a5a30#diff-ff86b360031ff8cca3fc4cc1806e769e unto ``master``

PR #760, opened 2015-02-07
Added option for timelapse on beep (M300) event

vector76:
I'm new to Git and GitHub, so sorry in advance if this is not the right mechanism for submitting a suggested change.

This modification allows G-code triggered frames for timelapse video.  This allows me to move the print head out of the way and position the XY stage in the same position for each frame, as seen here:
https://www.youtube.com/watch?v=b3LWAffjjic

Because of the firmware command queueing, it is not enough to add M300 to the layer-change G-code.  I added "G4 P100" about 20x prior to the M300 command to force execution and flushing of the X/Y movement prior to the snapshot.  This all happens on the slicing side, not within OctoPrint, but it is necessary to get the results.

Feedback or comments welcome.

### DIFF
--- a/src/octoprint/server/api/timelapse.py
+++ b/src/octoprint/server/api/timelapse.py
@@ -27,6 +27,9 @@ def getTimelapseData():
 	if timelapse is not None and isinstance(timelapse, octoprint.timelapse.ZTimelapse):
 		config["type"] = "zchange"
 		config["postRoll"] = timelapse.postRoll()
+	elif timelapse is not None and isinstance(timelapse, octoprint.timelapse.BeepTimelapse):
+		config["type"] = "onbeep"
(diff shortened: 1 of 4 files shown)


jneilliii:
Something like this would be a great addition for those of us that have non-box printers where the bed moves in the Y direction.  I think it creates better timelapses of the print itself and not the printer's operation.

I think you should do these changes against the devel branch though and not the master branch before foosel will even consider the addition.  There are already some timelapse changes in the devel branch that may conflict with your changes, but I'm unsure of that since I really didn't look.

vector76:
I haven't done any timelapses in a while but I'll definitely check out the plugin next time I decide to do this.  I agree, plugin architecture is definitely nicer than having main-line code pulled in a hundred directions with possibly conflicting requirements.

PR #761, opened 2015-02-08
Fixing a missing generic counter for the templates

silasb:
Unless I'm miss understanding, I think we are missing the counter for the Generic template.


foosel:
Already fixed by #759

PR #767, opened 2015-02-14
Add apiKey to discovery plugin (new param in plugin config and xml template)

scotthraban:
Adding the apiKey as a configuration option for the discovery plugin, which is then sent in the discover.xml template, to enable an application to be able to control OctoPrint after auto discovery.

It would be better if the apiKey could be pulled directly from the regular location in the configuration, but it appears that the plugins are only able to see their specific config.

On the other hand, it is a bit of a security risk to broadcast the apiKey to the local network, so maybe it is better to keep adding the apiKey to the discovery plugin as an explicit step.

NOTE: In parallel I am working on a printer connection for Cura that will enable Cura to auto-discover and print directly to OctoPrint - Slic3r can already do this (minus the auto-discovery).

### DIFF
--- a/src/octoprint/plugins/discovery/__init__.py
+++ b/src/octoprint/plugins/discovery/__init__.py
@@ -122,6 +122,7 @@ class DiscoveryPlugin(octoprint.plugin.StartupPlugin,
 		                                                     modelUrl=self._settings.get(["model", "url"]),
 		                                                     serialNumber=self._settings.get(["model", "serial"]),
 		                                                     uuid=self.get_uuid(),
+		                                                     apiKey=self.get_api_key(),
 		                                                     presentationUrl=flask.url_for("index", _external=True)))
 		response.headers['Content-Type'] = 'application/xml'
 		return response
@@ -674,6 +675,9 @@ class DiscoveryPlugin(octoprint.plugin.StartupPlugin,
 			self._settings.set(["upnpUuid"], upnpUuid)
 			self._settings.save()
 		return upnpUuid
+
+	def get_api_key(self):
+		return self._settings.get(["apiKey"])
 
 	def get_instance_name(self):
 		name = self._settings.globalGet(["appearance", "name"])
--- a/src/octoprint/plugins/discovery/templates/discovery.xml.jinja2
+++ b/src/octoprint/plugins/discovery/templates/discovery.xml.jinja2
@@ -18,5 +18,6 @@
         <serviceList>
         </serviceList>
         <presentationURL>{{ presentationUrl }}</presentationURL>
+        <apiKey>{{ apiKey }}</apiKey>
     </device>
 </root>


foosel:
Hm... I really don't like the security implications of this, especially since (at least right now) requesting ``discovery.xml`` is not limited to the local network, so you are basically allowing the API key to be retrieved from everywhere the OctoPrint instance is accessible, not just the local network - so without limiting the access to ``discovery.xml`` to private and/or local networks as well, this can't be merged. 

I thought about this previously and this was actually one of the reasons for creating [App Session Keys](http://docs.octoprint.org/en/devel/api/apps.html) - have you taken a look at those? They add an additional handshake step before you can access the API regularly and necessitate the app to be registered via a public RSA key and some identifier within OctoPrint, but the latter can also be done via a special type of plugin.

scotthraban:
Gina,

Thanks - I had not seen the APP Session Keys, so I took a look at those. I
do have a few questions about how they would be used in practice.

It seems that the expectation is that the private/public key is meant to be
baked into the App talking to OctoPrint, but if the App is open source, the
private key cannot really be considered secure,

Even if the private key is kept secure, it also means that you must allow
all instances of that App to talk to a given OctoPrint, so allowing Alice
to print from the App to OctoPrint means that Bob must also be able to
print from the App to OctoPrint, assuming that they have the same version
of the App.

However if the private/public key is generated on the fly by the user after
installation, then this would mean (I think?) that only one instance of any
given App / version combination would be able to talk to any given
OctoPrint, as you can only register the one public key.

I do like the concept of a temporary session key - have you considered
allowing the temporary session key to be generated on the fly through the
username/password protected section of OctoPrint, instead of, or in
addition to, the API Key protected section? The user could login to the
OctoPrint page, OctoPrint could generate a temporary API Key, which the
user could paste back into the App interface, and that API Key could be
used until it expires, at which point the user would need to go back to
OctoPrint to generate a new temporary API Key. Or maybe the App could even
be allowed to extend the expiration on its temporary API Key.

Without that, it would seem that the App would need to request the current
global API Key from the user after discovery, and it would store this (in
preferences/config or just in memory) for use in talking to the API. This
would allow limiting the usage the API to only those users that know the
API Key, and more than one user can have access to the API through the same
App/Version combination, but it does require potentially broad
dissemination of the API Key.

Thoughts?

Thanks,
--Scott

foosel:
Yeah, the confidentiality of the app private key is indeed a problem here, which is also why the configuration basically allows to disable a given public key (and hence corresponding private key) for an app if that happens to have been compromised. You are also right in that there's no distinction at the moment on a per-user basis for the API keys - it's something I thought about adding but I didn't get around to it yet.

But I'm getting the feeling it would be better to add OAuth2 to this thing and be done with it - what you are describing there is basically just that, with something like a password grant type. Funnily enough adding OAuth to OctoPrint was something I was thinking about anyways, so maybe it might be a good time to revisit that idea.

scotthraban:
I am going to close this pull request, since there are some fundamental issues with its approach.

PR #775, opened 2015-02-18
Error messages show the proper name of the objects

matobago:
Just try to figure out why some binding wasn't working on my version and came out with this small change


foosel:
Funnily enough I just stumbled across this myself while debugging something this afternoon, so great timing :+1:

PR #782, opened 2015-02-24
Timelapse FPS option

Mikk36:
Fixed up the timelapse view model response parsing a bit.
Added the option to be able to specify custom fps for the timelapse video and fixed bitrate argument for current versions of ffmpeg.


Mikk36:
Removed the duplication.

PR #817, opened 2015-03-19
Fix inability to set printer profile extruder feedrate from settings UI

CapnBry:
In Settings -> Printer Profiles -> Edit Pencil Icon, if a user changes the Extruder axis feedrate, the setting isn't saved. This is because the editorData JSON object doesn't include the E axis when the data is PATCH posted.

This code patch patches the PATCH to patch in the E... patch and allow a handsome user such as myself to modify a printer's extruder feedrate which is otherwise completely impossible and makes everyone feel hopeless and in need of cookies. I LOVE COOKIES


foosel:
How could I resist a PR from such a handsome devil who on top of everything else loves cookies...

PR #823, opened 2015-03-21
Fix: Settings modal did not display properly on iPad and other browsers ...

webmonger:
...under 1200px wide

The float in the modal over lay on browsers that are under 1200px wide cause the content that was supposed to be on the right hand side to flow below it.

I've changed the width of the menu and right content to use percentages, this might not be desired but it was the only way I could think to fix it. I'm far from a CSS expert so there might be a better way.


foosel:
I solved it a bit differently and made the settings dialog more responsive in general.

PR #828, opened 2015-03-30
update FileSelected event to be more consistent with other events

jminardi:
This PR makes the payload of the `FileSelected` event the same as the start, cancel, and pause events by adding a `filename` field.


foosel:
Could you also update the documentation in ``docs/events/index.rst``? That would make it complete.

jminardi:
Updated the docs.

PR #830, opened 2015-03-30
Show .gcode .gco and .g files in the SD file list

markwal:
I'm not sure, but it looked like the intent of the type hierarchy was to allow for this.  But perhaps you were waiting for more testing?


foosel:
I manually cherry picked the original commit from your branch. I can't merge this PR since you opened it against your own `devel` branch instead of a dedicated PR branch, hence it contains all of your other (personal) changes as well.

Btw, I like what you are doing there in regards to GPX support. Do I understand it correctly that the magic happens in a custom wrapper around the serial port which basically translates anything on the fly?

markwal:
Ah.  A PR branch.  I guess my github noob is showing.  I'll do the next one that way.

Yes, my thinking was that not only would I get up and running quicker if I jumped off of the GPX's work, but also it'd be a way to keep things running in the future.  GPX's purpose was to take slicer output and make printable machine code out of it and it attempts to be compatible with current gcode in the process.  Since many people use it, I figure it has a good chance of keeping up with the times.

My thought was that if I wrap the serial object, that keeps the impact on the existing code pretty minimal and pushes the complexity out of the main line.  In the current form it needs a plugin hook to do the wrapping.  I'll create a brainstorming pull request off of a clean branch.

PR #840, opened 2015-04-06
Allow user to specify arbitrary metadata when uploading a file.

jminardi:
This PR allows the user to specify arbitrary metadata when uploading a file through the API endpoint.

With these changes `POST`ing to `/files/<target>` will check to see if either `userdata` or `userjson` was included in the form data. If `userdata` was included then it is stored with the file by calling the `FileManager.set_additional_metadata()` method with the string "userdata" as the key and the supplied string as the value. If `userjson` was included in the request then the string is decoded as a JSON object before storing it under the key "userjson". Both parameters are optional.

This PR is usefull for users who want to associate additional data with a file, as any additional metadata stored in this manner is returned when `GET`ting the file's properties.

### DIFF
--- a/docs/api/fileops.rst
+++ b/docs/api/fileops.rst
@@ -219,6 +219,9 @@ Upload file
                     to ``false``.
    :form print:     Whether to start printing the file directly after upload (``true``) or not (``false``). If set, `select`
                     is implicitely ``true`` as well. Optional, defaults to ``false``.
+   :form userdata:  [Optional] An optional string that if specified will be saved along with the file as metadata.
+   :form userjson:  [Optional] An optional string that if specified will be intepreted as json and then saved along with
+                    the file as metadata.
    :statuscode 201: No error
    :statuscode 400: If no `file` is included in the request, or the request is otherwise invalid.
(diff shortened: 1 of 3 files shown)


foosel:
Can you move this PR to it's own branch please? There's another commit in there that doesn't belong (startup script) - this stuff happening is btw the reason for the requirement in the [contribution guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests) to file PRs from a dedicated branch, just happens too fast to accidentally include unrelated changes otherwise :/ You'll need to cherry pick the commits to a different branch (or just fork that from fb99c56 without 5590bd7) and then file a new PR.

The metadata attached to files was btw supposed to be quite flexible to begin with, so it might even make sense to just call the request parameter `metadata`, always assume it to be JSON and just add anything from there that's not a reserved keyword (so anything that's not `hash`, `analysis` or `links`). On the other hand that might cause trouble down the road when a new keyword is to be reserved by the core software. Difficult... Maybe prefixing the existing reserved keywords with a `_`? (yes, I'm thinking aloud here...)

jminardi:
Sorry, I should have read more carefully! New PR incoming.

PR #842, opened 2015-04-07
Web Security/Enhanced Privacy

captainha:
Please add a feature to allow no information to be shows without logging in. Even though a non-privileged user can only see read only information, there is still a lot of information available for them to read. No webcam, gcode, anything. Connecting to the system should only display a login page.


foosel:
I have no idea what that PR is supposed to do here, not merging.

Your problem can be solved by a proper authentication setup **in front of** OctoPrint, google HTTP Basic Auth and ideally VPN solutions.

captainha:
I've clearly submitted it wrong, but I'm simply trying to suggest a new feature.It would be nice to do what you suggested without having to put some additional infrastructure in front of octoprint. App-level authentication should be able to do this.

foosel:
[See the FAQ](https://github.com/foosel/OctoPrint/wiki/FAQ#i-want-to-access-my-octoprint-installation-from-the-internet-but-i-only-want-people-i-know-to-be-able-to-see-anything)

PR #849, opened 2015-04-13
Allow user to specify arbitrary metadata when uploading a file. (Redux)

jminardi:
(_Reopened from a new branch as specified in the docs. Original details:_)

This PR allows the user to specify arbitrary metadata when uploading a file through the API endpoint.

With these changes `POST`ing to `/files/<target>` will check to see if either `userdata` or `userjson` was included in the form data. If `userdata` was included then it is stored with the file by calling the `FileManager.set_additional_metadata()` method with the string "userdata" as the key and the supplied string as the value. If `userjson` was included in the request then the string is decoded as a JSON object before storing it under the key "userjson". Both parameters are optional.

This PR is usefull for users who want to associate additional data with a file, as any additional metadata stored in this manner is returned when `GET`ting the file's properties.

### DIFF
--- a/docs/api/fileops.rst
+++ b/docs/api/fileops.rst
@@ -219,6 +219,9 @@ Upload file
                     to ``false``.
    :form print:     Whether to start printing the file directly after upload (``true``) or not (``false``). If set, `select`
                     is implicitely ``true`` as well. Optional, defaults to ``false``.
+   :form userdata:  [Optional] An optional string that if specified will be saved along with the file as metadata.
+   :form userjson:  [Optional] An optional string that if specified will be intepreted as json and then saved along with
+                    the file as metadata.
    :statuscode 201: No error
    :statuscode 400: If no `file` is included in the request, or the request is otherwise invalid.
    :statuscode 404: If `location` is neither ``local`` nor ``sdcard`` or trying to upload to SD card and SD card support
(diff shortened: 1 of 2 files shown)


jminardi:
From @foosel in the original PR:


> The metadata attached to files was btw supposed to be quite flexible to begin with, so it might even make sense to just call the request parameter metadata, always assume it to be JSON and just add anything from there that's not a reserved keyword (so anything that's not hash, analysis or links). On the other hand that might cause trouble down the road when a new keyword is to be reserved by the core software. Difficult... Maybe prefixing the existing reserved keywords with a _? (yes, I'm thinking aloud here...)

jminardi:
I considered doing what you suggested but I was worried about key collisions (maybe easy to avoid now but could break in the future). I think namespacing the user data under only one key like I have here is more general than mangling their keys to make sure they don't clash. And your metadata system is general enough that nesting works as expected.

I also initially had it just decode everything as JSON but I wanted to allow someone to specify a simple string or int and not worry about making it valid JSON. However if you think that is not really needed I can remove it.

I do think putting all userdata under its own key is better than combining it into the data used by the system though.

foosel:
Did some small changes (I hope you can live with them):

   * Renamed `userjson` to `userdata`, removed original `userdata` (developers will live if they need to ensure their supplied userdata is valid JSON)
   * Input validation for `userdata` parameter - a HTTP 400 response will be generated if userdata can't be parsed as JSON
   * Adjusted documentation accordingly

jminardi:
Your changes are fine, thanks for merging!

PR #850, opened 2015-04-13
Add a plugin hook for opening the serial port

markwal:
This hook allows the plugin to create the serial object (or pass to the
next or the builtin object) which allows for serial communcation
filtering or a different communication transport.

The hook is pretty general for multiple purposes.  My purpose is to filter serial communication and translate to and from the s3g serial protocol and emulate g-code both send and response.  I've made a plugin at markwal/OctoPrint-GPX that does it.  I have it working for me currently.

The one change here that is specific is counting the x3g and s3g extensions as printable machine code.  It seems reasonable for this list to encompass future machine code types, but perhaps we should add extensibility here as well?

These two changes together are sufficient to allow OctoPrint-GPX to work.


markwal:
I attempted to resolve that merge conflict by doing a rebase, but also a github noob.  My branch looks good and it's rebased to the head of your devel branch so I'm not sure what github wants now.  I could create a clean branch and submit a new pull request.  What would you prefer?

foosel:
Closing that one since it was replaced by #854

foosel:
For the record, you should just have merged in your branch, not rebased. By Rebasing you basically rewrote history, that can cause merge issues.

http://justinhileman.info/article/git-pretty/ ;)

Issue #682, opened 2014-12-16
[Request] Allow Bed position to be set in gcode viewer


thinkl33t:
![screenshot from 2014-12-16 23 40 53](https://cloud.githubusercontent.com/assets/1016780/5464203/1058b1f2-857d-11e4-9fde-b57653950be6.png)

My bed center is at [0,0], not the usual [100,100], which means the grid on the gcode viewer is off to one side.  Not a major issue, but a config variable to set the bed's width, height, and 0 position would be very nice :)

stv0g:
Having the same problem. :+1:  Especially for Delta Robots where parts are centered around the origin coordinates.

foosel:
@stv0g for delta robots you can just set the print bed to be circular, this will automatically make the origin be in the center. While you need the `devel` branch for that, this actually IS already possible. 

This ticket is actually about an arbitrary center location for rectangular beds, which is NOT yet possible, not even in `devel`.

stv0g:
Oh thanks @foosel So I will check out the devel branch :-)

foosel:
THanks to work by @markwal, that's now available on `devel`. No arbitrary location on the printbed, but you can choose between "center" and "lowerleft" for rectangular beds. Circular beds are limited to "center" for now.

PR #852, opened 2015-04-15
Add a printer profile option to allow centered origin

markwal:
This is an attempt at fixing #682.  It doesn't add arbitrary origin, but adds a checkbox for the most common one (other than the default lower left).  Some printers have (0, 0) at the center of the bed.  Circular beds commonly, but also some rectangular beds.


markwal:
Here's some updates based on feedback.  I successfully sliced with the Cura addin for my printer which has the origin in the center.

PR #854, opened 2015-04-15
Add a plugin hook for opening the serial port

markwal:
This hook allows the plugin to create the serial port object (or pass to
the next hook or the builtin serial port handler) which allows for
serial communication filtering or a different communication transport.

Use @foosel's code to use the same loop for hooks as well as the
builtin.


markwal:
I really mangled the history on that other one.  This one is prettier.

PR #872, opened 2015-04-24
Last line number fix for Marlin

Booli:
I had issues with last line number being resend over and over. 

the_handleResendRequest looks for the wrong string to check for in Marlin. 

"line number is not line number"  should be "line number is not last line number" as per:

https://github.com/MarlinFirmware/Marlin/blob/5e3bb4bdf3a6434dbde9b0873e3b7d0c84c0e96d/Marlin/language.h#L125

Also checking against this might break different language Marlin, this is a quick fix for English.

(sorry I have a parethesis change in another file in here too :( it's late on friday)

Booli


foosel:
Made obsolete by making the whole line more flexible, see commit 517a32f08fdb676bb7b349681f3da2c8c058a2b8

PR #883, opened 2015-05-05
Fix for File not able to delete after Serial Error

Salandora:
As said in IRC, if you print and a Serial error occurs you were not able to delete the file.

This small fix takes care of it, every PrintingGcodeFileInformation subscribes to the FILE_DESELECTED event, and closes the _handle if still open.

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -356,7 +356,7 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 		self._setCurrentZ(None)
 
 	def unselect_file(self):
-		if self._comm is not None and (self._comm.isBusy() or self._comm.isStreaming()):
+		if self._comm is None or (self._comm.isBusy() or self._comm.isStreaming()):
 			return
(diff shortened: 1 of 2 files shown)


foosel:
Hm... I don't like using the event subsystem for that to be honest. First of all it adds a rather long roundtrip (through the event subsystem) for something that could be handled better by ``comm.py`` itself (which has the reference to the `PrintingGcodeFileInformation` anyways), secondly I'm not sure if that covers all bases (e.g. if the `_monitor` loop runs into an exception the deselected event will not be sent as far as I can see right now).

I'd suggest to instead close it if it's open when switching to the `ERROR` state. That should cover all bases from what I'm seeing.

Salandora:
So you mean when _changState(ERROR/CLosed_WITH_ERROR) is called it shall close the handle there?

Salandora:
Alright I'll change it :-)

foosel:
As talked about on IRC, quickly moved to comm.py and turned into one single commit as seen above.

PR #888, opened 2015-05-11
Stop sockjs from multiplying on reconnect

markwal:
Have you ever noticed when developing that every time you stop and start
the server, the terminal window gets an extra duplicate line for every
reconnect attempt?  Well, it's because (I think) "delete" in javascript
just removes the indicated name from the namespace, it doesn't actually
free up an object. Those zombie objects are still there and wake up (for
some transports) on reconnect. Might be different in SockJS v1 or later.

Anyway, this fixes it for me in Chrome.


nophead:
Yes I have seen the terminal lines get duplicated, then triplicated, etc.
Not developing though, just using OctopPrint.

Issue #851, opened 2015-04-13
[Brainstorming] SD card handling


markwal:
I really prefer always printing from the SD card.  It's nice and reliable and the bot doesn't have to work very hard to handle serial communication to keep the monitor up to date vs. keeping up with communications so there aren't any pauses or timeouts during a print.

I'd also like to take better advantage of Toshiba's FlashAir card by using the wifi bandwidth rather than serial bandwidth for uploading gcode to the SD card.

One approach for a plugin is to replace the Files sidebar part with a part from the plugin that does its own file manager for the SD card over the air along with a rest API for communicating with the front end.  This basically works, but there are a couple snags:

1. The main page already sinks the dragover event for the page, so I'd like to have a (blessed?) way for having the plugin get a shot at handling the event if it wishes

2. I'd like to hook the Print button on the PrinterState element to always send to the SD card and print the SD card version (with a setting to enable/disable that behavior).  I can stomp on the click event on that button in the javascript, but that seems a little grotty.  Is there a better way?  Perhaps a backend hook instead?  There's a backend event, but its async so doesn't quite serve my desired outcome.  I could replace the whole PrinterState part, but that's seeming like too far for what I'm trying to do.

I'm willing to do the work for creating the hooks what not, but I'd like to head in the right direction with them.

The other thing is that it seems like we could do better tracking/monitoring the print even when it is from the SD card.  My printer isn't willing to tell what byte it's on, but I can poll for current Z use that for time estimate, layer in the gcode viewer and triggering a timelapse photograph.

foosel:
One other idea... I've been meaning to extract the SD card handling into it's own `StorageInterface` implementation and allowing adding additional storage interfaces via plugins. If the regular SD card implementation was provided by a (pluggable) storage interface, it would be easy to swap it with an implementation that utilizes a FlashAir card instead (btw, I'm happy to hear you got this to work, I experimented briefly with a Transcend Wifi enabled SD card a while back but that did not implement access to the data layer via SPI and so I couldn't use it in the printer :/)

Something like

```
     File management                                                                           
+----------------------+                                                                       
|                      |                                                                       
| +------------------+ |    +-----------------------+                                          
| |                  | |    |                       |                                          
| |      local       +------>      File system      |                                          
| |                  | |    |                       |                                          
| +------------------+ |    +-----------------------+                                          
|                      |                                                                       
| +------------------+ |    +-----------------------+   +-----------------+   +---------------+
[...]

PR #889, opened 2015-05-11
[WIP] Initial SDCardFileStorage implementation of StorageInterface

markwal:
This isn't really ready for a pull. I just created it to continue the conversation on issue #851 with some code.

This initial change is intended to create a StorageInterface
implementation for the sdcard. At this point it doesn't attempt to
change any functionality outside of that already supported for SD cards,
just change the code flow through the fileManager.

Is this the direction you were talking about in issue #851?  What way
would you prefer to handle the unsupported (yet) functionality.
NotImplemented as here or with io.Unsupported as specified for some?

Also, note how I've got this calling back through the fileManager for
LOCAL storage on upload. Should it work this way or should it be a
subclass of LocalFileStorage or some other structure? To do some of the
other StorageInterface, more of it will want to rely on the
LocalFileStorage, I think.

Perhaps the SDCardFileStorage should create its own LocalFileStorage at
a different base path as an SDCard shadow to do some of the metadata,
etc. As an actual mirror rather than just for upload, it would be more
reliable for some of the metadata, analysis purposes. It could also be
enhanced if firmwares were willing to give us a hash for comparison (to
avoid streaming the file to check).

### DIFF
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -299,7 +299,7 @@ class FileManager(object):
 			result[dst] = self._storage_managers[dst].list_files(path=path, filter=filter, recursive=recursive)
 		return result
 
-	def add_file(self, destination, path, file_object, links=None, allow_overwrite=False, printer_profile=None, analysis=None):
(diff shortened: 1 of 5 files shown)


foosel:
I guess this is no longer being worked on? Not a problem, but closing it then. Can always reopen it :)

PR #894, opened 2015-05-15
Devel

prophetmaster:
it's possible to add a menu separator in config.yaml ?


foosel:
A pull request is not for asking questions (I guess you pressed the wrong button here). I also don't understand what a "menu separator in config.yaml" is supposed to be (what menu?). I'm guessing you mean if it's possible to add a separator in the system command menu, and now, it's not possible at the moment. [Open a Feature Request for that if you want](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#issues-tickets-however-you-may-call-them).

PR #901, opened 2015-05-21
Only eat two characters when the error indicator is '!!'

markwal:
Same as 3 lines earlier now that I do the PR.  Would you rather I did:
error = line[6:] if line.startswith("Error:") else line[2:]
before the if and then use the local in each case?


foosel:
Somehow I never got the notification about this and it's fine like it is

PR #912, opened 2015-05-28
scripts: init and default files now specify a process priority.

aerickson:
default to priority of 18 (nicelevel of -2).

///

It's nice to have the Octoprint process priority a little higher than everything else. This makes that happen.


foosel:
[Yes, but...](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests)

> Please create all pull requests against the devel branch.

Thanks in advance :)

aerickson:
Sorry about that. On the way.

PR #922, opened 2015-06-03
Pip result parsing version numbers and extra lines

markwal:
Some pip invocations return extra info after the Successfully installed
(like "Cleaning up...").  Also, not all packages include a version
number in their package names.


markwal:
I considered a regex [with a conditional group and a non-greedy initial group] for the version number, but this seemed easier to read.

PR #961, opened 2015-07-07
pluginmanager: case handling and submit binding

markwal:
* Convert query term to lower case so that it is case insensitive both ways
    * Knockout submit binding takes the form element as parameter and return value determines whether the form submit occurs. See http://knockoutjs.com/documentation/submit-binding.html


foosel:
Merged on `devel`, then cherry-picked on `maintenance`

PR #962, opened 2015-07-07
Change repo_unavailable path to absolute

markwal:
Relative paths are problematic with bundling since the css location will be different based on whether bundled or unbundled. Unfortunately, absolute path is sensitive to plugin name changes and plugin asset url changes, but better than being sensitive to a runtime change.


markwal:
Oh yeah.  That's configurable.  Oops.

PR #988, opened 2015-07-26
Add status support in init script

Diaoul:


Diaoul:
Ref #987

foosel:
Thank you for this. Could you please take a look at the following points?

1. I'm not too happy about the init script returning a return code indicating an error when the server is not running (`exit 1`). Any particular reason for that, e.g. a standard I'm not aware of right now? I'd expect a command like that to complete with a return code indicating success regardless of the status, and only indicate a failure if there was an actual error while trying to determine the status.
2. [The PR is created against the wrong branch](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests).
3. Don't forget to add yourself to the `AUTHORS.md` file.

Diaoul:
I think it is standard, cron does that:

    $ sudo service cron status && echo "Running" || echo "Not running"
    [ ok ] cron is running.
    Running
    $ sudo service cron stop
    [ ok ] Stopping periodic command scheduler: cron.
    $ sudo service cron status && echo "Running" || echo "Not running"
    [FAIL] cron is not running ... failed!
    Not running

I'll make another PR, sorry for the trouble.

foosel:
If cron does it like that, that's fine for me then. Although I really am a bit surprised by that behaviour, but I guess it makes  scripting easier.

Also thank you for understanding, looking forward to the new PR! :)

Diaoul:
New PR is #989
I'm already in the AUTHORS.md

It's indeed to make things easier for scripts. I'm using Salt to deploy OctoPrint and currently it doesn't know if the service is started or not so it restarts it whenever I deploy.

foosel:
Oh, sorry, I quickly scanned `AUTHORS.md` earlier but didn't see you. Next time I'll use the search ;)

PR #995, opened 2015-07-28
fixed bug: when new job start, the progress is 100%, not set to zero.

masterhou:


foosel:
Thank you but can you [please open that against the `devel` branch](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests)? :) 

I maintain a strict "devel first" workflow for changes and if necessary cherry pick fixes. `master` only gets updated upon new releases, which get prepared in `maintenance` for 1.2.x and `devel` for 1.3.0.

That way it's easier to keep track of changes.

PR #999, opened 2015-07-29
Temperature target events (foosel/OctoPrint#565)

spapadim:
I believe this addresses all things you mentioned in the discussion of foosel/OctoPrint#565; please take a look, looking forward to comments.

I've tested everything on the VIRTUAL printer, except for the "externally triggered heatup" logic and multiple toolheads (which I'm not sure how to check -- also, all of my physical printers are single-extruder without LCD).

I've added config parameters for most things (temperature and time thresholds, as well as an option to not trigger while printing which is True by default -- and this is checked first thing in the logic flow, to hopefully minimize overheads while printing, which I'm not sure is a real concern but.. whatever).

### DIFF
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -71,6 +71,9 @@ class Events(object):
 	E_STOP = "EStop"
 	REGISTERED_MESSAGE_RECEIVED = "RegisteredMessageReceived"
 
+	# Temperature
+	TARGET_TEMPERATURE_REACHED = "TargetTemperatureReached"
+
 	# Timelapse
 	CAPTURE_START = "CaptureStart"
 	CAPTURE_DONE = "CaptureDone"
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -170,7 +170,12 @@ default_settings = {
 			{"name": "ABS", "extruder" : 210, "bed" : 100 },
 			{"name": "PLA", "extruder" : 180, "bed" : 60 }
 		],
-		"cutoff": 30
+		"cutoff": 30,
+		"targetEvents": {
+			"valueThreshold": 2.0,
+			"timeThreshold": 8.0,
+			"manualOnly": True
+		}
 	},
 	"printerProfiles": {
 		"default": None,
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -149,6 +149,12 @@ class MachineCom(object):
 		self._pauseWaitTimeLost = 0.0
 		self._currentTool = 0
 
+		self._triggerTemperatureEvent = dict()  # Keyed by B, T0, T1, ...
+		self._temperatureWithinLimitStartTime = dict()
+		self._temperatureValueThreshold = settings().getFloat(["temperature", "targetEvents", "valueThreshold"])
+		self._temperatureTimeThreshold = settings().getFloat(["temperature", "targetEvents", "timeThreshold"])
+		self._temperatureManualOnly = settings().getBoolean(["temperature", "targetEvents", "manualOnly"])
+
 		self._long_running_command = False
 		self._heating = False
 
@@ -503,6 +509,13 @@ class MachineCom(object):
 		self._pauseWaitStartTime = 0
 		self._pauseWaitTimeLost = 0.0
 
+		if self._temperatureManualOnly:
+			# manualOnly semantics: If user manually commands a temperature change
+			# but starts a print before the target is reached,
+			# then no event will ever be triggered during the print
+			for tool in self._triggerTemperatureEvent:
+				self._resetTemperatureEvent(tool, triggerFlag = False)
+
 		try:
 			self._currentFile.start()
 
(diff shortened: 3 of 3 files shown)


spapadim:
Basically, should be ready for inclusion, after the code marked with `# TODO` is verified (for externally triggered heatup).

Also, a related patch, for the Pushbullet plugin: spapadim/Octoprint-Pushbullet@cc338f9ca2d850d3689f4fa9ae0111ac55d0b241  (I can create a separate PR later, if you'd like).

foosel:
Thanks! Currently on vacation and this deserves a more thorough look than I can currently give it (I'm trying to not work, as tricky as that may be ;)), so it might be that this has to wait until after my vacation (back at work on August 10th). Will get back to you ASAP though.

spapadim:
No problem, enjoy the vacation -- until then I'll probably have it more thoroughly tested on an older printer (still happily running Octoprint 1.0.0, for >1yr :) ).

spapadim:
Just so I don't forget (not trying to rush):  I tried testing for externally triggered heatup via smoothieware's console commands (http://smoothieware.org/console-commands), but it's not detected.  Not sure to what firmware the `("T:" in line or line.startswith("T") or ... ) and not line.startswith("ok")` logic applies to (BTW, isn't the `startswith("T")` check redundant?), but Smoothie responds in a different way:
```
Send: set_temp bed 35
Recv: bed temp set to: 35.0
```
and similar for the hotend.  

Anyway, I'll keep this PR to one thing and perhaps I can submit this as a separate patchset, if there is interest. Not sure who uses the console (for me it was the first time in >1yr :), _but_ perhaps the heatup detection logic can be implemented in a more generic fashion by just comparing previous and current target temps, and triggering every time these differ, thus handling all  cases (M104/M140/M109/M190/"external"/whatever...) uniformly.

spapadim:
Whoops, sorry, clicked wrong button! Doh!

foosel:
Sorry for only replying now, I'm having wrist issues and am only slowly coming back up to typing speeds, definitely no chance to review during the vacation.

In any case, I put some comments on a bunch of lines in your commit (see above), please check them.

> perhaps the heatup detection logic can be implemented in a more generic fashion by just comparing previous and current target temps, and triggering every time these differ, thus handling all cases (M104/M140/M109/M190/"external"/whatever...) uniformly.

Pointed this out in a line comment already (above) but just to reiterate...

The "external heatup detection" is actually a detection of "it's now heating up *and waiting* for that". This should only trigger when e.g. a SD card file being printed contains an `M109` or an `M190`. In that case the firmware will dump those temperature lines without an `ok` to serial. Once the heatup is complete, it will dump a final `ok` all on its own and become responsive to sent commands again. So basically the same output as when you send `M109` or `M190` through OctoPrint, just that you never did but the firmware did it on its own. 

When that final `ok` is received for that situation (that would be [here in the code](https://github.com/spapadim/OctoPrint/blob/94dee723f17106ec2476236ca975df1e39e72635/src/octoprint/util/comm.py#L977)) the "target temperature reached" event should be triggered. If something like `M109` or `M190` is active it should in fact in my opinion not try to do the threshold checking on its own, but instead use the firmware response - once the final `ok` is received, the event should trigger. So what you basically want to do is:

  * if an `M109` or an `M190` are sent or an external heatup is detected, *disable* your own temperature event detection
  * always fire a temperature event when such a heatup ends (`ok` is received, see [here](https://github.com/spapadim/OctoPrint/blob/94dee723f17106ec2476236ca975df1e39e72635/src/octoprint/util/comm.py#L977))
  * otherwise, if an `M104` or an `M109` are sent or a difference in the selected target temperature for a tool or the bed is detected (and the new target is not `None` or `0`, enable the event detection

This way you should have covered all bases I think:

  * if the user triggers a waiting heatup through OctoPrint, the event gets fired when the firmware sends the final `ok` for that
  * if a waiting heatup is triggered through the firmware itself (SD card file or controller or whatever other means there are), OctoPrint detects that and again only sends the event when the firmware send the final `ok` for that
  * if a "quick" temperature change is triggered through OctoPrint (`M104`, `M140`), the target detection is enabled and the event sent when that triggers
  * if an external "quick" temperature change is done, the reported target temperature changes, the target detection is enabled and the event sent when that triggers 

Hope this is understandable. That's what I was trying to point out in the original feature request too, but I think I explained myself not very well and hence there might have been some misunderstandings what I mean with an "external heatup" and how best to handle the various ways temperature changes can be triggered. Sorry :)

spapadim:
Thanks for the detailed comments/feedback!  Unfortunately summer nearing end for me means I have _less_ time for this. :(

I'll try to get back to this by weekend (basically: agree with most of what you wrote in comments, but would like to explain a couple of design decisions I made / alternatives I considered, before making further changes).  BTW, I've been running it on my both printers until now, seems to work reasonably.

foosel:
@spapadim are you still looking into this?

TylonHH:
So is the temperature event full implemented?

foosel:
@TylonHH I think at this point this PR can be considered abandoned. It was not merged due to the outstanding points outlined above and never updated to address them sadly.

PR #1000, opened 2015-07-29
Fix sorting of templates

Salandora:
I noticed that no sorting is happening no matter what "key" I gave for sorting.
So I investigated the situation and found that there was a hard coded check for key == name, 
but nothing for key != name

### DIFF
--- a/src/octoprint/server/views.py
+++ b/src/octoprint/server/views.py
@@ -230,8 +230,30 @@ def index():
 		# finally add anything that's not included in our order yet
 		sorted_missing = list(missing_in_order)
 		if template_sorting[t]["key"] is not None:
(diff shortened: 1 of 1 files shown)


Salandora:
I wasn't aware that this could be something else then a tuple. Otherwise I would have made a check like you did just with less lambda xD.

Shall I copy the code or do you commit it yourself?
As always I don't care ;-)

foosel:
Haven't tested it, only write it down (vacation ;)). Also currently having issues with my wrist, so if you could test it a bit I'd be grateful :)

Salandora:
Alright I test it.

Salandora:
Re opening PR with new Branch for better History

Issue #1001, opened 2015-07-30
"Connection" tab not unfolding


jminardi:
1. What were you doing?

Operating octoprint web interface

2. What did you expect to happen?

Connections tab to display when clicked

3. What happened instead?

Connections tab did not display when clicked

4. Branch & Commit or Version of OctoPrint:

Version: 1.2.4 (master branch)

5. Printer model & used firmware incl. version
   (if applicable - always include if unsure):

Marlin Integration branch

6. Browser and Version of Browser, Operating 
   System running Browser (if applicable - always 
   include if unsure):

Latest chrome on OS X

7. Link to octoprint.log on gist.github.com or pastebin.com
   (ALWAYS INCLUDE AND DO NOT TRUNCATE):

N/A

8. Link to contents of terminal tab or serial.log on 
   gist.github.com or pastebin.com (if applicable - always 
   include if unsure or reporting communication issues AND
   DO NOT TRUNCATE):

N/a

9. Link to contents of Javascript console in the browser 
   on gist.github.com or pastebin.com or alternatively a
   screenshot (if applicable - always include if unsure
   or reporting UI issues):

No errors seemed to show up:

```
Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.
packed_app.js?34dda18b:8354 Starting dependency resolution...
packed_app.js?34dda18b:8411 ... dependency resolution done
packed_app.js?34dda18b:8630 Going to bind 23 view models...
packed_app.js?34dda18b:8698 ... binding done
```

10. Screenshot(s) showing the problem (if applicable - always
    include if unsure or reporting UI issues):

I have read the FAQ.

ignaworn:
I would like to add that if you open the web  interface while the printer is desconected then the connection tab is expanded by default, and you can expand and contract normally.
Then if you connect to the printer and reload the webpage, the problem appears.

Using devel branch up to date and Google chrome on Windows 10.

markwal:
What plugins do you have installed?

markwal:
I thought maybe it was a side effect of something I did in ScreenSquish, but I've disabled it and still repro.  The reason it gets locked is that Collapse.transitioning gets stuck at 1 after #connection.collapse('hide') is called in the connection view model.  And it the Collapse.show it returns if transitioning is set.  I haven't figured out why transitioning gets stuck at true yet.

markwal:
[...]

PR #1002, opened 2015-07-30
Fix for #1001 connection tab not unfolding

markwal:
Defer collapsing the connection element until the whole wrapper part is
visible (it's hidden for non-admins).

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/connection.js
+++ b/src/octoprint/static/js/app/viewmodels/connection.js
@@ -86,6 +86,15 @@ $(function() {
             self._processStateData(data.state);
         };
 
+        self.openOrCloseOnStateChange = function() {
+            var connectionTab = $("#connection");
+            if (self.isOperational() && connectionTab.hasClass("in")) {
+                connectionTab.collapse("hide");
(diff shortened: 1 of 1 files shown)


src/octoprint/static/js/app/viewmodels/connection.js
```diff
@@ -97,15 +106,10 @@ $(function() {
             self.isReady(data.flags.ready);
             self.isLoading(data.flags.loading);
 
-            var connectionTab = $("#connection");
-            if (self.previousIsOperational != self.isOperational()) {
-                if (self.isOperational() && connectionTab.hasClass("in")) {
-                    // connection just got established, close connection tab for now
-                    connectionTab.collapse("hide");
-                } else if (!connectionTab.hasClass("in")) {
```
markwal: I also changed this line when I moved it. I think there's another bug here. At least, this isn't idempotent, because if the first isOperational is true but the tab has already been hidden, the else if doesn't check the isOperational again and will just show the tab incorrectly.

foosel:
Big thanks for that. I noticed that too from to time, but could never point my finger on what was causing it and of course was unable to reproduce it whenever I set out to hunt it down.

PR #1013, opened 2015-08-06
[WIP] Append environment information to the Bug Report link

r89m:
Initial implementation of including information in the bug report link. Not currently useful in it's present form, but could be used as a base for automating the inclusion of some of the data required for a bug report.

May need to consider privacy concerns (although there's no personal information included in the data, people may not like information about their system being sent to a third party)

Full implementation may need an additional page or two to be hosted at Octoprint.org as well as a couple of python scripts hosted on the OctoPrint instance itself.

Thought I'd post this as a mini proof-of-concept.

My thoughts on the process would be something along the lines of...
1. Link takes you to an Octoprint.org/.... page which strips off the information and stores it in a cookie which is valid for, say, an hour.
2. Octoprint page then automatically redirects user to bug reporting guide
3. Change link in bug reporting guide to point to Octoprint.org page. On this page are buttons that allow the user to automatically retrieve and upload their logs to pastebin or a gist. Request to user's host could be made via JSONP
4. Once user has uploaded logs (if desired) submit button takes them to GitHub new Issue page with  template partially filled in with Version, Browser / OS, urls to logs etc.

I'm happy to look into the implementation side on the host if you think this is a viable solution.

Comments welcome!

Richard

### DIFF
--- a/src/octoprint/server/util/flask.py
+++ b/src/octoprint/server/util/flask.py
@@ -600,6 +600,7 @@ def collect_plugin_assets(enable_gcodeviewer=True, enable_timelapse=True, prefer
 		'js/app/viewmodels/control.js',
 		'js/app/viewmodels/firstrun.js',
 		'js/app/viewmodels/files.js',
+		'js/app/viewmodels/footer.js',
 		'js/app/viewmodels/loginstate.js',
 		'js/app/viewmodels/navigation.js',
 		'js/app/viewmodels/printerstate.js',
--- a/src/octoprint/static/js/app/viewmodels/footer.js
+++ b/src/octoprint/static/js/app/viewmodels/footer.js
@@ -1,0 +1,27 @@
+$(function() {
+    function FooterViewModel(parameters) {
+        var self = this;
+
+        self.appendEnvironmentInfo = function(data, event){
+
+            // Build bug reporting URL
+            var environmentInfo = {
+                src : "ui",
+                version : $(".version").text(),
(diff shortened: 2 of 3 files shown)


shaggythesheep:
@Salandora no Python, but it's a start! :)

foosel:
Actuary on my way to bed, so only short:

I'd rather try to keep it all client side (with some multiple choice magic) than have multiple round trips over pages that don't exist yet and are purely static on top of that ;) gist allows anonymous uploads, so there skills be no problem pushing the locker that way, and screenshots should be possible via something like imgur from what I saw. So having the information that is now basically the contribution guidelines plus the bug reporting guide as some kind of multiple choice dialog:

* "I want to report a problem" 
* *collects log folders and js console* "what were you doing, what did you expect, what happened instead?"
* *enters data*
* *previews ticket and logs* "is this ok?"
* "yes"
* *uploads data* "please paste this at the issue tracker at $link. If you are already logged into github you can also click $this_link"


Should only be available to admins (log file access is necessary). Plus shortcut for people who've already done this a couple of times.

foosel:
Possibly interesting for some later date (link dump, sorry): https://github.com/ivoviz/feedback

shaggythesheep:
That flow seems much more simple / secure. I'll try my hand at putting this into a plugin, although I wouldn't expect any results in the short term! :)

Thanks for feedback link, I'm sure it'll make it's way into whatever I manage to put together.

foosel:
I guess this is no longer being worked on? Not a problem, but closing it then. Can always reopen it :)

PR #1014, opened 2015-08-06
Fix call to non-existent Mock method.

DanLipsitt:
I replaced a call to `mock.MagicMock.assert_called_once()`, which, as far as I can tell, doesn't exist in mock version 1.0.1 or 1.3.0 (`.assert_called_once_with()` exists). However, your travis builds look fine, so I must be missing something? It was causing test `AttributeError`s for me.

### DIFF
--- a/tests/filemanager/test_filemanager.py
+++ b/tests/filemanager/test_filemanager.py
@@ -259,4 +259,4 @@ class FileManagerTest(unittest.TestCase):
 		mocked_os.assert_called_once_with("tmp.file")
 
 		# assert that time.time was only called once
-		mocked_time.assert_called_once()
+		self.assertEqual(mocked_time.call_count, 1)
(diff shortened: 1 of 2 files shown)


foosel:
Whatever you are missing, I am missing it too, and it's also passing fine locally... but I don't see the method in the `mock` version installed in my dev environment, so it in fact *should* fail. It doesn't though. Which worries me I have to say.

DanLipsitt:
I figured out the problem. It has to do with the do-anything nature of Mocks:

```
>>> import mock
>>> m = mock.MagicMock()
>>> bool(m.assert_war_is_peace())
True
```

It looks like the solution is [auto-speccing](https://docs.python.org/dev/library/unittest.mock.html#auto-speccing). In fact, the documentation even mentions our situation:

> Because mocks auto-create attributes on demand, and allow you to call them with arbitrary arguments, if you misspell one of these assert methods then your assertion is gone.... Your tests can pass silently and incorrectly because of the typo.

foosel:
Ah, that certainly solves that. Thanks for clarifying! Evil trap though...

PR #1016, opened 2015-08-07
MacOS: fix path quirk in filemanager unit tests.

DanLipsitt:
On MacOS, `tempfile.mkdtemp()` doesn't necessarily return a canonical
path (it uses `/private/var` instead of `/var`, but they are the same
thing). Since the `filemanager.storage` module canonicalizes paths,
tests were erroneously failing due to prefix match failures.


foosel:
Yikes, that sounds like something that made you scratch your head for a while... Good finding

PR #1038, opened 2015-08-31
Turn the Com layer into a plugin

Lemmons:
This pull request abstracts out the com layer into a plugin type and adds the default com plugin.

It adds a selection dropdown box to the printer connection UI which allows for selection of a com layer to use, but only if multiple com layer plugins are installed.

This change gives plugin developers the option to use any sort of communication layer with the printer they desire.


foosel:
I know I still owe you a reply via email on this topic, my wrist was still acting up until early last week and then I was swamped with work so I didn't get around to it yet.

First of all, in principle I like the idea of turning the communication layer into something that is customizable through the use of plugins (it's been an idea I've been following with the `commRefactoring` branch as well, although work on that one stalled due to more pressing issues). So what follows is not a categorical "no way" but rather an outline of the problems I see with this exact approach at this exact time.

What I think is that once we establish it like that, there's no going back or changing course (or it's very hard to go back), making a later even more modularized approach - I envision distinct transport and protocol implementations - tricky or maybe even impossible to establish. So in a nutshell, I fear that this might shut a door more down the road if I now merge this. The way it is proposed now still makes everything dependent on serial connections at the core of things. The comm layer definitely does need a refactoring, but when done (and as consequence also exposing it for developing against by plugins) it should ideally also abstract more instead of just taking the current quite narrow minded model of what a printer connection does look like and extracting that into a plugin. The idea would be to basically define a generic transport interface with connection parameters (which could be serial, TCP, UDP, [RFC1149](https://tools.ietf.org/html/rfc1149), whatever you can imagine. And then on the other hand have a generic protocol interface as well which defines HOW things like jogging, temperature control etc are done. And - possibly - even a generic device interface on top of THAT (that's probably further down the road) which defines if something even HAS temperature control. So a connection would consist of a protocol spoken over a transport layer, and probably have its own state machine to make stuff less cluttered with "if ... else ...". You see that the vision is a quite broad one here, which will probably not realized for 1.3 (but maybe 1.4?), since it's quite a change. I'd be happy for any input regarding how to establish something like pluggable versions of the current implementation that won't cause major headaches when that big refactoring takes place however. What I definitely want to avoid is having to painfully maintain backwards compatibility for "the old way" for a long time to come due to plugins using that and their authors not updating them.

Another - more minor concern: You might also have noticed that I'm trying to migrate the naming scheme of documented interfaces more towards fitting the general Python naming scheme (snake_case instead of camelCase). I've been migrating all parts of the system that DO already have a more or less proper documentation and/or are officially exposed that way, and a comm layer plugin certainly should follow that.

Lemmons:
I totally agree, that this sort of change shouldn't be released until you are ready to commit to the change.

One thing that could at least ease the change is creating a way (if it doesn't already exist) for plugins to specify the version of a plugin type they are designed to work for.  That way, if a plugin breaking change is released, the plugin itself will know it probably won't work with the new version.  Not the ideal situation, but I think it would be nice to have some established way to make breaking changes to plugin types which allows refinement of plugin types to continue.  Ideally you would want to keep the plugin types as stable as possible, but this would give you a path to continue to develop them.  For example, we could call this version of the comm plugin type 1.0.  In a few months you break apart the comm into the separate transport and protocol layers.  The comm plugin is now just the transport layer, and the version moves to 2.0, as this is a major, breaking change.  After a user upgrades to the new version of OctoPrint using this newly updated plugin type, a plugin implementation which is versioned to the 1.0 version of this plugin will be disabled as OctoPrint starts-up, and a message will be presented to the user telling them the plugin was disabled because it is no longer supported with this version of OctoPrint.  If there is a newer version of the plugin available, the user will be given the options to then upgrade to it and continue as usually.  Obviously this sort of system would represent a decent change to the plugin system, but would give a path for updates to plugin types.

As for the dependency on serial: This refactor of the comm actually does away with it in a sense, though the printer is still specified using serial.  If you look at the plugin we wrote using this, https://github.com/Authentise/OctoPrint-Authentise, you can see we are actually using a web-api to do all printer communication, no serial at all.  It's a bit of a hack, but totally works.  If the connect function were made a little more generic, so it no longer takes port and baudrate, but instead a dictionary of connection parameters, or something like that, it could be potentially generalized to work with almost any protocol.

As for camelCase vs snake_case, I actually started to change things over as I was making the plugin, but stopped as it was taking too much time(I need to find a better way to change naming across multiple files).  I think it makes total sense to change it to using snake_case.  Honestly I would love to see spaces instead of tabs too, but I think I know your feelings on that subject and am not holding out for that. ;)

markwal:
How about using the interface one level up?  http://docs.octoprint.org/en/master/modules/printer.html#octoprint.printer.PrinterInterface

It still has some assumptions about serial in it (port, baudrate), but it is relatively narrow and PrinterInterface is an already documented thing so has already opened the door to dependencies.

I think @Lemmons could still use the current Authentise MachineCom implementation underneath it with a reimplementation of printer/standard.py over it for the Authentise printer.

Lemmons:
@markwal, I actually started at the printer interface, but that ended up not working well, because I wanted the user to be able to select which printer (or as the case is now, comm) to use at printer connection time (not OctoPrint startup).  If the printer itself is implemented, then you have to select a printer type and restart OctoPrint to use it.

As this is plugin is implemented, you can have as many comm plugins as you want enabled at the same time, and you just select which one to use as part of the printer connection options, which I feel is much more ideal as it has a bit less of an impact on the user when switching which comm to use.

foosel:
I just now realized I completely forgot about this... Awfully sorry about that.

My bad feeling about going down this route hasn't changed sadly. I still haven't gotten around to actually establish the new comm layer architecture as original planned (2016 was kind of crazy thanks to a funding crisis), but I hope to finally change that for 1.4.0. That would make the current comm layer and therefore anything depending on this PR completely obsolete, so merging this simply doesn't mesh with the future plans for the platform. 

Sorry, but I really don't want to merge this for these reasons, it just would cause compatibility and/or maintainability issues down the road.

PR #1051, opened 2015-09-08
Fix checks with os.path.realpath() for systems with symlinked paths

markbastiaans:
I'm working on getting a recent version of Octoprint (1.2.6) packaged for Synology NASes in [SynoCommunity/spksrc](https://github.com/SynoCommunity/spksrc). I've gotten it to work after some tinkering, but it needed two trivial fixes to work with Synology's habit of putting packages in a symlinked location. Specifically, two checks using `os.path.realpath()` fail on Synology's OS while they shouldn't be.

I've not filed a bug report for this yet, let me know if I should. I've tested the fix on Synology only but I think the fix is so trivial that it won't will negatively impact other platforms.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -53,6 +53,7 @@ date of first contribution):
   * [Andrew Erickson](https://github.com/aerickson)
   * [Nicanor Romero Venier](https://github.com/nicanor-romero)
   * [Thomas Hou](https://github.com/masterhou)
+  * [Mark Bastiaans](https://github.com/markbastiaans)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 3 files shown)


foosel:
Thanks for spotting this. Bug ticket is not necessary, I'll put that into the CHANGELOG with reference to the PR. Want me to also add you to the `AUTHORS.md` file while at it or do it yourself? :)

markbastiaans:
I've added my name to the authors file myself. Cheers!

foosel:
Perfect, thanks again!

markbastiaans:
@foosel, would it be possible to merge this PR into the upcoming maintenance release? I'm working on a new version of the OctoPrint package for Synology NASes. Basing that package on a tagged release would make that a lot easier.

Thanks in advance!

foosel:
@markbastiaans Just did. Planning on releasing 1.2.8 monday or tuesday (if nothing horrible happens again...)

markbastiaans:
Thanks @foosel!

PR #1058, opened 2015-09-11
Add errorhandler decorator for the blueprintplugin

punkkeks:
This issue annoyed me a lot, so I decided to solve it by myself. It will easily allow you to add an errorhandler for you blueprint, so you can raise exceptions in you code without having a try-catch everywhere, just use it like the regular route decorator, e.g

```python
@octoprint.plugin.BlueprintPlugin.errorhandler(CustomException)
def api_onError(self, code_or_exception):
    # do whatever you like here
    pass
```

To solve this the other way is to override the get_plugin method and do all the stuff manually. That's really time consuming and won't work.

Ps. you should mention in the docs, that the get_blueprint method is not meant to be called from the plugin code, because it will create a new blueprint, that will have nothing in common with the already registered blueprint. That costs me about half a day ;)
Pss. get your indention under control!!11! Either use all tabs or (to be pep8 conform) use 4 spaces, but currently it is a mess!

### DIFF
--- a/src/octoprint/plugin/types.py
+++ b/src/octoprint/plugin/types.py
@@ -651,6 +651,24 @@ class BlueprintPlugin(OctoPrintPlugin, RestartNeedingPlugin):
 			return f
 		return decorator
 
+	@staticmethod
+	def errorhandler(code_or_exception):
+		"""
+		A decorator to mark errorhandlings methods in your BlueprintPlugin subclass. Works just the same as Flask's
+		own ``errorhandler`` decorator available on blueprints.
+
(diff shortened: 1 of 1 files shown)


foosel:
Nice. 

One thing: Can you please do that PR against the `devel` branch as stated in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests)?  Thanks in advance!

PS: Feel free to file a PR against the docs ;)
PPS: All Python code in OctoPrint is indented with tabs!!1!eleven While I'd love to have the code PEP8 conform, at this point switching the whitespace (and also a bunch of other things like variable names etc) over to that will make `git blame` completely useless.

PR #1059, opened 2015-09-11
Add errorhandler decorator for the blueprintplugin

punkkeks:
for description refer to #1058 

Sorry for the branch thing. I totally forgot that.

### DIFF
--- a/src/octoprint/plugin/types.py
+++ b/src/octoprint/plugin/types.py
@@ -1055,6 +1055,24 @@ class BlueprintPlugin(OctoPrintPlugin, RestartNeedingPlugin):
 			return f
 		return decorator
 
+	@staticmethod
+	def errorhandler(code_or_exception):
+		"""
+		A decorator to mark errorhandlings methods in your BlueprintPlugin subclass. Works just the same as Flask's
+		own ``errorhandler`` decorator available on blueprints.
+
+		See `the documentation for flask.Blueprint.errorhandler <http://flask.pocoo.org/docs/0.10/api/#flask.Blueprint.errorhandler>`_
+		and `the documentation for flask.Flask.errorhandler <http://flask.pocoo.org/docs/0.10/api/#flask.Flask.errorhandler>`_ for more
+		information.
+		"""
+		from collections import defaultdict
+		def decorator(f):
+			if not hasattr(f, "_blueprint_error_handler") or f._blueprint_error_handler is None:
+				f._blueprint_error_handler = defaultdict(list)
+			f._blueprint_error_handler[f.__name__].append(code_or_exception)
+			return f
+		return decorator
+
 	def get_blueprint(self):
 		"""
 		Creates and returns the blueprint for your plugin. Override this if you want to define and handle your blueprint yourself.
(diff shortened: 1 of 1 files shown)


foosel:
No problem. Want to also add yourself to `AUTHORS.md` or should I?

punkkeks:
It is easier if you do it ;) but thanks

btw.
> $ find . -name '*.py' -exec grep ' &nbsp; &nbsp; ' {} \\+ | wc -l
> 3766

foosel:
Ah... you were referring to the alignment whitespace, not the indentation whitespace. That's what's commonly referred to as "smart tabs".

punkkeks:
Sorry, forgot the `^` Then `wc -l` will only print out 2071. But okay. I aggree with you, that the blame will print out nothing acceptable anymore. Thanks for the merge ;)

foosel:
If it helps, I bite myself in the arse for not converting the source to spaces back when I forked from Cura ;) I could bite myself for a *lot* of things that are now hard to change without making debugging more complicated, merging of bug fixes impossible etc or just have a completely ... 

The `^` will btw also capture stuff like:

``` python
class CuraPlugin(octoprint.plugin.SlicerPlugin,
                 octoprint.plugin.SettingsPlugin,
                 octoprint.plugin.TemplatePlugin,
                 octoprint.plugin.AssetPlugin,
                 octoprint.plugin.BlueprintPlugin,
                 octoprint.plugin.StartupPlugin):
```

which again is a case of smart tabs at work and 

``` python
deprecated = warning_decorator_factory(DeprecationWarning)
"""
A decorator for deprecated methods. Logs a deprecation warning via Python's `:mod:`warnings` module including the
supplied ``message``. The call stack level used (for adding the source location of the offending call to the
warning) can be overridden using the optional ``stacklevel`` parameter. If both ``since`` and ``includedoc`` are
provided, a deprecation warning will also be added to the function's docstring by providing or extending its ``__doc__``
property.

Arguments:
    message (string): The message to include in the deprecation warning.
    stacklevel (int): Stack level for including the caller of the offending method in the logged warning. Defaults to 1,
        meaning the direct caller of the method. It might make sense to increase this in case of the function call
        happening dynamically from a fixed position to not shadow the real caller (e.g. in case of overridden
        ``getattr`` methods).
    includedoc (string): Message about the deprecation to include in the wrapped function's docstring.
    extenddoc (boolean): If True the original docstring of the wrapped function will be extended by the deprecation
        message, if False (default) it will be replaced with the deprecation message.
    since (string): Version since when the function was deprecated, must be present for the docstring to get extended.

Returns:
    function: The wrapped function with the deprecation warnings in place.
"""
```

which is a valid napoleon docstring. 

However, that high number you saw there got me curious, so I just went through the output of `$ find . -name '*.py' -exec grep '^    ' {} \+ | less` in `./src/` where it got me 382 lines (not 2071, in which folder did you execute that?) and fixed the files with hits that were indeed not either smart tabs at the start of the line or generated (`src/octoprint/_version.py`) in commit 2c0eed266ceadeb3cd9e263d2170b9e118ce726a.

punkkeks:
Stupid me: in the root of octoprint of course :D something called versioneer.py has 1371 of the 2071 indentions. When i use the src/octoprint path, I will get 682, which is higher than your number though. Nevermind. Keep on the good work

foosel:
Yeah, `versioneer.py` is a) pep8 code and b) what generates the `src/octoprint/_version.py`. You really tickled my curiousity now why you see 300 more hits, but I guess I should try to refrain from diving deeper now and get back to pounding the javascript code into shape ;)

PR #1066, opened 2015-09-17
changed magic line from python->python2

punkkeks:
Since python is sometimes (and for me on arch linux) a symlink to the
python3 utils, you should be more explicit with that.

### DIFF
--- a/run
+++ b/run
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import os
 import sys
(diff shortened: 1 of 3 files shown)


foosel:
Thanks! 

One thing, if you happen to do other PRs in the future, please make sure to do them against a PR specific branch on your end, not just your version of devel (see also the steps for PRs in the contribution guidelines). Should only be one more git command or click here on github but removes the risk of you accidentally pushing other unrelated changes into the PR by pushing to your devel branch, making it unmergable.

PR #1074, opened 2015-09-28
Add superscript to trademark

MolarAmbiguity:


foosel:
and thanks for that too! ;)

PR #1099, opened 2015-10-28
Added HTML id's to control tab panels

BillyBlaze:
1. Add'd new HTML id's
1. Moved all HTML id's to the beginning of the tag
1. Wrapped feed rate slider/input into a div with jog-panel class and new id


foosel:
Ah, darn... I couldn't look into the code when we discussed this earlier on IRC or I'd noticed that earlier, would you mind changing the ids from "controls-..." to "control-..." to match the ids already existing there? Sorry, that was my mistake...

BillyBlaze:
No problem :smile:

foosel:
Perfect, thanks!

BillyBlaze:
Your welcome

PR #1102, opened 2015-10-29
Fix error in /api/job example

mtowara:
Examples seem outdated, repalaced POST /api/control/job with /api/job

### DIFF
--- a/docs/api/job.rst
+++ b/docs/api/job.rst
@@ -37,7 +37,7 @@ Issue a job command
 
    .. sourcecode:: http
 
-      POST /api/control/job HTTP/1.1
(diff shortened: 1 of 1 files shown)


foosel:
Thanks, but that was actually already fixed in the `devel` branch ([where all PRs should go against](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests) for this very reason) in 119f3ba45bf59e4d7304098731d546b2ea3ce44d and just didn't yet make it into `master`.

foosel:
Original commit however is now cherry picked on `maintenance` branch and will go into the `1.2.8` release.

PR #1103, opened 2015-10-30
Prevent cutoff of Flot's y-axis markers when using responsive CSS

kevingelion:
When using the Bootstrap responsive CSS manually or the ScreenSquish plugin, the y-axis markers get cutoff. This simple change prevents conflict.

image


foosel:
Merged, but onto ``devel`` and ``maintenance`` instead of ``master``.

``master`` only gets updated on releases, which is why [PRs should always go against the development branch](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests) (also to make sure that the issue is not already fixed there).

PR #1104, opened 2015-11-02
Allowed camera rotation by any multiple of 90 degrees.

eddieparker:
My camera is in an awkward position as I ran out of cable.  Rather than futzing with hardware, I wanted the option to play with it in software.

Note that I changed the default meaning of 'rotate90' to be 'rotate 90 clockwise'.  Perhaps it's subjective, but it confused me that the default 'rotate90' was actually rotate270 clockwise.

Feedback appreciated.


foosel:
Thanks!

In principle the same could already be achieved by a combination of the flip and rotate options:

  * rotate by 90° counter clockwise (270° clockwise): `rotate90`
  * rotate by 180°: `flipV` + `flipH`
  * rotate by 270° counter clockwise (90° clockwise): `rotate90` + `flipV` + `flipH`

I agree though that that is not very intuitive, so adding explicit options to rotate by the multiples of 90° indeed does make sense. Same goes for switching the description to clockwise rotation, considering that although the mathematical rotation direction is counter clockwise, humans fare way better with clockwise.

The problems I see with the proposed patch (apart from it being against the wrong branch and from your own version of `master` instead of a feature branch, [please do all pull requests against the `devel` branch from a dedicated feature branch for your pull request](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests)) are the following:

  * It removes an existing configuration setting (`rotate90`) without a migration scenario to the new one (`rotation`). So people who currently have their camera options set to `rotate90` will suddenly find their camera facing the wrong way and will have to reconfigure it all over.
  * It introduces a problem for people who currently do the rotate + flip combinations (see above) to achieve the desired rotation and also for how to handle the flip options in the future. We probably both need to migrate the flip combos to angle interpretations as well during a migration and also make the flip options themselves more like radio buttons ("not flipped", "flipped horizontally", "flipped vertically") in order to prevent them colliding with selected rotation values.
  * Finally I see a problem with the valid rotation options themselves being stored in the default configuration dictionary. That way people could actually override their values via `config.yaml` - I'm fairly sure that is not what is intended to be possible here (and if it was, I'd vote against it ;)).

So what needs to be done with this PR before it can be accepted is:

  * create a new feature branch off `devel` explicitely for this patch: `git checkout devel && git checkout -b dev/rotationOptions`
  * migrate the existing settings in `src/octoprint/settings.py`:
    ``` python
    def _migrate_rotation_options(self):
        if "webcam" in self._config and "rotate90" in self._config["webcam"]:
            rotate90 = self.getBoolean(["webcam", "rotate90"])
            flipV = self.getBoolean(["webcam", "flipV"])
            flipH = self.getBoolean(["webcam", "flipH"])

            rotate = "rotate0"
            if rotate90 and flipV and flipH:
                rotate = "rotate90"
            elif flipV and flipH:
                rotate = "rotate180"
            elif rotate90:
                rotate = "rotate270"

            self.setBoolean(["webcam", "rotate"], rotate)
            if flipV and flipH:
                self.setBoolean(["webcam", "flipV"], False)
                self.setBoolean(["webcam", "flipH"], False)
    ```
  * Remove the `rotationOptions` from the default settings, keep them only in the frontend
  * Turn the checkboxes for `flipV` and `flipH` into bootstrap radio buttons or a drop-down, something like "Flip:" with options "No", "Horizontally", "Vertically"

Do you want to do that (and have the time to do that)? Otherwise we can do it, just need to know :)

eddieparker:
Thanks for the feedback Gina.  I'll see if I get time to incorporate your feedback this weekend and submit a new pull request.  If you guys get to it sooner (I imagine you have more important stuff on your plate), so much the better. :)

Only semi related: I'm working on hacking together new Cura+GPX support as my priority, as I'm trying to switch over my Flashforge to using Octoprint, so only if that bodes well will I probably get around to resubmitting this in the proper manner.

jneilliii:
Didn't Mark do the GPX plugin already?

https://github.com/markwal/OctoPrint-GPX

eddieparker:
Sorry, I was unclear - I'm using Mark's stuff, and as far as I have been able to verify, the GPX stuff is likely working, it's the Cura integration that's breaking for me.  If you want more information, here's my rough scribbled notes on what's next:

The investigation I've done this far is that the Octoprint-Cura plugin seems to try sending a -v parameter to the cura.sh script, which isn't a valid parameter anymore.  Getting past that, it looks like it's sending a mangled version of the profile across too.  I've started skipping that section of code and just sending the stuff through using the script parameter and it *seems* to be working.  My only issue now is that it seems to disregard the settings I set in gcode.start and stomps them with the 'default' from the plugin.  So I'm parsing through the code to see why that's necessary (feedback is appreciated), or if it's not anymore.

foosel:
Careful, the built-in cura plugin expects you to use it with the CuraEngine binary directly, *not* the full Cura install including cura.sh.

MertMG:
is there anyway to rotate camera for Streaming? i`ve tried lots of stuff but couldn`t achieve it. Also is anyway to create a zoom option?

markwal:
@MertMG in OctoPrint itself, there are the rotation and flipping options mentioned in this bug which does rotate the camera image for the stream you see in the OctoPrint UI.  Settings->Webcam/Timelapse checkboxes.  There's no zoom however.

In some cases you can get what you want upstream from OctoPrint.  For example, some USB camera drivers have command line control options that you can modify (you'll have to Google for your camera and raspberry pi, I think v4l2 might help).  And mjpg-streamer has some options for it's capture modules that are documented here https://github.com/foosel/OctoPrint/wiki/MJPG-Streamer-configuration (see -rot near the bottom).

MertMG:
@markwal i`ve read and applied required commands for mjpg-streamer but they didn`t work.. i`m sure there is something i missed out but believe me i tried several way to fix it. I can modify resolution and fps but i cant get a rotated view for some reason when i`m streaming..

markwal:
@mertmg are you using a raspicam or Usb device?

MertMG:
it is a usb(logitech c270). i`ve tried to activate usb option but as i told before i can play with resolution but not working for rotation.

markwal:
@eddieparker, I've tried out the CuraEngine plugin and I believe I have a repro of your start.gcode problem. I think it isn't GPX related, but rather dual nozzle. I'll enter another issue for that.

markwal:
That's right. mjpg-streamer's USB module can't do rotation. That's why that wiki page only lists -rot for raspicam.  So your only available option at this point is to use the checkboxes in OctoPrint settings which only rotates in the browser UI. Unless you want to write some code.

MertMG:
argh :(

foosel:
@eddieparker are you still looking into this?

eddieparker:
@foosel Sorry Gina, I am not.  I switched from OctoPrint, so I don't need this feature anymore.  Apologies.

foosel:
Ok, closing as abandoned then

Issue #705, opened 2015-01-04
[Request] Manual start of a timlapse video


mlkoch:
I have not found a way to manually trigger a timelapse video.
I use a UM2 and always print using the printer's display-interface and sd-card.
Manually triggering a timlapse-video would be very helpful!

Best regards!
Michael

smsagent:
Hello,
I would like to add this feature too - while printing from SD card, only short still image of final print is made. It should be fine to have both options - timed and on Z axis change. I think, that OctoPi should read axis position while printing from SD card.

Regards!
Ondrej

foosel:
> I think, that OctoPi should read axis position while printing from SD card.

OctoPrint. OctoPi is the Raspberry Pi image. And no, it shouldn't. It would have to constantly query the data from the printer in order to achieve this and STILL would be off. That would lead to the printer firmware having to constantly answer queries (and hence risk slowing down processing) without adding real value. It just doesn't make sense to do that.

mlkoch:
I just thougt of a function to trigger the timlapse video by hand.
The trigger for the pictures needed for the video could be done by time (e.g. 120 sec.). Trigger by z-axis is not necessary.
That's all.
Best regards
Michael

foosel:
Duplicate of #452

PR #1124, opened 2015-11-27
Option for a printer triggered time-lapse on custom gcode

MoonshineSG:
The following changes are needed to marlin firmware to trigger
action:snapshot

```
case 808: //M808: send action camera trigger
  gcode_M808();
  break;
```
before "case 999: // M999: Restart after being Stopped"

and
```
 /**
  * M808: send action message
  */
inline void gcode_M808() {
  st_synchronize();
  refresh_cmd_timeout();
  SERIAL_PROTOCOLPGM("//action:snapshot\n");
}
```
before the implementation of "gcode_M999()"

Add the following to the layer change script (if your slicer supports
it):

G10 ;retract
G1 X0 Y100 F6000 ;move to fixed position
M808
G11 ;unretract

Note: M808 has been chosen as it is currently unused but any other
unused gcode would work just fine.

On Repetier shoudl be smth like 

https://github.com/repetier/Repetier-Firmware/blob/master/src/ArduinoAVR/Repetier/Commands.cpp#L1508

```
        case 808: //M808 trigger snapshot
          Commands::waitUntilEndOfAllMoves();
          Com::printFLN(PSTR("//action:snapshot"));            
          break;
```

but this is untested...


foosel:
First of all, thank you, that's a feature that definitely makes sense.

What I have a bit of a problem with is that it forces a new bundled plugin into the core system but does the extension of the timelapse system to support the new type in a very static kind of way. What do you think about going one step further here and making timelapses extendable in general? Basically introducing a new plugin mixin type "Timelapse Plugin" which allows to return a new timelapse type, which the existing timelapse system then just registers and offers up for configuration? 

That would allow the actual processing of the `action:snapshot` command to be completely encapsulated within a plugin and would remove the necessity to keep it within the core application (in general, I'm trying to reduce the amount of bundled functionality in favor of a more "select what your need" kind of approach).

And it also would allow other people to add more diverse timelapse configurations (like making snapshots on external triggers or manual timelapsing through a plugin, which would be interesting for #705 for example)

Thoughts?

MoonshineSG:
Sure. I did this as a "quick fix" this afternoon and I thought it would be worth sharing.... I will check out mixing and making a proper plugin... Any hints to jump start me ?

foosel:
Plugin types are defined in [octoprint/plugin/types.py](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/plugin/types.py) and also need to be added in [octoprint/plugin/__init__.py](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/plugin/__init__.py#L99) (I really need to get around doing this a bit more dynamically...).

Then you can collect all plugins of any type with [PluginManager.get_implementations(...)](http://docs.octoprint.org/en/devel/modules/plugin.html#octoprint.plugin.core.PluginManager.get_implementations) and utilize them to perform your functions.

I'd suggest defining something like this for the plugin:

``` python
class TimelapsePlugin(octoprint.plugin.OctoPrintPlugin, octoprint.plugin.RestartNeedingPlugin):
    def get_timelapse_types(self):
        return dict()
```

and then adjust [octoprint/timelapse.py](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/timelapse.py) to something like this:

``` python
def get_timelapse_types():
	result = dict(zchange=ZTimelapse, timed=TimedTimelapse)

	timelapsePlugins = octoprint.plugin.plugin_manager().get_implementations(octoprint.plugin.TimelapsePlugin)
	for name, implementation in timelapsePlugins.items():
		try:
			timelapse_types = implementation.get_timelapse_types()
		except:
			logging.getLogger(__name__).exception("Error while retrieving timelapse information from plugin {}".format(name))
			continue
		else:
			for identifier, timelapse_type_info in timelapse_types.items():
				if not isinstance(timelapse_type_info, dict) or not "class" in timelapse_type_info or not "display" in timelapse_type_info:
					continue
				result[identifier] = timelapse_type_info

def get_timelapse_info():
	available = getTimelapseTypes()
	return {key: value["display"] for key, value in available.items()}

def create_timelapse_instance(type, post_roll, fps, **options):
	available = getTimelapseTypes()
	if not type in available:
		return None
	try:
		return available[type]["class"](postRoll=postRoll, fps=fps, **options)
	except:
		logging.getLogger(__name__).exception("Error constructing instance of {}".format(type))

def configureTimelapse(config=None, persist=False):
	# TODO should be renamed to configure_timelapse, but we'll leave that out now
	global current

	if config is None:
		config = settings().get(["webcam", "timelapse"])

	if current is not None:
		current.unload()

	type = config["type"]

	postRoll = 0
	if "postRoll" in config and config["postRoll"] >= 0:
		postRoll = config["postRoll"]

	fps = 25
	if "fps" in config and config["fps"] > 0:
		fps = config["fps"]

	current = create_timelapse_instance(post_roll=postRoll, fps=fps, **config.get(options, dict()))

	notifyCallbacks(current)

	if persist:
		settings().set(["webcam", "timelapse"], config)
		settings().save()
```

That gives us the registration. But we also need to tell the frontend about what timelapse types we support. So what you'll also need to do is add the info from `get_timelapse_info` to the API used for retrieving the timelapse data, so that we can display it in the frontend. That stuff is defined [here](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/server/api/timelapse.py#L26) and I'd suggest simply adding another field `types` to the returned response and return the data from `get_timelapse_info` there. And the finally the frontend [javascript](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/static/js/app/viewmodels/timelapse.js) and [jinja template](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/templates/tabs/timelapse.jinja2) code needs to be adjusted to also allow selecting other timelapse types. You basically want to take the info regarding the available types from the backend here and make them selectable. I suggest to leave custom configuration via the frontend out for now, that would need quite some changes.

You'd then define a new timelapse time something like this:

``` python
class MyTimelapse(octoprint.timelapse.Timelapse):
	# ...
```

and have your plugin register it with the system like this:

``` python
class MyTimelapsePlugin(octoprint.plugin.TimelapsePlugin):
	def get_timelapse_types(self):
		return dict(my_timelapse=dict(class=MyTimelapse, display="My Timelapse")
```

Note: This is only a rough sketch, basically outlining a suggestion how I'd tackle it. There might be some places where I didn't think of a problem that you'll encounter when trying the above (might already start with syntax error, I only typed that down here ;)). In any case, feel free to just ask any questions you might have, I'll try to answer them as fast as possible. Also, you maybe might want to checkout our IRC channel (#octoprint on freenode), there a quite some people in there who have some OctoPrint coding experience under their belt.

MoonshineSG:
This might be too much for me to chew on... I'll leave it like this for now and if later on someone creates a TimelapsePlugin type I would change the codes... 
Meanwhile I'll leave it like this... 

Pull it in or not, it's up to you...

PR #1133, opened 2015-12-04
Added brazilian portuguese language

YanFM:


foosel:
Thanks, but I don't bundle languages that I cannot update myself, which is why OctoPrint supports the uploading of language packs. See also [this section in the Translation guide](https://github.com/foosel/OctoPrint/wiki/Translating-OctoPrint#creating-a-language-pack).

The problem is simply that I'd rather not have users look at a semi translated or badly translated version in their own language since that severely impacts the user experience - and I can't do the regular updating and quality assurance of the translation files of languages I don't speak myself, hence currently only english and german is bundled and the rest should be added via separately installed language packs.

Do you need assistance in creating a language pack from your installation?

PR #1155, opened 2015-12-23
All filters selected by default

MoonshineSG:
By default, select all available filters in the terminal tab. No more annoying re-selection... :D

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/terminal.js
+++ b/src/octoprint/static/js/app/viewmodels/terminal.js
@@ -244,6 +244,15 @@ $(function() {
             }
         };
 
+        self.onAfterBinding = function(){
+            allFilters = [];
+            _.each(self.filters(), function(entry) {
+                allFilters.push(entry.regex);
+            });
+            self.activeFilters(allFilters);
+        }
+    }
+
     }
 
     OCTOPRINT_VIEWMODELS.push([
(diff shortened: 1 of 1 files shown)


foosel:
Hm, not too happy with this, since what is annoying for you might be what others expect with regards to behaviour. How about turning this into a plugin or making it use the browsers local storage instead to remember the last settings (for that there is actually a feature request if I remember correcly)?

Also currently on a vacation and really need to take a break, so no reviewing until after new years, hope you understand :)

MoonshineSG:
I don't know. In my opinion if you set filters is because you don't want to see smth. Anyway, a quick workaround would be to have a setting "default on/off" 
A plugin would be the same, with the option to turn it on/off.. 
I'll be back after holidays as well and think about options.

MoonshineSG:
I found an easy way to save the status of the filters and build it into a plugin (https://github.com/MoonshineSG/Octoprint-ActiveFilters/archive/master.zip)

foosel:
@MoonshineSG Why don't you also [register your plugin on OctoPrint's plugin repository](http://plugins.octoprint.org/help/registering/)? That way other people can find it easier :)

MoonshineSG:
Hmmm... I didn't know you can do that :D I'll have a read but looks like a lot of steps... Most likely I'll mess it up...

PR #1159, opened 2015-12-25
Add $data.type as className for better styling to the new files list

BillyBlaze:
It would be helpful if we can make a distinction between entries. As we have ``$data.type`` we can simply add that. I have manually added "back" to the back button.


foosel:
Thanks for this! Currently on vacation, will review and merge after new years, hope you understand :)

PR #1160, opened 2015-12-26
Allow octoprint.comm.protocol.gcode.queuing hook to return a list of commands

mrvn:
The M3D Micro plugin comes with a Z compensation that replaces single G1 commands with a series of smaller G1 commands with adjusted Z value. For this to work cleanly the octoprint.comm.protocol.gcode.queuing hook needs to return a list of commands.

The pull request theoretically also allows the octoprint.comm.protocol.gcode.sending hook to return a list of commands but at that point the line number is already chosen and expanding a command would mess that up. Therefore the code asserts that commands from sending are not expanded to multiple commands and only documents the new feature for queuing. This breaks backward compatibility for plugins returning a list instead of a tuple in their octoprint.comm.protocol.gcode hooks. But the API documentation clearly says to return a tuple.

Note: This is probably not ready for merging yet but it's late. Consider this a request for review and broader testing please. It works for me for simple cases but has the potential to break a lot of plugins.


mrvn:
Rebase to devel branch pending...

mrvn:
github wouldn't let me rebase the pull request so I refiled a clean one

PR #1162, opened 2015-12-27
Allow octoprint.comm.protocol.gcode.queuing hook to return a list of commands

mrvn:
The M3D Micro plugin comes with a Z compensation that replaces single G1 commands with a series of smaller G1 commands with adjusted Z value. For this to work cleanly the octoprint.comm.protocol.gcode.queuing hook needs to return a list of commands.

The pull request theoretically also allows the octoprint.comm.protocol.gcode.sending hook to return a list of commands but at that point the line number is already chosen and expanding a command would mess that up. Therefore the code asserts that commands from sending are not expanded to multiple commands and only documents the new feature for queuing. This breaks backward compatibility for plugins returning a list instead of a tuple in their octoprint.comm.protocol.gcode hooks. But the API documentation clearly says to return a tuple.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -56,6 +56,7 @@ date of first contribution):
   * [Mark Bastiaans](https://github.com/markbastiaans)
   * [Marcel Hellwig](https://github.com/punkkeks)
   * [Kevin Murphy](https://github.com/kevingelion)
+  * [Goswin von Brederlow](https://github.com/mrvn)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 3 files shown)


foosel:
@mrvn Thanks for that! No worries about the backward compatibility there, I clearly stated that plugins should return tuples, not lists. I shifted things around a bit to remove a bit of repetition (basically reusing the single entry parsing for the lists as well) and added a bit of sanity checking during phase processing for the hooks to log an error if anything but the ``queuing`` phase returns multiple entries. Also added a small example to the docs.

Could you take a look at the [`pr/mrvn/feature-multiple-gcodes` branch](https://github.com/foosel/OctoPrint/tree/pr/mrvn/feature-multiple-gcodes) and check if I broke anything in the process?

foosel:
@mrvn Still waiting for feedback here :)

foosel:
Forgot to update this... The adjustment mentioned above was merged/cherry-picked to `maintenance` a while ago and will be part of the 1.3.5 release.

foosel:
1.3.5 was released yesterday.

PR #1172, opened 2016-01-02
Intermediary page: Incorrect image url with hashtags and different reload method

BillyBlaze:
Yet Another PR... :+1:

1. __Prevent hash tag from generating incorrect image url__
If you start with the Intermediary page and you have a hashtag in your url then the script would make it ``http://localhost:5000/#touchonline.gif``. This would also make the start screen show an incorrect error. (Because it failed to load the image which the normal url would have responded on.)

1. __Use reload method to leave url intact__
In addition to removing the hashtag, I also used location.reload to keep the URL intact, e.g. If someone  made a plugin that would remember the tab state by adding hashtag to the URL.

### DIFF
--- a/src/octoprint/static/intermediary.html
+++ b/src/octoprint/static/intermediary.html
@@ -132,7 +132,7 @@
                 var intervals = [1, 1, 2, 3, 5, 8];
                 var timeout = 1500;
 
-                var baseUrl = window.location.href;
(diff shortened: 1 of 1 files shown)


BillyBlaze:
Ahh never mind I see you fixed this already on the maintenance branch, you might want to consider point 2 through.

PR #1177, opened 2016-01-04
Fix missing kwargs in PluginSettings remove

richjoyce:
Trying to run `self._settings.remove(path)` in a plugin caused the following error before:
```
File "OctoPrint/src/octoprint/plugin/__init__.py", line 475, in __getattr__
    settings_name, args_mapper, kwargs_mapper = self.access_methods[item]
ValueError: need more than 2 values to unpack
```

This simply adds the third `kwargs_mapper` value as a passthrough lambda function, allowing remove to function as intended.


foosel:
Thanks for spotting this.

Manually merged on `maintenance` and `devel` branch - `master` only gets updated on new releases (it really is stable).

PR #1183, opened 2016-01-10
Refresh button at Files list windows

guigashm:
Hello,

Is there a way to refresh the files list window, or add a button?

I activated the samba on my Octopi, and sometimes I use CURA to slice and generate Gcode, and then I save the gcode directly to the upload folder on the Octopi.

What happens is that the files list is not updated and I have to Refresh the browser ;((.

Refresh the browser, is to slow, because Octoprint,  will redo the connect to server and login..  and it takes a while.

thank you


foosel:
I'm fairly sure you didn't intend to create a Pull Request...

PR #1198, opened 2016-01-25
allow multiple commands in terminal

MoonshineSG:
allow comma separated commands in terminal tab.

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/terminal.js
+++ b/src/octoprint/static/js/app/viewmodels/terminal.js
@@ -182,20 +182,24 @@ $(function() {
             }
 
             var re = /^([gmt][0-9]+)(\s.*)?/;
-            var commandMatch = command.match(re);
-            if (commandMatch != null) {
-                command = commandMatch[1].toUpperCase() + ((commandMatch[2] !== undefined) ? commandMatch[2] : "");
-            }
+            var commands = command.split(",")
+            
+            $.when.apply($, $.map(commands, function(command) {
+                var commandMatch = command.match(re);
+            
+                if (commandMatch != null) {
+                    command = commandMatch[1].toUpperCase() + ((commandMatch[2] !== undefined) ? commandMatch[2] : "");
+                }
 
-            if (command) {
-                OctoPrint.control.sendGcode(command)
-                    .done(function() {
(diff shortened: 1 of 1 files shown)


foosel:
What if you want to send "M117 foo, bar"?

MoonshineSG:
Yes, problem. I don't use 117 so I didn't think about it. 
Suggestion for another separator or just drop the whole thing...?

foosel:
The only safe separator sadly is newline, and that won't work with a single-line input. Multi-line input would be too much wasted screen space if you only want to entry single commands however, which should be the case for most people considering that for multiple lines you have to enter repeatedly a configured custom button should be better suited.

Let's try to figure out the problem you are solving here. Is it more about being able to copy-paste multiple commands easily? Or more about being able to enter multiple commands quickly? If the latter, why does just hitting "Enter" (which will send the command) not work for that?

MoonshineSG:
It's more like sending the same sequence of commands multiple times (but not enough to justify the creation of a custom control)
Maybe it's just me... and that's ok.

foosel:
Been thinking about this the past days, I currently don't see a good solution to make this work without introducing incompatibilites (e.g. no comma in `M117`).

So not going to merge this, sorry. Maybe someone comes up with THE solution for the problem of allowing both full fledged inputs for regular commands as well as multiple commands without overloading the UI, but right now I'm drawing a blank sadly :/

nophead:
Why not just ignore comma after M117? That is backward compatible. You just
don't have the new ability to have other commands after M117, which is no
great hardship.

foosel:
Yeah, that works as long as you are only thinking about people sending a limited and known set of GCODE through that input field. Then Smoothie comes along with additional in-band commands that aren't GCODE which you have to recognize there too (that is already a fact). Then $awesomeNewFirmware comes around the block and then you have even more you have to think about there. Then suddenly you need to have a user configurable set of command regexes after which to please ignore a comma as separator and the whole parsing takes longer than just hitting enter a couple of times in between. I'm exaggerating a bit here of course but you see where this is going I guess. As soon as you have one exception in something that is not standard bound, you can have a rat's tail of other things you need to consider so people don't start running down your door because "this worked in the last version and now it's broken!".

PR #1215, opened 2016-02-10
Typo in git_commit.py

geoporalis:
found a little typo in git_commit.py/68.  return_code -> returncode
Don't think it's intentional.


foosel:
Definitely unintentional, thanks for spotting this!

PR #1223, opened 2016-02-13
Add python-dev and virtualenv to dependencies

crazyquark:
Yes, this a bit obvious but it might help beginners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ OctoPrint currently only supports Python 2.7. You will also need the python-dev 
 
     sudo apt-get install python-dev
 
+If you are using virtualenv as recommended, you will aso need to
+
+    sudo pip install virtualenv
+    
 Usage
 -----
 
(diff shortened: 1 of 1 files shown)


foosel:
Thank you, but the instructions in the README are supposed to stay platform agnostic - OctoPrint can run on way more than just a raspberry pi :) - and really only give an overview about the generic steps, with more specific guides tailored to specific operating systems available (and linked to) on the wiki. The [specific Raspbian guide](https://github.com/foosel/OctoPrint/wiki/Setup-on-a-Raspberry-Pi-running-Raspbian) already contains the python-dev and virtualenv install instructions in the "Basic Setup" section.

You might also want to [take a look at the pull request section in the contribution guide](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) which is linked on top of every pull request form - your second PR wouldn't have been necessary since the commits you pushed to your version of master were automatically included in it, which btw is the reason why you should always create a specific branch for your PR to prevent this from happening accidentally. Also, all PRs should be done against the devel branch - the reasons for that can also be found in the guide :)

crazyquark:
Alright :)

PR #1244, opened 2016-02-27
[Brainstorming] Added a new functionality to warn user if what is going to be printed exceeds the printing area

Javierma:
I have added to the code a new functionality to warn the user in case that what is going to be printed exceeds the printing area. It checks some commands (G0 and G1) from the gcode and in case of detection it will warn the user.

Folder docs-tfg can be ignored, as it is not part of the functionality


foosel:
Adding to what the bot pointed out, your PR in fact does contain not only commits that do not relate to your described change (slic3r integration? that should btw be done through a separate plugin...), but also your whole python virtual environment (you should never check that in!) plus a folder you already pointed out yourself as "not belonging" (why is it there if it doesn't belong?). A first look at the probably relevant sections of the code also reveals a lot of commented out lines and indentation issues.

I'd love to give your changes a proper look and if possible include them, but it is impossible to do a review on the current change set, there's just too much there that doesn't belong. Could you please extract the changes that do what your PR's description contains into a new branch based off of `devel` instead, remove unneeded clutter (commented out code), make sure you have a uniform indentation and do a PR for that? That would be awesome! :)

Javierma:
Thanks Gina. I can see that I made some mistakes like adding lots of commits when I really should have done (an wanted indeed) only one. I will try to fix it during the next days (I am still quite new using GitHub as it can probably be seen).

About Slic3r integration, this is just a comment in a file as it is part of my Final Degree Project so my tutor can trace in which point I am. I already did a fork and I am working on it separately as it is a plugin (hopefully it will be ready soon as I made lots of advances)

foosel:
No worries, mistakes are part of learning. I'll close the PR if you don't mind since it sadly can't be merged, but you can still comment here and I'll look forward to a cleaned up version :)

It is no problem to have additional stuff committed in your own fork. There only is a problem when you do a PR that also contains these unrelated things. A PR or any git merge (what a PR basically boils down to) is an all-or-nothing kind of deal - either I accept all of the commits and file changes within the commits and merge them, or I don't, I can't just pick the tiny parts that are relevant. That's why you always have to make sure your PR really only consists of (ideally one) atomic commits that only contain the specific change your PR is about, nothing else.

The way github implemented PRs also makes it so that a PR is not from your source branch at the point in time you made the PR, but it really is a PR against your source branch, meaning anything you put in there before or after the PR that differs from the target branch you are doing it against will be part of the PR. That is what happened here with all those slic3r related commits above and is the reason why I usually don't accept PRs from something that is not an obvious feature branch only created for the PR , stuff like that happen way too easily and basically make the PR unmergable.

As general rule of thumb, the best course of action for a PR is usually to create a new feature branch off the upstream branch (so OctoPrint's `devel` branch here)

    git add remote upstream https://github.com/foosel/OctoPrint
    git fetch upstream
    git checkout -b area_warning upstream/devel

making and testing your changes on that, committing them, then pushing the feature branch to your repo

    git push -u origin area_warning

and doing the PR from that branch. You can then also merge the branch on your own (different) version of `devel`.

When you get review feedback you can then easily fix any issues by working on your feature branch and pushing to that, keeping the PR as minimal as possible. 

Why the focus on as little and atomic commits as possible? Because commit logs that consist of "added feature", "oops, forgot something", "fixed a bug with my feature" make reviewing tricky and add unnecessary noise to the history, and also make reverting or cherry picking unnecessary complicated.

Hope this helps a bit, and good success with your Final Degree project!

PR #1258, opened 2016-03-05
Add output in Fahrenheit as well as Celsius.

therealbstern:
For Americans who grew up using Fahrenheit, it's difficult to glance at temperatures in Celsius and decide if it is safe to touch the hotend or the bed. This adds temperatures in Fahrenheit in parenthesis after the temperature in Celsius.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -60,6 +60,7 @@ date of first contribution):
   * ["bwgan"](https://github.com/bwgan)
   * [Siim Raud](https://github.com/2ndalpha)
   * ["geoporalis"](https://github.com/geoporalis)
+  * [therealbstern](https://github.com/therealbstern)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -207,6 +207,7 @@ default_settings = {
 		"color": "default",
 		"colorTransparent": False,
 		"defaultLanguage": "_default",
+		"showFahrenheitAlso": False,
 		"components": {
 			"order": {
 				"navbar": ["settings", "systemmenu", "login"],
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -413,7 +413,7 @@ function cleanTemperature(temp) {
 
 function formatTemperature(temp) {
     if (!temp || temp < 10) return gettext("off");
-    return _.sprintf("%.1f&deg;C", temp);
+    return _.sprintf("%.1f&deg;C (%.1f&deg;F)", temp, temp * 9 / 5 + 32);
 }
 
 function pnotifyAdditionalInfo(inner) {
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -411,9 +411,13 @@ function cleanTemperature(temp) {
     return temp;
 }
 
-function formatTemperature(temp) {
+function formatTemperature(temp, showF) {
     if (!temp || temp < 10) return gettext("off");
-    return _.sprintf("%.1f&deg;C (%.1f&deg;F)", temp, temp * 9 / 5 + 32);
+    if (showF) {
+        return _.sprintf("%.1f&deg;C (%.1f&deg;F)", temp, temp * 9 / 5 + 32);
+    } else {
+        return _.sprintf("%.1f&deg;C", temp);
+    }
 }
 
 function pnotifyAdditionalInfo(inner) {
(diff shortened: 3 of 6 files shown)


foosel:
That adds some unnecessary clutter though for the rest of the world who grew up with the metric system. I see the reason though... 

Tell you what, make it something that can be toggled in the settings (defaulting to off) and I'll merge it - having the Fahrenheit value displayed next to the Celsius might actually help in getting a feel for the "metric world" which is after all standard in 3d printing. 

But I really don't want to force Fahrenheit on the rest of the world that doesn't happen to be located in the US and was thus raised with that unit, it is completely meaningless for us.

therealbstern:
Can't ask for fairer than that.  I'll update this thread when I've done that.

Loup911:
I always had to live with the metric and imperial. Maybe a display option could be activated for display on the website.
The convertion of formula is relatively simple.
[° F] = [C] + 9/5 x 32

therealbstern:
I'm not having trouble with the formula, I just can't figure out how to get the pref from the features object in the code that does the temperature formatting.

markwal:
@therealbstern since this is a helpers function, instead of making it get the settings perhaps an easier way would be to make a new helper that includes Fahrenheit on a boolean parameter.  Call the new helper from the old with "false" and then fix your particular caller [temperature.js](https://github.com/foosel/OctoPrint/blob/dedadbc9ac0305799e94ae279d3bca131629c4c5/src/octoprint/static/js/app/viewmodels/temperature.js#L249-L250) to retrieve the settings (it has a pointer to the settings object already) and pass the boolean based on that.

```
function formatAndTranslateTemperature(temp, translate) {
    if (!temp || temp < 10) return gettext("off");
    if (!translate)
        return _.sprintf("%.1f&deg;C", temp);
    return _.sprintf("%.1f&deg;C (%.1f&deg;F)", temp, temp * 9 / 5 + 32);
}

function formatTemperature(temp) {
    return formatAndTranslateTemperature(temp, false);
}
```

therealbstern:
@markwal So it looks like to me like the only place from which formatTemperature is called also doesn't have access to the features settings, so I can't do that either, unless I can get the features options exposed to src/octoprint/static/js/app/viewmodels/temperature.js

foosel:
> unless I can get the features options exposed to src/octoprint/static/js/app/viewmodels/temperature.js

You do already have full settings access via [self.settingsViewModel](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/static/js/app/viewmodels/temperature.js#L6), which contains all settings including the features.

So something like e.g. this changed [here](https://github.com/foosel/OctoPrint/blob/devel/src/octoprint/static/js/app/viewmodels/temperature.js#L251):

``` javascript
var useFahrenheit = self.settingsViewModel.settings.feature.useFahrenheit();
var actualTemp = actuals && actuals.length ? formatTemperature(actuals[actuals.length - 1][1], useFahrenheit) : "-";
var targetTemp = targets && targets.length ? formatTemperature(targets[targets.length - 1][1], useFahrenheit) : "-";
```

should already work (assuming that you add a feature flag `useFahrenheit` and have the server return that on the API - also note that this depends on the `devel` branch, on which you should base the PR anyhow though).

therealbstern:
Fantastic, that was the piece I was missing.

I'll get an updated PR to you shortly.

therealbstern:
So, I don't know why it says there are conflicts, but I think this is good to go.

gddeen:
[on soapbox]
Not to be mean, but as an American I learned 0 centigrade is cold. 100 is boiling.
Boiling is 212. Half of that is a high fever. 50. Anything over 100 is very hot.
--
People should ALWAYS assume the hotted is HOT. I move my find in close to see
if there is radiant heat before touching something that can be 250 C.
--
ALL scientific equipment is metric.  It IS that way so we don't put up another Hubble Space Telescope
with mismatches. In High School and above only metric is used in measurements.
--
Parameters for the terminal are given in METRIC only. (for my printer, I assume yours too)
People should always think "This dang thing is metric..." so there is NEVER any confusion
about the temperature. Maybe a sign "Caution: Metric Scientific Equipment!"
--
Parameters for PLA/ABS, etc are all in metric. It does not say 1/8 inch. it does not say 450 degrees - 500 degrees. All settings, labels, inputs, (AND outputs) should stay unambiguously metric.

Let us Americans catch up to the world.
[off soapbox]

therealbstern:
If you don't want to turn on the conversion, it defaults to off.

For future reference, if you fell the need to put the disclaimer "not to be mean," you're being mean.

foosel:
There was a small bug with actually exchanging the settings between frontend and backend, but I fixed it. Now pushed to `devel`. Sorry for not getting around to this earlier, the [funding situation](http://octoprint.org/blog/2016/04/13/i-need-your-support/) has me in a rather iron grip.

therealbstern:
I apologize for the data exchange bug, but thank you for adopting this code.  I understand about the funding thing - my day job contract is due to expire in a few weeks.

foosel:
No biggie with the bug, was solved quickly. And good luck to you!

PR #1283, opened 2016-03-17
Added pause support for rf1000 and rf2000 printer M307[01] GCodes

paulsponagl:
not sure if this meets the guidelines, but i'll give it a try.
Added Support for the rf1000 and rf2000 Printers (Conrad Elektronik SE)
Additionally the following changes in config.yaml are necessary:

```
printerParameters:
  pauseTriggers:
  - regex: continuePrint
    type: disable
  - regex: pausePrint
    type: enable
```
of course i'd love to configure this within the printers itself...
btw. thanks Gina for octoprint and having the SD Upload back ! it was (even for me not being a python guy) very easy to read and understand your code / and that is - of course all i.m.O. - a sign for high quality code!

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -150,10 +150,10 @@ gcodeToEvent = {
 	"M0": Events.WAITING,
 	"M1": Events.WAITING,
 
-    # rfXXX printer pause
-    # see: http://www.rf1000.de/wiki/index.php/GCodes#M3070_-_Druck_pausieren_.28als_ob_Pause-Taste_gedr.C3.BCckt_wurde.29
-    "M3070": Events.WAITING,
-    "M3071": Events.WAITING,
+	# rfXXX printer pause
+	# see: http://www.rf1000.de/wiki/index.php/GCodes#M3070_-_Druck_pausieren_.28als_ob_Pause-Taste_gedr.C3.BCckt_wurde.29
+	"M3070": Events.WAITING,
+	"M3071": Events.WAITING,
 
 	# dwell command
 	"G4": Events.DWELL,
@@ -1970,8 +1970,8 @@ class MachineCom(object):
 		if self.isPrinting() and not self.isSdPrinting():
 			self.setPause(True)
 
-    _gcode_M3070_queuing = _gcode_M25_queuing # additions for rfxxx printer
-    _gcode_M3071_queuing = _gcode_M25_queuing
+	_gcode_M3070_queuing = _gcode_M25_queuing # additions for rfxxx printer
+	_gcode_M3071_queuing = _gcode_M25_queuing
 
 	def _gcode_M28_sent(self, cmd, cmd_type=None):
 		if not self.isStreaming():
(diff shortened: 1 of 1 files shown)


foosel:
First of all, thanks for the nice feedback!

Now, regarding those changes, I'd actually rather suggest to turn this into a plugin, since this is a prime example of something that could be easily accomplished through one. Thing is, it is a rather printer specific change, and I try to avoid it to let those creep into the main code base - have that going on for a while and maintainability goes to zero, since every printer vendor adds their own custom commands to the firmware...

So due to that I cannot really merge this without setting a bad precedent, sorry :/

But as I said, it's something that can be easily accomplished via a plugin, so you could turn it into one, [register it in the plugin repository](http://plugins.octoprint.org/help/registering/) so other owners of a Renkforce printer can find it and become famous! ;) Plus, if you'd want to make more Renkforce specific functionality available (e.g. custom pause triggers injected into the settings), you also then could just update your plugin, without having to depend on anyone else!

I'm not sure if you've already looked at the [plugin system in OctoPrint](http://docs.octoprint.org/en/master/plugins/gettingstarted.html), but if I don't misunderstand your code, you could easily achieve the same effect (and more!) by implementing a plugin that has a handler for the [octoprint.comm.protocol.gcode.queuing hook](http://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-comm-protocol-gcode-phase) and just invokes the same processing for `M3070` and `M3071` commands as for `M25`.

Something like:

``` python
import octoprint.plugin
import octoprint.event

class RenkforceSupportPlugin(octoprint.plugin.OctoPrintPlugin):

    def parse_renkforce_pause_gcodes(self, comm, phase, cmd, cmd_type, gcode, *args, **kwargs):
        if gcode not in ("M3070", "M3071"):
            # we are only interested in the above GCODEs
            return

        # handle like an M25: pause if printing but not streaming & fire an event
        comm._gcode_M25_queuing(cmd, cmd_type=cmd_type)
        self._event_bus.fire(octoprint.event.Events.WAITING

        # we do not return anything since we don't want to rewrite or suppress the
        # GCODE in any way, we just had to do some special processing for it

__plugin_name__ = "Rewrite Renkforce Pause GCODEs"

def __plugin_load__():
    # initialize our plugin implementation
    global __plugin_implementation__
    __plugin_implementation__ = RenkforceSupportPlugin()

    # register our handler method for the hook
    global __plugin_hooks__
    __plugin_hooks__ = {
        "octoprint.comm.protocol.gcode.queuing": __plugin_implementation__.parse_renkforce_pause_gcodes
    }
```

would probably already do it.

paulsponagl:
and again, great and detailed support - thanks. yes, i absolutely understand your concerns regarding a clean code base and i am happy that your response is better than i expected. i did take a look at the plugin structure, but realized that my limited time and python knowledge made "issue chatting and hint hunting" with you a better and more pleasant choice to get things done - thanks again! - happy to checkin your code as a plugin! (but gimme a bit of time - have to work on my project first)

Issue #1149, opened 2015-12-17
[Request] Layer up/down buttons in the G-Code Viewer


julianrinaldi:
It can occasionally be hard to select the exact layer you want in the G-Code viewer.  The scroll bars generally work well, but having the ability to click a button to move up and down a layer would be useful and more convenient in certain situations.  Having an up single layer button at top of the scroll bar, and a down single layer at the bottom of the scroll bar would be perfect.

woferry:
If possible it would be great to support the up and down keys as well, I'd rather navigate the layers with my keyboard than clicking up/down buttons in the browser window.

agarwali:
@foosel my team and I are working on OctoPrint and would like to tackle this issue. The bootstrap slider at https://github.com/foosel/OctoPrint/blob/536bb31965db17b969e7c1c53e241ddac4ae1814/src/octoprint/static/js/lib/bootstrap/bootstrap-slider.js 
does not support this arrow functionality. We could either try to edit the bootstrap-slider.js or we could try to add an extra button functionality on top of the Gcode viewer jinja template. What do you think would be a better approach?

BillyBlaze:
@agarwali, editing library files is a really bad practice, libraries tend to get upgraded and your work will then get dissolved. Also the bootstrap slider exposes itself on ```$element.data("slider")``` and on ```$element.slider("property")```

The best and easiest way is to edit the Gcode viewer jinja template and to add a click binding function in the corresponding viewmodel with something like:

```js
self.nextLayer = function() {
    self.layerSlider.slider("setValue", self.layerSlider.slider("getValue")+1);
});
```

Salandora:
This one was bothering me quite a while now.
I tackled this one for getting back into Javascript, it's inside the devel branch.
Move your Mouse over the GcodeViewer and use up arrow, down arrow, page up and page down buttons for a layer change.

Up/Down arrow: +1/-1 Layer
Page up/down: +10/-10 Layers

agarwali:
@Salandora it's ok. So you tackled the key binding issue, but the issue also mentions the implementation of up/down buttons which I don't see in the devel branch. Have you implemented the buttons?

Salandora:
I haven't added buttons for it, just the keyboard control, I personally like a keyboard solution more than extra buttons.

agarwali:
@Salandora my group and I tested your changes and it is working. If the button is not required than would this issue be called as resolved or marked as a milestone for a later version?

[...]

PR #1306, opened 2016-04-08
Dev/gcode layer button

agarwali:
highly appreciated!

checklist:

    Brainstorming ticket
  * [x] Your PR targets OctoPrint's devel branch (not master,
    maintenance or anything else)
    e.g. dev/my_new_feature
  * [x] Your changes follow the coding style
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

This pull request adds Layer Up/Down buttons to the Gcode viewer. It also fixes a bug by disabling key-bindings for layer up/down, which was requested in the same issue [#1149](https://github.com/foosel/OctoPrint/issues/1149)
It was tested by connecting to 3-D printer and trying out all the possible cases like: 1) when g-code is not uploaded 2) when g-code is uploaded 3) when g-code is loaded and deleted
The changes were automatically applied to the TouchUI plugin.
[#1149](https://github.com/foosel/OctoPrint/issues/1149)
screenshot from 2016-04-04 00 18 38

#### Further notes

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/gcode.js
+++ b/src/octoprint/static/js/app/viewmodels/gcode.js
@@ -52,6 +52,8 @@ $(function() {
 
         self.reader_sortLayers = ko.observable(true);
         self.reader_hideEmptyLayers = ko.observable(true);
+        
+        self.layerSelectionEnabled = ko.observable(false)
 
         self.synchronizeOptions = function(additionalRendererOptions, additionalReaderOptions) {
             var renderer = {
@@ -419,9 +421,7 @@ $(function() {
                     self.layerSlider.slider("disable");
                     self.layerSlider.slider("setMax", 1);
                     self.layerSlider.slider("setValue", 0);
-                    $('#btn_layer_up').prop('disabled', true);
-                    $('#btn_layer_down').prop('disabled', true);
-  
+                    self.layerSelectionEnabled(false);
                 }
                 self.currentLayer = 0;
             } else {
@@ -437,8 +437,7 @@ $(function() {
                     self.layerSlider.slider("enable");
                     self.layerSlider.slider("setMax", model.layersPrinted - 1);
                     self.layerSlider.slider("setValue", 0);
-                    $('#btn_layer_up').prop('disabled', false);
-                    $('#btn_layer_down').prop('disabled', false);
+                    self.layerSelectionEnabled(true);
                 }
             }
         };
@@ -529,7 +528,7 @@ $(function() {
                     value = value - 1; // No need to check against min, this is done by the Slider anyway
                     break;
             }
-            self.incrementLayer(value);
+            self.shiftLayer(value);
         };
 
         self.changeCommandRange = function(event) {
@@ -554,7 +553,7 @@ $(function() {
             self.tabActive = current == "#gcode";
         };
         
-        self.incrementLayer = function(value){
+        self.shiftLayer = function(value){
             if (value != self.currentLayer) {
                 event.preventDefault();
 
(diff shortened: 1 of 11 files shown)


src/octoprint/static/js/app/viewmodels/gcode.js
```diff
@@ -419,6 +419,9 @@ $(function() {
                     self.layerSlider.slider("disable");
                     self.layerSlider.slider("setMax", 1);
                     self.layerSlider.slider("setValue", 0);
+                    $('#btn_layer_up').prop('disabled', true);
```
foosel: Hm... While there are still some places within OctoPrint where there are such direct jquery references, in general it's less repetitive and cleaner to fetch those elements once and store them on the view model (e.g. like with `self.layerSlider`).

In that specific case here you manipulate the `disabled` property only (plus set a click handler further down below, see the comment to that), and I wonder if that wouldn't work better through a separate property on the viewmodel (e.g. `self.layerSelectionEnabled = ko.observable(false)`) and then setting both buttons' [enabled binding](http://knockoutjs.com/documentation/enable-binding.html) to that within the HTML markup (e.g. `data-bind="enabled: layerSelectionEnabled"`).

Would have the advantage to better separate view from view model (and in fact, the only reason the `self.layerSlider` element gets manipulated directly is because I didn't jump through the hoops of implementing a custom knockout binding for it, would also have been the cleaner approach actually and might be something to do in the future).

src/octoprint/static/js/app/viewmodels/gcode.js
```diff
@@ -566,6 +553,38 @@ $(function() {
         self.onTabChange = function(current, previous) {
             self.tabActive = current == "#gcode";
         };
+        
+        self.incrementLayer = function(value){
```
foosel: That's a bit of a misleading name - it doesn't actually increment but rather changes the layer. What about `changeLayer` instead?

src/octoprint/static/js/app/viewmodels/gcode.js
```diff
@@ -566,6 +553,38 @@ $(function() {
...
+                        type: 'slide',
+                        value: value
+                    }).trigger({
+                        type: 'slideStop',
+                        value: value
+                    });
+            }
+        };
+        
+        $( "#btn_layer_up" ).click(function() {
```
foosel: The `click` handler should rather be bound through knockout, so that this stays consistent with other parts of the application.

I suggest to define to methods `incrementLayer` and `decrementLayer` on `self` that do the job, then binding those to the buttons through `data-bind="click: incrementLayer"` and `data-bind="click: decrementLayer"`.

src/octoprint/templates/tabs/gcodeviewer.jinja2
```diff
@@ -3,6 +3,16 @@
         <input id="gcode_slider_layers" type="text">
         <canvas id="gcode_canvas" width="568" height="568"></canvas>
         <input id="gcode_slider_commands" type="text" style="width: 554px">
+        <div id ="button_control" class="btn-group-toolbar" style="width: 100%; padding-bottom: 7px; padding-top: 5px;text-align: center;">
+            <button id = "btn_layer_up" type="button" class="btn btn-primary btn-medium" style="width:48%" disabled='disabled'>
+                <i class="icon-white icon-arrow-up"></i>
+                <span>Layer Up</span>
```
foosel: The `Layer Up` text should be translatable. All you need to do for that to work is wrap the text in `{{ _('...') }}`, so `{{ _('Layer up') }}` and `{{ _('Layer down') }}` further below.

foosel:
Great work! Sorry for just getting around to reviewing this...

I added a bunch of comments, could you take a look at them? It's only small stuff, don't worry.

agarwali:
@foosel we fixed all the things you mentioned above. We replaced jquery, with ko click and enable functions. Also, when we renamed self.incrementLayer with self.changeLayer, it was throwing an error because a function named self.changeLayer was already defined. So, we renamed it to self.shiftLayer instead. The jinja template is also changed to enable the ko. functions. The text on the layer button is now replaced with the specific jinja formating to allow it to be transalted.

foosel:
@agarwali sounds good, I'll take a look as soon as possible. I currently have my hands somewhat full with handling the [funding situation](https://plus.google.com/u/0/+OctoprintOrg/posts/5m53jf6cdme), but let me already say: very nice work!

foosel:
Merged :)

PR #1308, opened 2016-04-10
Added setting to change GCode-visualizer file size threshold to ui #1145

philphilphil:
It adds the ability to change the file size threshold of the gcode visualizer in the ui settings.
It also moved the gcode settings to a new tab.
I tested on my machine with multiple files and different settings. It can be tested with the virtual printer, adding some Gcode-Files with different file sizes and load them.
#1145 
image
#### Further notes

### DIFF
--- a/src/octoprint/server/api/settings.py
+++ b/src/octoprint/server/api/settings.py
@@ -60,6 +60,8 @@ def getSettings():
 		},
 		"feature": {
 			"gcodeViewer": s.getBoolean(["gcodeViewer", "enabled"]),
+			"sizeThreshold": s.getInt(["gcodeViewer", "sizeThreshold"]),
+			"mobileSizeThreshold": s.getInt(["gcodeViewer", "mobileSizeThreshold"]),
 			"temperatureGraph": s.getBoolean(["feature", "temperatureGraph"]),
 			"waitForStart": s.getBoolean(["feature", "waitForStartOnConnect"]),
 			"alwaysSendChecksum": s.getBoolean(["feature", "alwaysSendChecksum"]),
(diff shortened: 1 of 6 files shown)


gddeen:
I use Opera, FireFox, Safari, Chrome, and a linux browser. 
I think Safari takes sizes much larger than others.
Maybe I need to test this, but I believe having a single desktop value wouldn't work in that case.

I assume the response might be "set it to 0 and you are no worse off?"

philphilphil:
The way it's implemented right now uses the exact number given, 0 would mean 0 bytes and no file would load automatically. If wanted i can implement, that 0 means unlimited.
Setting up a value for each browser in use is, afaik, not possible because those values are stored on your octoprint server.

foosel:
Could it be that you forgot to add a file to the commit? Only looking at the diff right now, but it seems like `sizeObservable` isn't declared anywhere, only used for the two size observables.

Also, I have to admit, I'm a bit unsure about the positioning of the value configuration. Looks a bit "misplaced" right within the feature checkbox army there, but I'm also a bit unsure what might be a better place... I wonder if it might make sense to extract the gcode visualizer settings (enable flag + those two settings) into their own tab under "Features" maybe?

foosel:
m( never mind, the `sizeObservable` is already there in `helpers.js`, just forgot about it :)

philphilphil:
I think you are right. Making a new Tab for the GCode-Visiualizer makes more sense because it's a feature.. I'll add it.
What is the procedure here? Do i close this Pull-Request and open a new one with my changes?

philphilphil:
ah, never mind, i didn't knew github updates pull-requests when adding a new commit. I'll add my code tonight.

philphilphil:
JFI: I added the code but my git is fucking something up... i'll rebase everything into one commit

foosel:
Now merged, sorry that it took me so long, that stupid [funding situation](http://octoprint.org/blog/2016/04/13/i-need-your-support/) is keeping my hands full these days.

Issue #1026, opened 2015-08-17
[Request] Automatically uppercase Gcode parameters where it makes sense


eovnu87435ds:
I'm not sure if this is an issue or not, but I found that if you type a gcode command with no arguments into the terminal page, it works whether the letter is upper or lower-case(g28 and G28 both work). Commands with multiple arguments only work if each letter is uppercase. (M280 P0 S150 works, m280 p0 s150 does not).

I noticed that if I use a different host program, such as Repetier host, upper and lower case both work.

**1. What were you doing?**   Sending mixed-case gcode through the terminal page

**2. What did you expect to happen?** The printer would execute the gcode

**3. What happened instead?** The printer did not execute gcode which had more than 0 arguments and was mixed case

**4. Branch & Commit or Version of OctoPrint:** 1.2.4 (master branch) via OctoPi

**5. Printer model & used firmware incl. version
   (if applicable - always include if unsure):** Folger Technologies RepRap Kossel running Marlin firmware

**6. Browser and Version of Browser, Operating 
   System running Browser (if applicable - always 
   include if unsure):** Windows 8.1x64 using Chrome 44.0.2403.155 m

**7. Link to octoprint.log on gist.github.com or pastebin.com
   (ALWAYS INCLUDE AND DO NOT TRUNCATE):** see http://pastebin.com/8xhwukWV

**8. Link to contents of terminal tab or serial.log on 
   gist.github.com or pastebin.com (if applicable - always 
   include if unsure or reporting communication issues AND
   DO NOT TRUNCATE):** See terminal tab, specifically lines 36-58 http://pastebin.com/KKS8xkd7

**9. Link to contents of Javascript console in the browser 
   on gist.github.com or pastebin.com or alternatively a
   screenshot (if applicable - always include if unsure
   or reporting UI issues):**

**10. Screenshot(s) showing the problem (if applicable - always
    include if unsure or reporting UI issues):**

I have read the FAQ.

foosel:
First of all: **Thank you for the effort put into this ticket**. Not many people read the ticket guidelines, follow them and then even make sure that the ticket is formatted well. That makes looking into this so much easier.

Now, regarding your problem, the current behaviour is intentional due to reasons I'm going to outline in a second, but I'm open to discuss a better solution.

[...]

PR #1321, opened 2016-04-30
Auto uppercase Gcode except commands in configurable blacklist

agarwali:
highly appreciated!

checklist:
- [x] If your changes are large or otherwise disruptive: You have
  Brainstorming ticket
- [x] Your PR targets OctoPrint's devel branch (not master,
  maintenance or anything else)
  e.g. dev/my_new_feature
- [x] Your changes follow the coding style
- [x] If your changes include style sheets: You have modified the
  lessc)
  have added unit tests
  nothing broke
- [x] You have added yourself to the AUTHORS.md file :)

The more details the better!

---

This PR enables auto upper casing of the G-Code commands. A new config setting was added: auto uppercase blacklist that is configurable in the advanced option of the serial tab, with M117 included in the default settings of the blacklist. The sendCommand method in the TerminalViewModel is changed to auto uppercase the entire G-code command except for commands in blacklist.
Updated documentation to include the default auto uppercase blacklist command M117.

This was tested on a 3D printer with all possible cases of G-code commands that were checked in the G-code terminal i) Commands in the blacklist, and ii) commands not in the blacklist.

Resolves: #1026 
#### Further notes

Resolves: #1026


wolfmanjm:
please note this will break the ability to send console commands to Smoothie which uses lowercase for the commands and uppercase for G/M codes. We increasingly have more useful console commands in smoothie and one of the nice thigns about octoprint is it sends them without needing any prefixes (like Pronterface requires the @ prefix).

foosel:
@wolfmanjm unless I misunderstand you, it shouldn't be a problem since it will only uppercase detected gcode (g, m or  the followed by a number) that's not blacklisted. So if you send "g0 x10" it will turn it into "G0 X10", but it will not touch "hello world" or any other arbitrary string that doesn't match `^[GMTgmt]\d+.*`

src/octoprint/static/js/app/viewmodels/terminal.js
```diff
@@ -259,10 +259,14 @@ $(function() {
...
             var commandMatch = command.match(re);
+            self.blacklist =  splitTextToArray(self.settings.serial_autoUppercaseBlacklist(), ",", true);
             if (commandMatch != null) {
                 command = commandMatch[1].toUpperCase() + ((commandMatch[2] !== undefined) ? commandMatch[2] : "");
+                if (self.blacklist.indexOf(commandMatch[1].toUpperCase()) < 0){
+                    command = command.toUpperCase()
+                }
             }
-
+
```
BillyBlaze: Can you fix this leading whitespace? A good editior should remove it by default if you saved (e.g. Atom)

src/octoprint/util/comm.py
```diff
@@ -566,7 +567,7 @@ def sendCommand(self, cmd, cmd_type=None, processed=False, force=False):
 			cmd = process_gcode_line(cmd)
 			if not cmd:
 				return
-
+
```
BillyBlaze: Ditto

agarwali:
@wolfmanjm, @foosel is right, it will not auto uppercase any arbitrary string unless the first part of the string matches /^([gmt][0-9]+)(\s.*)?/. If any smoothie command happens to start with [gmt][0-9], that we do not want to auto uppercase, it needs to be added to the blacklist.

wolfmanjm:
Oh excellent thanks. I read it to mean it upper cased everything unless it was blacklisted :)
So things like get pos will work and stay lower case :)

src/octoprint/static/js/app/viewmodels/terminal.js
agarwali: @BillyBlaze, it is now taken care of

src/octoprint/static/js/app/viewmodels/terminal.js
```diff
@@ -259,8 +259,12 @@ $(function() {
 
             var re = /^([gmt][0-9]+)(\s.*)?/;
             var commandMatch = command.match(re);
+            self.blacklist =  splitTextToArray(self.settings.serial_autoUppercaseBlacklist(), ",", true);
```
foosel: Setting an (undeclared) instance variable like this within a method is unclean. Especially since you don't use this instance variable outside of this method.

Two ways to go about this. One is simply declaring a local `var blacklist` instead of `self.blacklist`. That has the disadvantage however that on ever single command that is send the whole blacklist will be fetched from the settings view model and turned into an array before matching.

Cleaner would be to use `self.settings.serial_autoUppercaseBlacklist.subscribe(...)` to be notified of changes to the value (see [Kockout's docs for details on subscribe](http://knockoutjs.com/documentation/observables.html#explicitly-subscribing-to-observables)) and then updating a (previously declared) `self.blacklist` with the array-ified value that in turn is then used in the `sendCommand` routine.

So basically: 

  * add `self.blacklist` to the view model properly (top area), initialize to empty array
  * subscribe to `self.settings.serial_autoUppercaseBlacklist`:

    ``` javascript
    self.settings.serial_autoUppercaseBlacklist(function(newValue) {
        self.blacklist = splitTextToArray(newValue, ",", true);
    });
    ```
  * Use `self.blacklist` in `sendCommand`

src/octoprint/util/comm.py
```diff
@@ -270,6 +270,7 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
 
 		self._long_running_commands = settings().get(["serial", "longRunningCommands"])
 		self._checksum_requiring_commands = settings().get(["serial", "checksumRequiringCommands"])
+		self._auto_uppercase_blacklist = settings().get(["serial", "autoUppercaseBlacklist"])
```
foosel: What is this needed for?

foosel:
Hey, took a look at the PR and found a couple of points, could you take a look at the comments? Thanks :)

foosel:
Ok, this has now been rebased against `devel` and then cherry picked as 51406b7 and ce16f86 to `maintenance`, as part of #2177. Thanks to @pbackx for picking up the reigns and taking care of the review comments 👍 

Closing manually since Github doesn't pick up on cherry-picks.

Issue #1324, opened 2016-05-04
Daemonized OctoPrint does not clean up pidfile anymore


kevans91:
#### What were you doing?

Writing an rc script for FreeBSD. Reproduction steps:

```
octoprint --daemon start --pid /tmp/octoprint.pid
kill -TERM `cat /tmp/octoprint.pid`
ls -l /tmp/octoprint.pid
```

/tmp/octoprint.pid persists -- this was previously addressed in issue #711 with PR #712, but it has since broken. The following SIGTERM registration clobbers this handler: https://github.com/foosel/OctoPrint/blob/master/src/octoprint/server/__init__.py#L511-L516 -- so the daemonize'd SIGTERM does not get called, and the pidfile does not get cleaned up.

#### What did you expect to happen?

pidfile to get cleaned up

#### What happened instead?

pidfile did not get cleaned up

#### Branch & Commit or Version of OctoPrint

Verified on:

master & maintenance: e79703b
devel: 1b1148d

#### Link to octoprint.log

https://gist.github.com/kevans91/b11ea167ec1fd9d4c4159400608abcd2

I have read the FAQ.

foosel:
Fixed in both `maintenance` and `devel`, will be in next release, can be closed, right?

kevans91:
Yup! =)

PR #1328, opened 2016-05-06
Make Server class octoprint_daemon-aware, make sure terminated() gets…

kevans91:

This PR is for cleaning up the pidfile of a daemonized instance upon termination. This is my proposed solution to address issue #1324 

```
octoprint --daemon start --pid /tmp/octoprint.pid
kill -TERM `cat /tmp/octoprint.pid`
ls -l /tmp/octoprint.pid
```

#1324 

#### Further notes
This seems like the necessary solution for this case. The Server class handles clean-up and termination, so it seems to make the most sense to make it octoprint_daemon-aware and force a clean-up of the Daemon instance. The exit() is no longer needed, as the termination should otherwise be handled as it normally is (now).

### DIFF
--- a/src/octoprint/server/__init__.py
+++ b/src/octoprint/server/__init__.py
@@ -113,13 +113,14 @@ def load_user(id):
 
 
 class Server(object):
-	def __init__(self, settings=None, plugin_manager=None, host="0.0.0.0", port=5000, debug=False, allow_root=False):
+	def __init__(self, settings=None, plugin_manager=None, host="0.0.0.0", port=5000, debug=False, allow_root=False, octoprint_daemon=None):
 		self._settings = settings
 		self._plugin_manager = plugin_manager
 		self._host = host
 		self._port = port
 		self._debug = debug
 		self._allow_root = allow_root
+		self._octoprint_daemon = octoprint_daemon
 		self._server = None
 
 		self._logger = None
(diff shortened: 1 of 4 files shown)


src/octoprint/server/__init__.py
```diff
@@ -521,6 +522,9 @@ def on_shutdown():
 		def sigterm_handler(*args, **kwargs):
 			# will stop tornado on SIGTERM, making the program exit cleanly
 			def shutdown_tornado():
+				if self._octoprint_daemon is not None:
```
foosel: This may be overthinking it but - don't we actually want that call to the `terminated` callback to happen after we've shutdown the ioloop instead of before? Should normally not make much of a difference I believe, but it still feels wrong to remove the pidfile while the process is still running. 

Or am I misremembering things here about the `ioloop.stop` call and its consequences and it will actually do a `sys.exit` meaning having the `terminated` call after would be too late?

foosel:
Hey, big thanks for that! I had a question about one part (more like "what's the best approach here" than anything), could you take a look at that?

kevans91:
Hi!

That's a valid concern =). I do know that some consumers of the pidfile will wait for the process to actually terminate before considering it terminated (pidfile's existence notwithstanding), but that's not necessarily a good assumption to make across the board. I'm closing this one, and I will re-open it with a cleaner solution here shortly.

PR #1330, opened 2016-05-09
Make Server class octoprint_daemon-aware, make sure terminated() gets…

kevans91:

This PR is for cleaning up the pidfile of a daemonized instance upon termination. This is my proposed solution to address issue #1324 

```
octoprint --daemon start --pid /tmp/octoprint.pid
kill -TERM `cat /tmp/octoprint.pid`
ls -l /tmp/octoprint.pid
```

#1324 

#### Further notes
This seems like the necessary solution for this case. The Server class handles clean-up and termination, so it seems to make the most sense to make it octoprint_daemon-aware and force a clean-up of the Daemon instance. The exit() is no longer needed, as the termination should otherwise be handled as it normally is (now). 

This was previously created as #1328, but this iteration addresses the concern by @foosel that we should be cleaning up the pidfile later on in termination. This iteration moves the octoprint daemon cleanup to on_shutdown() instead, just before printing 'Goodbye!', and mentions in the log that it's cleaning up the pidfile. This does mean that we're no longer cleaning up the pidfile strictly on SIGTERM, but that's probably okay because if _octoprint_daemon is not None then we really need to clean up after it if we can.

### DIFF
--- a/src/octoprint/cli/server.py
+++ b/src/octoprint/cli/server.py
@@ -11,7 +11,7 @@ import sys
 
 from octoprint.cli import pass_octoprint_ctx, bulk_options, standard_options
 
-def run_server(basedir, configfile, host, port, debug, allow_root, logging_config, verbosity):
+def run_server(basedir, configfile, host, port, debug, allow_root, logging_config, verbosity, octoprint_daemon = None):
 	"""Initializes the environment and starts up the server."""
 
 	from octoprint import init_platform, __display_version__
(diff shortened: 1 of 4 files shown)


kevans91:
I should also note here that this may be better for future cases, too- in the case that there's other options (than SIGTERM) for termination of a daemonized OctoPrint, this cleanup should always happen as long as we're exiting gracefully.

foosel:
Perfect, thank you :) I'm also going to backport this to the `maintenance` branch so it will be included (and #1324 fixed) in `1.2.12`.

kevans91:
Oooo, thanks! =)

Looks like it was a pretty straightforward backport -- just needs the two self._logConf and self parameters added to https://github.com/foosel/OctoPrint/blob/maintenance/src/octoprint/__init__.py#L27 =)

Thank you!

PR #1347, opened 2016-05-20
Corrects example URL in Slicing API Doc

dougcalobrisi:
This fixes a simple typo in the API doc for fetching slicing profiles. The previous example had /profiles appended to it, whereas it should just be /api/slicing to fetch a list of all slicers and profiles.


foosel:
Thanks for spotting this!

Issue #1320, opened 2016-04-29
Can't delete mp4 timelapses and wrong filename displayed in notification of completion


ntoff:
With the addition of the ability to view and download mp4's through a wrapper script it seems you can't delete them. Additionally the wrong file extension is used when displaying the rendering complete notification.

https://github.com/guysoft/OctoPi/issues/184 wrapper script

----

#### What were you doing?

Testing time lapse mp4 rendering

#### What did you expect to happen?

Render out to mp4 and be able to delete / download mp4

#### What happened instead?

mp4 rendered fine, downloads work, deleting doesn't remove any files. Additionally even though mp4 is rendered the mpeg extension is still used in the notification of rendering completion. I suggest perhaps simply removing the extension all together in the notification popup since detecting the type of file is probably impossible when using the wrapper script.

#### Branch & Commit or Version of OctoPrint

Version: 1.2.11.dev12+g119f727 (maintenance branch)

#### Printer model & used firmware incl. version

Mendel90 Sturdy running Marlin from ye olde times

#### Browser and Version of Browser, Operating System running Browser

Chrome Version 49.0.2623.112 m on windows 10 pro

#### Link to octoprint.log

https://gist.github.com/ntoff/09fb33daaf309c100116c20aaf96b32f

#### Link to contents of terminal tab or serial.log

nothing there aside from the usual temperature request / responses

#### Link to contents of Javascript console in the browser

Nothing obvious http://i.imgur.com/uoB65wf.png and it's pretty empty

#### Screenshot(s) showing the problem:

screenshot of incorrect file extension
http://i.imgur.com/A3X2Jy2.png

I have read the FAQ.

foosel:
Solved on both `devel` and `maintenance` branch, will be part of the next release.

PR #1358, opened 2016-05-31
Deletemp4

CapnBry:
The timelapse api has allowed mp4 timelapse files for a bit now, but it is impossible to delete them because the api only allows you to perform the DELETE request on files with extension ,mpg.

This pull request switches from the standard extension-based mapping to use the filemanager's `valid_file_type` method. A new type "video" has been added with the two mime types for mpeg and mp4.

Note that this is the last usage of util.is_allowed_file() in all of OctoPrint so that might be able to be removed if you're no longer going to use it (and prevent it from being used by contributors instead of the filemanager).

This should close the deletion portion of #1320, but the notification still does say .mpeg so that part is still open.

### DIFF
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -38,7 +38,8 @@ def full_extension_tree():
 		),
 		# extensions for videos (timelapse)
 		video=dict(
-			mpeg=ContentTypeMapping(["mpeg", "mpg", "mp4"], "video/mpeg")
+			mpeg=ContentTypeMapping(["mpeg", "mpg"], "video/mpeg"),
+			mp4=ContentTypeMapping(["mp4"], "video/mp4")
 		)
 	)
 
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -35,6 +35,10 @@ def full_extension_tree():
 		# extensions for printable machine code
 		machinecode=dict(
 			gcode=ContentTypeMapping(["gcode", "gco", "g"], "text/plain")
+		),
+		# extensions for videos (timelapse)
+		video=dict(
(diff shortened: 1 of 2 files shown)


foosel:
The problem with that is that it also allows uploading mpg files ;) The file manager is actually only for what you find in the "Files" section in the sidebar. So basically anything that is related to "printables" or "sliceables".

Timelapses need a different handling. For now the best solution of the issue would probably be to extend the original valid file check with the new possible endings. Do you want to do that or shall I?

About the name problem, that is not really solvable currently. The way the mp4 wrapper script works is that OctoPrint simply doesn't know if it's generating the mpg file it's specifying or if some replacement happens in between. So it can't really know that it should display something else but "mpg" there.

CapnBry:
Oh goodness me, see this is what I get for looking deeply enough to do damage.

* Ok here's the code that prevents it, let's patch that.
* No let's look and see where else this is used to make sure there are no caveats to using it
* Oh this is the only place it is used. I know there are checks to make sure file uploads are of the right type or if they need to be sliced let's see what that code does
* Makes change allowing mp4 files to be uploaded.
* Angers our many armed mascot of OctoPrint

So yes, please just make the 4 character change to allow mp4 files to be deleted from the timelapese list.

As far as the notification, I know. There's no way for you to know what file type is returned by the process, or even if the user script has generated a completely different name. It doesn't bother me, I rarely even see the notification because I am off picking daises or something. Perhaps the notification can strip the extension? On the plus side that would make the toast message easier to read because the filename is already like 43 characters long. On the down side, it doesn't fix the issue of knowing what the actual filename is.

Thanks for your time, Gina!

foosel:
Changed the allowed file extensions and also stripped the extension from the reported filename.

Also, the many armed mascot has asked me to tell you that all is well and it still loves you :P

PR #1375, opened 2016-06-17
Some options for handling RepRapFirmware SD card behavior

markwal:

It allows firmwares to neglect to respond with the file name for SD card file selection/open
Also, adds an option that causes a relative path to be used for SD file selections

The PR adds an option to the virtual printer to test the selection responses.  And the UI option sends a different path name.  I tested both using the virtual printer and then a RRF user on the forum tested against an actual printer.

#### Further notes

I did as a PR to make it easier to discuss the proposed changes

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -300,6 +300,7 @@ default_settings = {
 			"okWithLinenumber": False,
 			"numExtruders": 1,
 			"includeCurrentToolInTemps": True,
+			"includeFilenameOnOpened": True,
 			"movementSpeed": {
 				"x": 6000,
 				"y": 6000,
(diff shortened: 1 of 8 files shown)


src/octoprint/settings.py
```diff
@@ -300,6 +300,7 @@ def settings(init=False, basedir=None, configfile=None):
 			"okWithLinenumber": False,
 			"numExtruders": 1,
 			"includeCurrentToolInTemps": True,
+			"includeFilenameOnOpened": True,
```
markwal: This typo got fixed in the next commit.  Looks like I squashed it with the wrong one.

foosel:
Looks good on first look (and sorry for not getting around to do that earlier). Would it make sense to backport this to the maintenance branch?

markwal:
I think so.  There's at least a couple of people running on my test build that'd probably be happier on regular releases.

PR #1384, opened 2016-06-22
Replaced broken link

docdawning:
This is a super simple change - I found a broken url and so I found what I think is a suitable replacement.

Replaced dead url: https://pythonhosted.org/setuptools/setuptools.html#dynamic-discovery-of-services-and-plugins
with what I think is a suitable alternative: http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins


foosel:
Thanks! :+1:

PR #1386, opened 2016-06-23
fix intermediary.html for url with parameters

MoonshineSG:
handles the OctoPrint startup intermediary.html for UI handling plugins that need parameters

tested using the "Nautilus" plugin

review test:
`stop octoprint service`
`start octoprint service` and immediately go to `http://your_octoprint_url/?bar=yes&foo=no`
The intermediary.html should load the correct URL preserving the parameters (previously it would give the 'Looks like something went wrong during startup, the server is gone again.' error message)

The mobile ui provided by a plugin sets a parameter to load the `apikey`

none

none

#### Further notes
none

### DIFF
--- a/src/octoprint/static/intermediary.html
+++ b/src/octoprint/static/intermediary.html
@@ -141,14 +141,21 @@
                 var timeout = 1500;
 
                 var baseUrl = window.location.href;
+                var reloadUrl = window.location.href;
                 if (baseUrl.indexOf("#") > -1) {
                     baseUrl = baseUrl.substring(0, baseUrl.indexOf("#"));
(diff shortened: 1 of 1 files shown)


foosel:
[I modified the change slightly](https://github.com/foosel/OctoPrint/commit/674d625150f2761ed4dd62e5a8be8652c371e4e8), put you into `AUTHORS.md` and backported the commit to `maintenance`. So consider this merged even though since I [cherry picked the commit](https://github.com/foosel/OctoPrint/commit/bfecb5fadf0a841f20c8cc6487ce3c20aedcf103) Git/Github doesn't detect this properly. Thanks!

PR #1398, opened 2016-07-04
python 3 compatibility changes

ByReaL:
highly appreciated!

checklist:

    Brainstorming ticket
  * [ ] Your PR targets OctoPrint's devel branch (not master,
    maintenance or anything else)
    e.g. dev/my_new_feature
  * [ ] Your changes follow the coding style
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

python 3 compatibility changes
1)
-	if callable(after_settings):
+	if isinstance(after_settings, collections.Callable):

2) import Queue is now lower caps

3) exceptions
except KeyError as e:

4) assertEquals is deprecated but assertEqual does exactly same thing

made sure is still works in python 2.7 and number of errors reduced in python 3.5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ OctoPrint.egg-info
 *.orig
 
 *.codekit
+
+venv
+venv26
+venv27
+venv35
--- a/setup.py
+++ b/setup.py
@@ -14,8 +14,8 @@ import octoprint_setuptools
 
 # Requirements for our application
 INSTALL_REQUIRES = [
-	"flask>=0.9,<0.11",
-	"werkzeug>=0.8.3,<0.9",
+	"flask>=0.10,<0.11",
+	"werkzeug>=0.9,<=0.9",
 	"tornado>=4.0.2,<4.1",
 	"sockjs-tornado>=1.0.2,<1.1",
 	"PyYAML>=3.10,<3.11",
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -8,6 +8,7 @@ import logging
 #~~ version
 
 from ._version import get_versions
+import collections
 versions = get_versions()
 
 __version__ = versions['version']
@@ -58,7 +59,7 @@ def init_platform(basedir, configfile, use_logging_file=True, logging_file=None,
                   logging_config=None, debug=False, verbosity=0, uncaught_logger=None,
                   uncaught_handler=None, after_settings=None, after_logging=None):
 	settings = init_settings(basedir, configfile)
-	if callable(after_settings):
+	if isinstance(after_settings, collections.Callable):
 		after_settings(settings)
 
 	logger = init_logging(settings,
@@ -69,7 +70,7 @@ def init_platform(basedir, configfile, use_logging_file=True, logging_file=None,
 	                      verbosity=verbosity,
 	                      uncaught_logger=uncaught_logger,
 	                      uncaught_handler=uncaught_handler)
-	if callable(after_logging):
+	if isinstance(after_logging, collections.Callable):
 		after_logging(logger)
 
 	plugin_manager = init_pluginsystem(settings)
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -8,7 +8,10 @@ __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms
 import datetime
 import logging
 import subprocess
-import Queue
+try:
+	import queue
+except ImportError:
+	import Queue as queue
 import threading
 import collections
 
@@ -120,7 +123,7 @@ class EventManager(object):
 		self._registeredListeners = collections.defaultdict(list)
 		self._logger = logging.getLogger(__name__)
 
-		self._queue = Queue.Queue()
+		self._queue = queue.Queue()
 		self._worker = threading.Thread(target=self._work)
 		self._worker.daemon = True
 		self._worker.start()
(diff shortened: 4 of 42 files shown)


foosel:
Hm... This is starting to become worryingly invasive. Which - don't get me wrong - is completely to be expected considering that it's basically trying to achieve compatibility with a completely different python version, but the timing is bad. I'm currently in the process of finalising my work on the 1.3.0 release-to-be, which is basically what's happening in `devel` at the moment. I don't want to risk introducing new bugs now by merging so many invasive changes that might risk stability under 2.7. Not merging them however will probably lead to introducing further incompatibilities, making maintenance of this PR difficult to "keep up".

I think we have to approach this a bit more structured to allow for a merge after release of 1.3.0.

First of all it should be decided on which Python3 version this is supposed to achieve compatibility with - the `callable` built in for example was only absent from 3.0 until 3.2, current Python 3 versions [do include it](https://docs.python.org/3/library/functions.html#callable). Personally I don't think it makes sense to target a Python 3 version from 2009 - especially if doing so reduces code readability, `callable(foo)` is so much easier on the eyes than `isinstance(foo, collections.Callable)`, more so since the latter even needs an import to work correctly - and instead target current Python 3.5 and above.

Also I see problems brewing down the line when str vs unicode vs bytes comes into play, with lots of code complications. I'd actually rather tackled this task of achieving Python 3 compatibility with something like a 2/3 compatibility layer (`six` comes to mind, there might be other/better alternatives though) instead of increasing the code complexity throughout the code base by wrapping everything in queries like "if py3 do this else do that" or "try py3 variant and if you fall on your face do the py2 one". Stuff like that will make the code incredibly tricky to read (the try-catchs around the queue imports are harmless compared to what I'm fearing) and hence also to maintain, and that's undesirable.

Finally, please set up your git client with a proper mail address, "invalid-email-address" as an author is not very helpful when trying to figure out who to ask about a specific change in the future based on data from the git log :)

ByReaL:
Hi,

i have a lot of changes done for 2/3 compatibility all submitted in my branch or still working on them.
but i would like to submit (make pull requests) small targeted changes and not a huge change list in order to make it more manageable and be able to asses the risk and explain each change.

six and future packages are things that i'm already using in my branch

now regarding this specific changes, i selected for this pull request only the safest changes 


2) import Queue is now lower caps
import Queue as queue is already being used in one of the code files already and is working fine, adding it in all files should be ok

3) exceptions
except KeyError as e:
removing the comma KeyError, e and adding as instead of it again should not introduce any bug

4) assertEquals is deprecated but assertEqual does exactly same thing
https://docs.python.org/2.7/library/unittest.html
For historical reasons, some of the TestCase methods had one or more aliases that are now deprecated. The following table lists the correct names along with their deprecated aliases:

Method Name	Deprecated alias(es)
assertEqual()	failUnlessEqual, assertEquals



the bytes/unicode/str changes will be the last i will do because this have the greatest risk to introduce bugs and may require more advance changes. I may need to add even new tests because in python3 str includes unicode and "🆃🅴ⓢⓣ①②③𝟜⑤" will be a valid input
and to have same functionality in python 2.x i will also have to add 
from __future__ import absolute_import, division, print_function, unicode_literals (unicode_literals will be part of my last set of changes)

the email address issue i noticed it also and fixed it for further pull requests.

foosel:
Ok, what about `callable` and the targeted Python 3 version?

ByReaL:
callable is used for type check isinstance and this should also be a safe change 
python targeted version will be 3.3+

regarding current progress
 in my fork i added python 3.5 and configured a travis bot and working through the bugs, it compiles/builds but does not passes all tests yet and is not working either.
once 3.5 works fine without braking 2.7 i will test and do final changes if needed for python 3.3+ compatibility.

i also noticed you prefer one big commit with all the changes and i can do that once i have it fully working on python 3.5 but if you are ok i will like to keep submitting smaller and smaller changes

foosel:
> callable is used for type check isinstance and this should also be a safe change 
> python targeted version will be 3.3+

You don't need to replace `callable(foo)` with `isinstance(foo, collections.Callable)` though if you are targeting Python 3.3 and above, as mentioned earlier, [that *does have* the `callable` builtin just like 2.7](https://docs.python.org/3/library/functions.html#callable) and I don't see an advantage in replacing that with an `isinstance` check that even requires an additional import of the type to compare with and which makes the whole code less readable. If there is an advantage, please explain. If not, please roll that switch out back :)

In general the goal should be to make only the changes that are absolutely necessary to achieve functionality, nothing more, to keep the impact to a minimum. So if something is a "must" for Py 3 - yes. If it's a "should" - no.

> i also noticed you prefer one big commit with all the changes and i can do that once i have it fully working on python 3.5 but if you are ok i will like to keep submitting smaller and smaller changes

That's ok in principle if they are self contained, I just have this issue with - as mentioned - working on finalising 1.3.0 and I have to be very careful what I merge at that state, or otherwise I risk postponing the release even further. So the more lines your changes touch at this point, the more critical I'll have to look at them. Sorry, it really is a tricky point in time.

Long term the question is also how to ensure compatibility down the road. Even running the unit test suite under both versions via tox will probably not be enough, that doesn't have enough coverage as of now I fear to be a good safeguard. Even firing up the server will probably not be enough. So if you have any idea about reliable integration tests, feel free to share :)

ByReaL:
perfect i will resubmit the changes only for 2 3 and 4 without 1
and regarding the release on 1.3.0 i totally understand ( my changes they don't even have to be in there) they can be integrated past the 1.30 release.

PR #1408, opened 2016-07-12
wait 5 seconds for the plugins to be fully loaded

MoonshineSG:
Wait 5 seconds for all the plugins to start (especially those handling alternative UI) before reloading the page.

### DIFF
--- a/src/octoprint/static/intermediary.html
+++ b/src/octoprint/static/intermediary.html
@@ -135,6 +135,10 @@
                     }
                 }, timeout);
             }
+            
+            function sleep (time) {
+              return new Promise((resolve) => setTimeout(resolve, time));
+            }
 
             window.onload = function() {
                 var intervals = [1, 1, 2, 3, 5, 8];
@@ -172,13 +176,15 @@
                         serverIsOnline = true;
                         message.className = "pulsate1 green";
                         message.innerText = "OctoPrint server online, reloading page...";
-
-                        var reloadUrl = baseUrl + currentQuery + currentFragment;
-                        if (reloadUrl == window.location.href) {
-                            window.location.reload(false);
-                        } else {
-                            window.location.href = reloadUrl;
-                        }
+             
+                        sleep(5000).then(() => {
+                            var reloadUrl = baseUrl + currentQuery + currentFragment;
+                            if (reloadUrl == window.location.href) {
+                                window.location.reload(false);
+                            } else {
+                                window.location.href = reloadUrl;
+                            }
+                        });
                     } else {
                         // online.gif still not available, let's look at
                         var interval = 10;
--- a/src/octoprint/static/intermediary.html
+++ b/src/octoprint/static/intermediary.html
@@ -136,10 +136,6 @@
                 }, timeout);
             }
             
-            function sleep (time) {
-              return new Promise((resolve) => setTimeout(resolve, time));
-            }
-
             window.onload = function() {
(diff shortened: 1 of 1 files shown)


src/octoprint/static/intermediary.html
```diff
@@ -172,13 +176,15 @@
...
                         message.innerText = "OctoPrint server online, reloading page...";
-
-                        var reloadUrl = baseUrl + currentQuery + currentFragment;
-                        if (reloadUrl == window.location.href) {
-                            window.location.reload(false);
-                        } else {
-                            window.location.href = reloadUrl;
-                        }
+             
+                        sleep(5000).then(() => {
```
foosel: No ES6 syntax for now please :) Might introduce browser compatibility issues

MoonshineSG:
I'm not sure why the build failed...

foosel:
Not your fault, I broke something in the tests myself but haven't yet
gotten around to fixing it due to hunting down some other issue for the
past days.

foosel:
Could you describe the problem you encountered a bit more for which this is the solution? Plugins should usually not need to "start up", the only thing that might not have happened yet is that the preemptive caching not having taken place.

UiPlugins in general shouldn't need time to startup though, either they can deliver a page or they can't, and introducing a general delay of 5s (or more, your current implementation sets a timeout of 10000ms instead of 5000) into the reload to mask a different issue (which I suspect is in the way here) is something I'd like to avoid :)

MoonshineSG:
Yes, it's the preemptive caching. The plugin itself does not have any startup time needed. If the request comes before that (analysed from the log files) OctoPrint serves the default page without checking if there's a UIPlugin that matches the conditions for a different index page. 

I increased to 10sec (too much if you ask me) because I had few instances where 5sec still was not enough... 
I understand your reluctance to add 10sec to the startup time. Please let me know if there's an alternative. 

As you suggested, I am trying to change the passing of APIKEY from GET to POST. I wonder how would that work with the "intermediary.html" ?
For now I am having problems, with that as the root does not accept a POST method. But this is a different problem.

foosel:
>  If the request comes before that (analysed from the log files) OctoPrint serves the default page without checking if there's a UIPlugin that matches the conditions for a different index page.

Definitely a bug then that should rather be addressed than adding arbitrary delays. I actually looked into that stuff about a week ago since I spotted a similar behaviour and thought I'd had solved that. Note that the preemptive cache is completely server side though, so it will still cache the wrong things potentially, even if the intermediary.html does the redirect at a later point in time. So I'm not entirely sure what you are observing there specifically, it would be great if you could give some steps to reproduce reliably.

> As you suggested, I am trying to change the passing of APIKEY from GET to POST

Sorry, you misunderstood me there. I meant for the apikey to not be supplied in the query but in a custom header, [as described here](http://docs.octoprint.org/en/master/api/general.html#authorization). I'm also still unsure why you need it at all for querying the index page (you do not need an apikey to access `/`, or it would be impossible to ever access that page, the api key is only needed for api access, not for accessing HTML pages), but I'll try to get back to you on that on hangouts asap.

MoonshineSG:
Ok... I updated to version `dev1223_g7908f52` (i was on `dev1209_g9b6b4a9`) and not it seams that the problem is gone (without the delay from this pr). Seams like you crushed that bug.  Good news!

foosel:
So that change isn't necessary for you anymore?

PR #1412, opened 2016-07-15
python3: from __future__ import absolute_import, division, print_function

ByReaL:
it makes the behavior between python 2 and 3 similar for the below items
from __future__ import absolute_import, division, print_function
it also ensures that changes submitted after this will benefit from this behavior changes
travis-ci
visually inspected the divisions in the code 
division is good  it ensures we get same results in both python 2 and 3 but is hard to determine if any existing division 3/2 = 1 will now become =1.5 and brake something
the change is necessary but has a risk.

for example 
line_width = wall_thickness / line_count
will work fine because wall_thickness is a float and division will be fine

or here where the code explicit calls for float division by using 2.0
self.get_float("machine_width") / 2.0 

and all i could see looked ok, but cannot be 100% sure one division in all the code has not modified it's value.

if there is such case found i could add unit test and fix the problem in that specific instance.

### DIFF
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python2
 # coding=utf-8
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, division, print_function
 
 import sys
 import logging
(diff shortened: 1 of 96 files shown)


foosel:
There were three locations which required integer division, see e961234458b3fb54ff948e7e521117d080b0b720. Rest should be fine though :+1:

PR #1414, opened 2016-07-15
python3: os.umask(002) to os.umask(0o002)

ByReaL:
#### Further notes

os.umask(0o002) is the python 2/3 way


foosel:
Spotted a left over in the corresponding unit test, see commit above.

PR #1415, opened 2016-07-15
python3: iterators, bytes and next

ByReaL:
python 3 compatibility

iteritems got removed items is to be used for python 2 and 3
compatibility , it returns a view in python 3 and must be casted to a
list if the list is needed.

bytes form builtins

using builtin function next is the compatible way between python 2 and 3
travis ci

### DIFF
--- a/src/octoprint/server/__init__.py
+++ b/src/octoprint/server/__init__.py
@@ -17,6 +17,7 @@ from babel import Locale
 from watchdog.observers import Observer
 from watchdog.observers.polling import PollingObserver
 from collections import defaultdict
+from builtins import bytes
 
 import os
 import logging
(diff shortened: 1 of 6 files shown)


foosel:
It will also need a version of the Werkzeug dependency >= 0.9 (plus possible further dependency updates), since 0.8.3 is not compatible yet to Python 3. Just FYI, you probably already noticed that though.

I'm a bit worried about that since I remember doing some hack-ish stuff with regards to the WSGI environment in order to allow uploading large files that do not fit in memory, by streaming uploaded files to disk and modifying the WSGI environment of the request to indicate to the underlying Flask app where to find the uploaded file on disk. That might run into problems, or might not, would need to be tested.

PR #1416, opened 2016-07-15
python3: [DO NOT MERGE] just for feedback

ByReaL:
this is all work in progress but i would like everyone feedback on few aspects with will determine how i approach the actual change for further pull requests. on python3 compatibility.

1) unicode, i will like to add 
from **future** import ....... unicode_literals 
in all the source code files, and mark with b'ABCxyz' wherever needed and remove u'' for eveywhere else to have everything compatible and conforming to the python3 while having same behavior also on python2. 
or
u'' is recognized by python3  (redundant since all string in py3 are unicode anyway) so we could add u'' on all strings that are not b'' to make things very clear
so is b'xyz' & 'abc' versus b'xyz' & u'abc'  with unicode_literals  imported in both cases. 
choices have same behavior on both python2 and 3 so it's just a matter of preference.

 2) for file opening instead of using open i will like to use io.open that is equivalent with python3 open
and specify the file type as text or binary and the encoding to make things run same on both py2 and 3 and make things very clear

3) update packages versions to include packages that are python 2 and 3 compatible
all changes look more or less ok but i would like some feedback on below change and the reson is that netifaces is a bit abandoned and somebody else made fixes for python3 and renamed the package
    - "netifaces>=0.10,<0.11",
    + "gns3-netifaces>=0.10,<0.11",

4) migrate doctest that looks for u'' strings to unitest so they can pass in both versions or use the doctest-ignore-unicode package to make doctest ignore the u'' and have the tests passing in both python2 and 3

5) not exactly related to python3 migration but noticed it while doing unicode/bytes work is the dos8.3 file names function witch is not generating 8.3 names (as dos/windows) and planing to rewrite the function to behave exactly like what you will find under dos or windows in term of SFN DOS8.3 file names

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,11 @@
 language: python
 python:
 - "2.7"
+- "3.5"
 install:
 - pip install -e .[develop]
 script:
-- nosetests --with-doctest
+- nosetests --with-doctest --with-coverage
 sudo: false
 git:
   depth: 250
--- a/setup.py
+++ b/setup.py
@@ -14,13 +14,13 @@ import octoprint_setuptools
 
 # Requirements for our application
 INSTALL_REQUIRES = [
-	"flask>=0.9,<0.11",
-	"werkzeug>=0.8.3,<0.9",
+	"flask>=0.10,<0.12",
+	"werkzeug>=0.9,<=0.9",
 	"tornado>=4.0.2,<4.1",
 	"sockjs-tornado>=1.0.2,<1.1",
 	"PyYAML>=3.10,<3.11",
 	"Flask-Login>=0.2.2,<0.3",
-	"Flask-Principal>=0.3.5,<0.4",
+	"Flask-Principal>=0.4,<0.5",
 	"Flask-Babel>=0.9,<0.10",
 	"Flask-Assets>=0.10,<0.11",
 	"Flask-Markdown>=0.3,<0.4",
@@ -28,7 +28,7 @@ INSTALL_REQUIRES = [
 	"netaddr>=0.7.17,<0.8",
 	"watchdog>=0.8.3,<0.9",
 	"sarge>=0.1.4,<0.2",
-	"netifaces>=0.10,<0.11",
+	"gns3-netifaces>=0.10,<0.11",
 	"pylru>=1.0.9,<1.1",
 	"rsa>=3.2,<3.3",
 	"pkginfo>=1.2.1,<1.3",
@@ -47,8 +47,10 @@ EXTRA_REQUIRES = dict(
 	# Dependencies for developing OctoPrint
 	develop=[
 		# Testing dependencies
-		"mock>=1.0.1,<1.1",
+		"mock>=2.0.0,<2.1.0",
 		"nose>=1.3.0,<1.4",
+		"doctest-ignore-unicode>=0.1.2,<0.2"
+		"coverage>=4.1,<4.2",
 		"ddt",
 
 		# Documentation dependencies
@@ -158,6 +160,7 @@ def params():
 		"Natural Language :: English",
 		"Operating System :: OS Independent",
 		"Programming Language :: Python :: 2.7",
+		"Programming Language :: Python :: 3.5",
 		"Programming Language :: JavaScript",
 		"Topic :: Internet :: WWW/HTTP",
 		"Topic :: Internet :: WWW/HTTP :: Dynamic Content",
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ EXTRA_REQUIRES = dict(
 		# Testing dependencies
 		"mock>=2.0.0,<2.1.0",
 		"nose>=1.3.0,<1.4",
-		"doctest-ignore-unicode>=0.1.2,<0.2"
+		"doctest-ignore-unicode>=0.1.2,<0.2",
 		"coverage>=4.1,<4.2",
 		"ddt",
 
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 import logging
+import io
 
 #~~ version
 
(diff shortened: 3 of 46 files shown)


ByReaL:
copied the comment there so i dont forget about this aspect.
foosel commented a day ago
It will also need a version of the Werkzeug dependency >= 0.9 (plus possible further dependency updates), since 0.8.3 is not compatible yet to Python 3. Just FYI, you probably already noticed that though.

I'm a bit worried about that since I remember doing some hack-ish stuff with regards to the WSGI environment in order to allow uploading large files that do not fit in memory, by streaming uploaded files to disk and modifying the WSGI environment of the request to indicate to the underlying Flask app where to find the uploaded file on disk. That might run into problems, or might not, would need to be tested.

ByReaL:
any thoughts on this planed changes ? or feedback ?

foosel:
Sorry... first I was completely under water with stuff, then I needed a timeout since my batteries were fully depleted. 

1) + 2) are fine by me

3) is a high risk factor, as already indicated, more due to the werkzeug issue. Regarding the netifaces replacement, I'm a bit worried that might vanish once netifaces proper actually DOES get updated (considering the big fat WARNING on top of the PyPi page), but if that's the only way to get that stuff Py3 compliant, so be it. The only reason for the dependency is that we need all IP addresses configured for the host in order to be able to perform proper advertising for network discovery - maybe there's an easier way than netifaces existing now to achieve this?

4) "doctest-ignore-unicode" sounds good. I actually prefer doctests where possible for short helper methods since they make it way easier to keep tests updated properly with the tested code, so I'd prefer being able to keep them

5) should rather be its own PR, unrelated to this

Considering that we are now at a point where the really critical changes are needed for py3 compatibility, I'll have to ignore this PR for a while. 1.3.0 is burning a hole in my pockets, and before that isn't released and `devel` becomes 1.4.x I won't be able to find the time to properly look into this, sorry.

ByReaL:
perfectly understood, i'll do the work on my clone and i'll submit them in small pieces  after 1.3.0 is released.
the feedback is well appreciated. now know on what pieces to focus on first.

kevans91:
Other thoughts, @ByReaL 

1. There's a lot of places where map() is used and expected to return a list, rather than an iterable -- these will need to be fixed (wrapped in a list() proper?)

2. There's a lot of hashlib stuff that needs to be updated -- repr() is used for a lot of these update()s, but repr() cannot return an encoded str.

3.  Some of this stuff opens files (YAML stuff, generally) in binary mode and tries to write encoded strings to them -- which is a no-no. I'm not sure what the portable solution here is. Perhaps opening the file in binary mode if major version < 3?

4. With PyYAML, the indent= parameters are all expected to be numeric rather than the actual indentation used.

FirefighterBlu3:
@ByReaL do you have a current branch of this that pairs with 1.3.0? trying to keep a working python2 on my systems is getting painful as it interferes with things and i'd really love to stay pure py3

nunofgs:
Also interested in this. Will this be released on 1.4.0?

ByReaL:
i can do some more work on this and even commit in smaller stepspy3 support will be great
    On Friday, March 9, 2018, 5:23:48 AM PST, Nuno Sousa <notifications@github.com> wrote:  
 
 
Also interested in this. Will this be released on 1.4.0?

nunofgs:
@ByReaL any news on this?

razerraz:
To have a more complete state of what need to be done :

- all thoughts provided by @kevans91 are dawn right in fact

- Slugify needs to be imported as "from slugify import **s**lugify" and doesn't have to be first instantiated anymore. safe_chars property is not used anymore, equivalent feature is done by regex_pattern (text = slugify(text, regex_pattern=r'[^-A-z0-9._()[\] ]+'))

- flask-login provides "is_anonymous" and "is_admin" as attributes, not functions

- The minimal http server used during (at least) loading and in some plugins need import change : "from BaseHTTPServer import BaseHTTPRequestHandler" becomes "from http.server import BaseHTTPRequestHandler", SimpleHTTPRequestHandler doesn't exist anymore but BaseHTTPRequestHandler can be used instead.

- Since tornado now use asyncio, the event loop need to be created in every thread (asyncio.set_event_loop(asyncio.new_event_loop())). Especially relevant for sockjs-tornado implemetation.

- unicode/bytes/string conversion is a mess. It will be very hard to make something running on both python versions. Major modules (serial, hash) need bytes like objects nowdays

foosel:
> Slugify needs to be imported as "from slugify import slugify"

Is this a Py2.x => Py3.x change or rather a change due to a new version of slugify?

> flask-login provides "is_anonymous" and "is_admin" as attributes, not functions

`devel` already takes care of that IIRC

> [BaseHTTPServer]

The intermediary server should already default to Py3 here and only fallback to Py2 in case of an import error. The discovery plugin still was unadjusted in that regard but I just fixed that.

> [tornado and sockjs]

sockjs-tornado is bundled starting with 1.3.9 and should allow fixes like this

> [unicode vs bytes]

Yep, cause of my nightmares.

razerraz:
My mistake on slugify, I was using debian provided one (not in venv). No change at all is needed.
Additionally, using tornado 5 involve a lot of changes since you overwrite tornado.httpserver.HTTPServer, tornado.http1connection.HTTP1ServerConnection init with your 'max_body_sizes' add....

foosel:
I actually already had switched to tornado 5 once and all those changes are still somewhere in the git history. I originally planned to upgrade to tornado 5 with 1.3.9 but then ran into a memory leak issue with that, see #2585. I haven't yet gotten around to looking into this more, but as I said, the code changes are still in the history.

razerraz:
Can be the way async event loops are created into the threads.
I have 10/15% drop of memory usage comparing to py2 version, and it seems running faster, especially on initial load.

ByReaL:
broke most of the changes from here in smaller pull requests, most are very safe , few are risky i added comments in all with my thoughts on the risk

foosel:
I think this PR is superseded by all the others, right?

ByReaL:
hi,
around what date do you estimate you will be able to integrate the py23 changes,curious because i would like to grab that code and do further testing and fixes.
Thanks,Mircea Dan
    On Wednesday, December 19, 2018, 4:47:20 AM PST, Gina Häußge <notifications@github.com> wrote:  
 
 
Closed #1416.

PR #1474, opened 2016-09-03
Fix for bed depth in the printer profile not saving or loading correctly

mikenew12:

Simple bug fix in the printer profile. The setting Volume -> Y would not save after setting a new value, and in stead would reflect the Volume -> Z setting. 

Easy to test. In the current devel branch, set the print volume to X:215 Y:200 Z:210 (for example) and then save. Re-open the printer profile and it will read X:215 Y:210 Z:210. Testing with this PR should behave correctly.

#### Further notes

Just a 3 line fix. Should be pretty obvious from the file diff what was wrong.


foosel:
What an embarrassing case of typo/copy-paste-mistake, thank you!

PR #1493, opened 2016-09-12
Small Bugfix in events.py that prevented event subscribed command from execution

clemniem:
Fixed a KeyError in events.py that prevented a command associated with the ClientOpened event from execution.

The if-statement checked against currentData["job"] is not None, although it was not None even if there was no job.

I tested it with the ClientOpened and ClientClosed event.

I couldn't find any relevant tickets.


foosel:
Couldn't merge fully due to spurious merge commit, but cherry picked fb8fece on `devel`, so still merged in a way :)

PR #1502, opened 2016-09-18
Gcode visualizer color hack

sanchosk:
Hi. I made a small adjustment to the gcode visualizer.
First of all, the colors for tools are now an array of colors.
And when making the visualization, the colors go through some sort of range to visualize where the printer is printing.
The goal was to mimic the Repetier-host software.
Let me know if you like it.


sanchosk:
I made a better pull request for this including proper branch and description.

PR #1503, opened 2016-09-18
G-code color hack

sanchosk:
This time I've read the guidelines ;)
I made this small hack to enable visualization of current print progress and print head location.
The color goes from red through blue to black.
It can be adjusted by modifying the variable renderOptions.
I made all the current colors into an object of 1 child to enable the functionality for multiple tool heads.
I tested it with 2 heads and virtual printer, works OK.
I don't know how to include unit tests for JS to be honest, but I did the manual testing, including one real test print to verify performance impact.

I don't think adding myself for 5 lines of code to the list of Authors is relevant.
----

It's nice-to-have feature. As the print progresses, the G-code visualizer renders the last few lines in different color to enable better understanding of current progress/print head.

I tested manually, using real printer and using virtual printer. Tested 4 different G-codes (from small to 30MB).

I plan to implement another feature - as a plugin - to render the same G-code visualizer output to the screenshot of the timelapse ;)
But I need to learn python first :(

No tickets, just nice to have feature.

### DIFF
--- a/src/octoprint/static/gcodeviewer/js/renderer.js
+++ b/src/octoprint/static/gcodeviewer/js/renderer.js
@@ -26,7 +26,7 @@ GCODE.renderer = (function(){
         colorGrid: "#bbbbbb",
         bgColorGrid: "#ffffff",
         bgColorOffGrid: "#eeeeee",
-        colorLine: ["#000000", "#3333cc", "#cc3333", "#33cc33", "#cc33cc"],
+        colorLine: [["#000", "#001", "#002", "#003", "#004", "#005", "#006", "#007", "#008", "#009", "#00a", "#00b", "#00c", "#00d", "#00e", "#00f", "#00f", "#30d", "#60a", "#908", "#b05", "#d03", "#f01", "#f00"], ["#3333cc"], ["#cc3333"], ["#33cc33"], ["#cc33cc"]],
         colorMove: "#00ff00",
         colorRetract: "#ff0000",
         colorRestart: "#0000ff",
@@ -425,6 +425,7 @@ GCODE.renderer = (function(){
         //~~ render this layer's commands
 
         for (i = fromProgress; i <= toProgress; i++) {
+
             ctx.lineWidth = 1;
 
             if (typeof(cmds[i]) === 'undefined') continue;
@@ -452,6 +453,10 @@ GCODE.renderer = (function(){
             // current tool
             var tool = cmds[i].tool;
             if (tool === undefined) tool = 0;
+
+            // color array length inspection
+            var colorArrayLength = renderOptions["colorLine"][tool].length;
+
 
             // line color based on tool
             var lineColor = renderOptions["colorLine"][tool];
@@ -486,7 +491,9 @@ GCODE.renderer = (function(){
             } else if(cmds[i].extrude) {
                 if (cmds[i].retract == 0) {
                     // no retraction => real extrusion move, use tool color to draw line
-                    ctx.strokeStyle = pusher.color(renderOptions["colorLine"][tool]).shade(shade).alpha(alpha).html();
+                    // changing color to visualize progress
+                    ctx.strokeStyle = pusher.color(renderOptions["colorLine"][tool][(toProgress - i < colorArrayLength - 1)?colorArrayLength - 1 - toProgress + i:0]).shade(shade).alpha(alpha).html();
+                    //    ctx.strokeStyle = pusher.color(renderOptions["colorLine"][tool]).shade(shade).alpha(alpha).html();
                     ctx.lineWidth = renderOptions['extrusionWidth'];
                     ctx.beginPath();
                     ctx.moveTo(prevX, prevY);


foosel:
Hm... to be honest, I'm a bit worried this might confuse people rather than actually providing them with valuable information. Scrolling through the commands (lower slider) will already tell you the order (plus you do see it while it's being printed), and after the initial novelty wears off most people probably won't even monitor every single layer there and just wait until their print is done.

The current implementation also clashes with multi extruder support, since that already is implemented by color coding the extrusions:

image

I also had people asking for a visualization of the printing speeds through line color, and that is also still not implemented due to the very same reason. Trying to cram too much information into the colors of the visualization is not a good idea from a usability point of view I fear :/

Maybe it would rather be interesting to somehow open up the color assignment so that a) it becomes selectable (as in, do I want the separate extruders to be identified through different colors, or do I want different speeds, or do I want the progress within the layer, etc) and b) ideally also make different color coding strategies extendable through plugins? Admittedly that is a larger change though (which definitely I couldn't do nor merge at the moment since I have 1.3.0 in a feature freeze to get it ready for an RC). Does that maybe sound like an interesting alternative?

sanchosk:
By "clashes" you mean only the colors match, right? I've tested it with dual extruder virtual printer and everything was OK. But I admit I modified the color for the second tool to bright yellow (and did not commit such change, as the color was "a bit weird") :)
To do color assignment based on speed would be possible, but probably on the edge of my programming skills :(

sanchosk:
When I re-read your comment, I have to admit this would sound interesting. I'm not sure I'll be able to make it as a plug-in (as I am really lame in coding :( ) but I can try.
The goal would be to add a drop-down box to the g-code viewer tab with selectable "coloring strategy", add some sort of color array selector into settings and make both of them as a plugin.
When do you plan to release 1.3? Can you, please, advice me how can I switch to 1.3 so I can do the coding on the correct source?

foosel:
> By "clashes" you mean only the colors match, right? I've tested it with dual extruder virtual printer and everything was OK. But I admit I modified the color for the second tool to bright yellow

Yeah, it didn't clash because of that color change. See the screenshot I pasted up there. The color for the second extruder (blue) and the color for the last commands of the first extruder (also blue) are tricky to distinguish. But even with a change of the extruder colors, trying to use color coding to transfer two unrelated pieces of information (extruder and completion percentage) is just bad usability.

I'm really not convinced that the advantage of knowing "these are the last few commands of the layer" (which is information I can already gain when just looking a couple pixels down to the command slider which also shows the current progress through the layer) is big enough to sacrifice to be able to properly distinguish between multiple extruders. I'm fairly sure a lot of people with dual extruder printers out there would become very annoyed with me if I hit the merge button, sorry :/

Hence my suggestion to rather look into a (more long-term) approach of allowing things like your PR to be done as a plugin, so that people who want that kind of information but don't care about multi extruder stuff can just install and use it.

But again, that's nothing that can be done quickly since I'm currently booked solid with other things sadly, and if something like that was provided as a PR I also couldn't merge it now since have a feature freeze on `devel`.

> When I re-read your comment, I have to admit this would sound interesting. I'm not sure I'll be able to make it as a plug-in (as I am really lame in coding :( ) but I can try.

It might be a quite tricky thing to get right. I didn't necessarily meant that you should do everything about it, more that it would be the cleanest approach to allow for something like your above PR to be turned into a plugin for people who desire that kind of functionality rather than having multi extruder color coding.

> When do you plan to release 1.3? 

As soon as it's possible but only when it's done. When I'm not busy with tickets and regular maintenance currently I spent all my work time on getting it to a state to allow for a first release candidate. Currently that mostly means writing a lot of documentation and while doing that discovering some design mistakes here and there and correcting those before they become uncorrectable.

> Can you, please, advice me how can I switch to 1.3 so I can do the coding on the correct source?

1.3 is on the `devel` branch. The usual instructions on [how to switch branches](https://github.com/foosel/OctoPrint/wiki/FAQ#how-can-i-switch-the-branch-of-the-octoprint-installation-on-my-octopi-image) apply :) (potentially adjusted from OctoPi-specific paths to your local install).

Now... about this PR. Please don't hold it against me, but I really can't justify merging it considering the disadvantages it brings with it with regards to color cording already being used for multi extruder setups :/ I don't want an angry mob with pitchforks and dual extruder printers after my hide. But I did add a new feature request for the plugin option so that doesn't get forgotten: #1515

sanchosk:
No harm here. I already started to investigate how to write a plugin ;)
You are right - it deserves a separate "settings page" where you can:
- choose the color grading for each extruder
- choose the base color for each extruder
I also already modified the script a bit - when it shows complete layer, it renders all lines in the standard tool color and only renders the gradient while printing.
Re usability - I think if there is a possibility to change colors, it would actually increase the usability for the end user.
And, honestly, I just like how software like Cura or Repetier Host render the ongoing print ;)
But for now I am consumed by learning how to write a different plugin - I want my timelapses to display the g-code progress in corner - and that seems to be WAY more difficult than I expected ;)

PR #1538, opened 2016-10-07
print() uses file=, not stream= -- fix that, fix the error

kevans91:

Fix error resulting from passing stream= to print() - - print() actually takes file=. This results in a crash rather than a graceful exit when there's an error in the daemonizing process.

...I caused an error. =P Tried starting the daemon a second time when it was already running, got the error message rather than a stack trace.

Nope.

None.

Not appropriate.

#### Further notes

No further notes.


foosel:
:+1:

PR #1549, opened 2016-10-18
Update gettingstarted.rst

Pinaute:
I propose this change to avoid the following error message :

> Could not import OctoPrint's setuptools, are you sure you are running that under the same python installation that OctoPrint is installed under?

The problem happens when I run the following command :
`$ python setup.py develop`


FHeilmann:
There's also the workon command.

However this is not strictly necessary, as you merely have to call the python  executable from that virtual environment. On octopi this would be
 
`oprint/bin/python setup.py develop`

foosel:
This also has actually already been taken care of [on the `devel` branch](http://docs.octoprint.org/en/devel/plugins/gettingstarted.html) (which all PRs should go against anyhow), where all example prompts are prefixed with the virtualenv indicator and also instructions have been added right to the very beginning on how to get it there.

PR #1550, opened 2016-10-19
Add confirmation dialog to cancel to prevent accidental cancel of prints

dattas:

This adds confirmation to a destructive process (canceling a print). I've accidentally canceled a few prints on accident and want to contribute to making so it doesn't happen to other people as well

Tested via putting it on 3 of my octoprint installations (some with different plugins and such) and playing around with it. I didn't see an automation framework for the ui 

http://i.imgur.com/p7deBKF.png


foosel:
Good point, thanks!

Issue #1422, opened 2016-07-25
[Request] Add option for post roll for timed timelapse to duplicate last frame


CptanPanic:
I was trying to utilize the post-roll feature of the timelapse and realized that it does not really work as I would expect for timed timelapses.  I would like a 30 second post-roll but since I have my timer at every 10s, octoprint says it will keep recording for 30 seconds \* 25fps \* 10second interval = 125minutes.  But all I really want is to just have the last frame duplicated for the next 30 seconds like it does on the z-change timelapse.  Can we get an option for this?

mtowara:
I agree that this would be a useful addition and should maybe also be standard behavior.

foosel:
Merged to `devel`, will be part of 1.3.0rc2 (and naturally also 1.3.0)

foosel:
1.3.0rc2 is out

PR #1553, opened 2016-10-21
Add option for post roll for timed timelapse to duplicate last frame

dattas:

Adds ability to simply repeat the last frame of capture instead of capturing several minutes (or hours) of footage after print is completed.

Tested on 3 octoprint instances with 3 different cameras and some different plugins as well. It can be tested via going to timelapse, unchecking the new checkbox (ensuring there is a postroll interval) and saving, then running a print.
#1422

### DIFF
--- a/src/octoprint/server/api/timelapse.py
+++ b/src/octoprint/server/api/timelapse.py
@@ -66,7 +66,8 @@ def getTimelapseData():
 		config = dict(type="timed",
 		              postRoll=timelapse.post_roll,
 		              fps=timelapse.fps,
-		              interval=timelapse.interval)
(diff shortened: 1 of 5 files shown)


foosel:
Hmm, what I don't like here is the code duplication going on. ZTimelapse and TimedTimelapse with non captured post roll now basically contain the exact same code in process_postroll. That screams for extracting that into a custom "copying_postroll" (or better named) method on the Timelapse parent class and then just calling that from the child classes were appropriate.

Could you modify your pr accordingly? No problem if not, would just mean merging would have to wait until I'm back on my feet (which I'm currently not).

Cheers!

PR #1554, opened 2016-10-22
Adjust output of values() classmethod to not include function itself

FHeilmann:

The way values() works right now, the resulting array will also return itself. This isn't really a problem here, but if someone takes this function as a guide (like me) they can run into problems when iterating over the returned array, as it contains an element representing "values()" itself. As such, the change is more of a cosmetic nature, but since the fix is trivial I feel it makes sense.

In the plug-in I am currently developing, adjusting the output this way yields the expected result array.
#### Further notes

Another approach would be make use of the inspect package and replace name="values" with inspect.ismethod(name). This would also cover cases where other functions are added to the two classes that should be excluded in the values() call.

### DIFF
--- a/src/octoprint/printer/profile.py
+++ b/src/octoprint/printer/profile.py
@@ -34,7 +34,7 @@ class BedTypes(object):
 
 	@classmethod
 	def values(cls):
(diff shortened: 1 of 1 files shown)


FHeilmann:
Crap i messed up with the commit, let me fix it!

foosel:
Thanks!

PR #1561, opened 2016-10-26
Fixed Cura plugin's get_pos_x/y returning the wrong value when self._…

donovan6000:
Fixes an issue in the Cura plugin that incorrectly sets the posx and posy Cura Engine parameters under certain circumstances.

This can be tested by setting the current printer profile's origin to lowerleft and passing a position parameter where X and/or Y is 0 to the Cura plugin's do_slice function. Then the resulting posx and posy Cura Engine parameters can be compared to their expected values to see how this PR resolves the issue.

Python evaluates 0 as false, so the Cura plugin's get_pos_x/y functions will return machine_width/depth / 2 if self._posX/Y is 0. It should be returning the value of self._posX/Y when self._posX/Y is set to anything including 0. 

None

Not appropriate

#### Further notes
None

### DIFF
--- a/src/octoprint/plugins/cura/profile.py
+++ b/src/octoprint/plugins/cura/profile.py
@@ -835,7 +835,7 @@ class Profile(object):
 		return 1
 
 	def get_pos_x(self):
-		if self._posX:
+		if self._posX is not None:
 			try:
 				return int(float(self._posX))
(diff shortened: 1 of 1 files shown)


foosel:
Thanks!

PR #1583, opened 2016-11-07
Fixes issues with pip arguments being used when they shouldn't and not being used when they should

donovan6000:
Makes it so that installing, uninstalling, and updating plugins uses the appropriately provided pip arguments.

Testing was done by setting the pluginmanager's pip_args setting to various values, like `--user` and nothing, and seeing if I could then install, uninstall, and update a plugin without encountering any unexpected issues.

Pip wont uninstall a package if it's provided with inapplicable arguments like `--user`, `--ignore-installed`, etc, and it'll just return an error whenever that happens. So using the same user provided pip arguments to install and uninstall a plugin can result in breaking the ability to uninstall plugins.

I felt like the softwareupdater plugin should use the pluginmanager's pip_args setting when it uses pip to update a plugin. Since those pip arguments are what was needed to install the plugin in the first place then they are also needed to install the newest version of that plugin.

None

Not appropriate

#### Further notes
I left in code to make these changes easily scalable to allow adding more arguments that can be blacklisted depending on if pip is installing or uninstalling a package. As of right now it just includes blacklisting the `--user` argument when uninstalling a package since that's what I was most concerned with. I know that there's already a pip_force_user setting that deals with the `--user` argument, but this PR deals with a different issue since it's designed to fix problems with the pip arguments that are provided by the user.

### DIFF
--- a/src/octoprint/plugins/softwareupdate/updaters/pip.py
+++ b/src/octoprint/plugins/softwareupdate/updaters/pip.py
@@ -8,6 +8,7 @@ __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms
 
 import logging
 import pkg_resources
+import octoprint.plugin
 
 from octoprint.util.pip import PipCaller, UnknownPip
 from .. import exceptions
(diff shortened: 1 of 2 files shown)


src/octoprint/plugins/softwareupdate/updaters/pip.py
```diff
@@ -71,6 +72,18 @@ def _log(lines, prefix=None, stream=None):
 	if "dependency_links" in check and check["dependency_links"]:
 		pip_args += ["--process-dependency-links"]
 
+	additional_args = octoprint.plugin.plugin_manager().plugin_implementations["pluginmanager"]._settings.get(["pip_args"])
```
foosel: Hm... I don't like that. I know why you are doing it, and the reason is valid, but it couples both plugins and that should be avoided if possible.

I'm wondering if it would be a better idea to maybe extract the whole pip configuration from the plugin manager and into a general platform wide configuration, to be used by both the plugin manager and the software update plugin. Basically taking the extraction of the pip helper one step further. Would necessitate a migration step for the config.

donovan6000:
Updated the PR to use the recommendations in your code review.

I removed the changes to the softwareupdate plugin since you prefer that the plugins remain independent of one another.

foosel:
Squashed as new single commit 1b9bfc6 and merged to `devel`. Thanks!

PR #1617, opened 2016-12-02
Make M600 look like a tool change.

ryanneufeld:

This PR detects the M600 command used by Prusa flavored firmwares and replaces the command with either a T0 or T1 command. This allows users of Prusa's Colorprint utility to visualize when in the print they can expect a filament change to take place.

I tested by loading this [gcode](https://gist.github.com/ryanneufeld/419f5b7e0ce9ea759bbb2abf3f2f7e93) file and using the slider on the right to watch the colors change as I jogged through the layers.

In short, it was manually tested.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -74,6 +74,7 @@ date of first contribution):
   * [Mathias Rangel Wulff](https://github.com/mathiasrw)
   * [Clemens Niemeyer](https://github.com/clemniem)
   * ["I-am-me"](https://github.com/I-am-me)
+  * [Ryan Neufeld](https://github.com/ryanneufeld)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/static/gcodeviewer/js/Worker.js
+++ b/src/octoprint/static/gcodeviewer/js/Worker.js
@@ -267,6 +267,15 @@ var doParse = function () {
 
         var log = false;
 
+        if (/^(?:M600)/i.test(line)) {
+            console.log("Replacing M600 with tool change");
+            tool++
+            if (tool > 1) {
+                tool = 0;
+            }
+            line = "T" + tool;
+        }
+
         if (/^(?:G0|G1|G2|G3)\s/i.test(line)) {
             var args = line.split(/\s/);
 


markwal:
I think this would be better done in a plugin.  Less likely to be a compatibility problem for some printer that we're not testing.  Doing gcode replacement in a plugin is easy.  See http://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-comm-protocol-gcode-phase for documentation on the gcode hook.

ryanneufeld:
Actually, with what you linked, this will not work.  The point here is only to change the visualization. Not the actual code sent to the machine.

markwal:
Oh, I see.  Just want to make M600 look like a tool change.

ryanneufeld:
I'm open to other options, this was my first hack at it.

foosel:
Adding some understanding and visualization of M600 to the GCODE viewer is a very good idea. I'm not so sure though if having it basically be a fake tool change is the best approach. Consider the scenario where someone with multiple extruders is also utilizing M600 in their GCODE. Stuff might look completely off then, since an actual tool change usually also involves a certain X and Y offset to compensate for a different nozzle position. And the filament measurements in the GCODE viewer would also be mislead that way.

Maybe it would be a better approach to introduce some special symbol for the M600, similar to how retracts and extrude starts are now rendered? For that it would probably be necessary though introduce a new field in the parsed command representation, so that the rendered knows what to do.

Thoughts?

ryanneufeld:
That sounds like a better idea. And to be fair I had originally considered
something like that.

What I personally liked about the tool change approach is that when moving
through the layers in the viewer you can see when the change happens. And
for how long that change is active. I was searching for a way to show that
all the following layers were of a different color.

M600 is effectively a tool change on a single extruder machine.

Perhaps a "Treat M600 as tool change" option would work. In this way the
user can choose how they want it displayed.

http://reprap.org/wiki/G-code#M600:_Filament_change_pause

I did not include the extra information that could be sent with a filament
change. So that needs improvement too.

ryanneufeld:
@foosel what are your thoughts on this?

foosel:
Hm, the conflict with actual multi extruder setups and concurrent usage of M600 (which is a valid scenario, you might want to swap out your model filament in tool 0 but leave your soluble support material in tool 1 as is). That way the visualization would go completely off the rails (and the next bug report would be guaranteed ;)).

I understand your reasoning behind wanting it to show in a different line color, but it basically makes the gcode viewer lie :/ An M600 is an event at a specific time, but what that event causes for the print is entirely up to the printer operator. People might be using it to swap another color in, or another material of the same color or even just a fresh spool of the exact same color and material. We don't know that. We are assuming it's a color change, but it might be something else.

I'll happily merge this with a special symbol to denote an event, but I really don't want to intentionally break tool visualization since I'm 100% sure it will cause issues for users with multi extruder setups and (rightful) bug reports as a consequence of that.

ryanneufeld:
I thought you'd feel that way.

What would be the best way to add a symbol? I'm not super familiar with
three.js

--Ryan Neufeld BSc CIS
ryan@neufeldmail.com

foosel:
@ryanneufeld no three.js involved here, just painting on a HTML5 canvas :) You'll have to adjust two things for a special symbol. 

First of all you'll need to extend the data model used for the parsed command representation to include something like a "filamentchange" parameter. The code in question is located [here](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/gcodeviewer/js/Worker.js#L502-L520).

Then in the renderer you'll need to check for that parameter and if it's there do something with it. [Here](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/gcodeviewer/js/renderer.js#L466-L476) is the code that is responsible for drawing the retract dots for example - you'll want something similar to that I guess.

As a heads-up, I'll be on vacation starting Saturday and will do a real two week timeout from all OctoPrint related things starting next Wednesday, to recharge the batteries.

foosel:
@ryanneufeld are you still looking into this?

ryanneufeld:
I was. But haven't had time to investigate much further.

I was thinking an arbitrary gcode highlight feature would be better than
this specific use case.

Like provide a list of commands that you want highlighted/indicated and
choose a marker of some kind for each.

foosel:
That actually sounds like a nice idea. Sounds more like it might make sense to close this for now though, or what do you think?

Currently doing a long overdue spring cleaning in issue tracker and PR list, stuff became overwhelming, hence why I'm asking all of a sudden.

ryanneufeld:
Yeah, you can close this then.

--Ryan Neufeld BSc CIS
ryan@neufeldmail.com

PR #1639, opened 2016-12-10
Docker

gaetancollaud:

This PR add the docker description files to run Octoprint in a docker environment.

I build the image and tested in with my 3D printer (Ultimaker).
I printed a file successfully and generated a timelapse.
I destroyed and created a new container using the same volume and everything came back (configuration, uploaded files and timelpases).

Using docker can reduce the setup time. If someone just want to give it a try he can run OctoPrint very quickly without going through the installation (I also included ffmpeg for timelapses).

Additionally, using container can help using octoprint with multiple printers at the same time on the same machine. See the example in the docker-compose file, I mapped /dev/ttyACM0 to /dev/ttyS0 and I use the port 5000. But imagine you have multiple printers, you can for instance:
* run one container using /dev/ttyACM0 on port 5000
* run one container using /dev/ttyACM1 on port 5001
* etc.

Additionally using proxy server like [traefik](https://traefik.io) you can easily add domain name to your printer. For instance:
* /dev/ttyACM0 is ultimaker1.my-fablab.ch
* /dev/ttyACM1 is ultimaker2.my-fablab.ch
* etc.

None

No screenshots

#### Further notes

Currently I'm using my personal dockerhub account ! Feel free to change it to yours.

Building and pushing the image to dockerhub:
```
docker-compose build
docker-compose push
```
Running the image using docker compose
```
docker-compose up
```
Running the image without docker compose
```
docker run -it --volume /my/config/folder:/root/.octoprint --device=/dev/ttyACM0:/dev/ttyS0 -p 5000:5000 gaetancollaud/octoprint
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM python:2.7.12
 EXPOSE 5000
 
-RUN git clone https://github.com/foosel/OctoPrint.git /opt/octoprint
-#ADD * /opt/octoprint/
+ADD ./ /opt/octoprint/
 RUN cd /opt/octoprint && python setup.py install
 
 CMD ["octoprint", "--iknowwhatimdoing"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:2.7.12
+EXPOSE 5000
+
+RUN git clone https://github.com/foosel/OctoPrint.git /opt/octoprint
+#ADD * /opt/octoprint/
+RUN cd /opt/octoprint && python setup.py install
+
+CMD ["octoprint", "--iknowwhatimdoing"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM python:2.7.12
 EXPOSE 5000
+VOLUME /root/.octoprint
+
+RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz && mkdir -p /opt/ffmpeg && tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
 
 ADD ./ /opt/octoprint/
 RUN cd /opt/octoprint && python setup.py install
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,23 @@
 FROM python:2.7.12
 EXPOSE 5000
-VOLUME /root/.octoprint
+VOLUME /user/octoprint/.octoprint
 
-RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz && mkdir -p /opt/ffmpeg && tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
+WORKDIR /opt/octoprint
 
+#install ffmpeg
+RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz \
+	&& mkdir -p /opt/ffmpeg \
+	&& tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
+
+#Copy repo to /opt/octoprint
 ADD ./ /opt/octoprint/
-RUN cd /opt/octoprint && python setup.py install
 
-CMD ["octoprint", "--iknowwhatimdoing"]
+#Perform installation
+RUN virtualenv venv \
+	&& ./venv/bin/python setup.py install
+
+#Create an octoprint user
+RUN useradd -ms /bin/bash octoprint
+USER octoprint
+
+CMD ["/opt/octoprint/venv/bin/octoprint"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,10 @@
+version: '2'
+services:
+  octoprint:
+    build: .
+    image: gaetancollaud/octoprint
+    container_name: octoprint
+    ports:
+      - 5000:5000
+    devices:
+      - /dev/ttyACM0:/dev/ttyS0
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     container_name: octoprint
     ports:
       - 5000:5000
-    devices:
-      - /dev/ttyACM0:/dev/ttyS0
+#    devices:
+#      - /dev/ttyACM0:/dev/ttyS0
 #    volumes:
 #      - /yourDataFolder/octoprint:/root/.octoprint
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,3 +8,5 @@ services:
       - 5000:5000
     devices:
       - /dev/ttyACM0:/dev/ttyS0
+#    volumes:
+#      - /yourDataFolder/octoprint:/root/.octoprint


mathewpeterson:
I think there should probably be named volumes for uploaded files. Also, how do the timelapses work with this container?

HWiese1980:
Just curious, is this still multi platform? Does the --device parameter also work on Docker for Windows for instance?

gaetancollaud:
@mathewpeterson In my last commit I added a volume for the configuration files. Do you now where the uploaded files are stored ? I'm new with OctoPrint and I don't know it very well yet. For the timelapse, I added ffmpeg in the container.

@HWiese1980 I only use linux but if you have docker for windows could you give it a try ? But normally it should work.

I see that you can use cura to upload stl files and make the slicing directly in OctoPrint. Shall I include the Cura engine in the container ?

Renha:
personally I think that cura is even more important than ffmpeg

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:2.7.12
+EXPOSE 5000
+VOLUME /root/.octoprint
+
+RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz && mkdir -p /opt/ffmpeg && tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
+
+ADD ./ /opt/octoprint/
+RUN cd /opt/octoprint && python setup.py install
+
+CMD ["octoprint", "--iknowwhatimdoing"]
```
foosel: Do I understand it correctly that this setup is basically enforcing to run OctoPrint as root and that's why that parameter is required? This is a bad idea then. Container or not, a user space program should never be run as root, and that option only exists for very very specific usecases where this can absolutely not be avoided - it should never be something set by default and really only be an individual user's/admin's choice, ideally a user/admin who indeed knows what they are doing in running that server as root. To be able to merge this we'll need to find a way around this.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:2.7.12
+EXPOSE 5000
+VOLUME /root/.octoprint
+
+RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz && mkdir -p /opt/ffmpeg && tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
+
+ADD ./ /opt/octoprint/
+RUN cd /opt/octoprint && python setup.py install
```
foosel: If I understand this correctly (sorry, no experience with Docker myself), this will basically copy the sources to `/opt/octoprint` and then run a global install? That way you'll get OctoPrint twice - once as a pure source checkout and once installed "globally" (well, container wide I guess?) in Python's site-packages (not under `/opt` I think). The copy under `/opt/` will have basically no function. I'm not sure this is the best/cleanest approach here, especially since the official installation instructions pretty much mandate the usage of a virtual environment. Yes, I know, in this containerized setup that currently is not necessary, but if this file is to be merged into the official source repository (and I have to admit I'm still not entirely sure whether that's the best approach and if it wouldn't rather make sense to put that into a different repo that then pulls in the source tree) it should not go against all recommendations in the docs.

foosel:
First a heads-up: I have next to no experience with Docker.

That being said, from my understanding of this approach, there are a couple of things that are still problematic: 

  * OctoPrint runs as root - never run user space software as root, containerized or not. Bugs exist, and Docker is not a security sandbox. The parameter that allows that is called ``--iknowwhatimdoing`` for a reason, it should be an individual choice on a per-case basis by a knowledgeable admin, not a default in any kind of setup.
  * Setup of OctoPrint goes against official instructions by installing into global python site packages - yes, I know, container, no issue installing globally, but contradicting oneself like that leaves a bad taste.
  * No working software update out of the box - for that at least the checkout folder will need to be configured in a pre-filled config.yaml. Update mechanism currently also relies on a local checkout (here `/opt/octoprint`) and git being present (not sure that's working in this setup). I want to get away from this but currently it's still the state of the art. How would a user update this when a new release gets pushed?
  * Static compiled ffmpeg binary from a source at least unknown to me. Also limited to 32bit this way. The first thing is the bigger red flag here and to be honest gives me the heeby jeebies (especially with the root-user thing mentioned in item 1).

Additionally, while I love the idea of an all-in-one Docker build for OctoPrint, I'm not sure about having the Dockerfile actually live in the repository. Simply because that will mean that next thing I know I'll also have a vagrant buildfile to maintain, and whatever else might come up in the next couple of years. Now this is where my lack of knowledge about Docker clearly shows, but would it maybe be possible to instead have a docker build that pulls in the sources from this repo? I'd be happy to provide a repository under the OctoPrint organization for that.

That would also make more sense for the setup that would be recommended once I finally get around to shifting to a PYPI based deployment of the releases (which is where I actually want to get in the near future - having to checkout a git repo when you are only a user is ugly compared to `pip install OctoPrint`, plus it would allow better tooling support for the actual packaging than what I can use now since it would only be developers who'd ever see the actual checked out sources anymore and one can expect from them to install things like less compilers and such).

Dockerfile
foosel: That would be cleaner I think. Running everything as root is scary ;)

gaetancollaud:
Thanks for your feedback. I will try to take a moment this week-end to address all the issues that you come up with.

- Docker as root : I understand your point and I will try to use a standard user
- Install : Sorry if I didn't follow the guideline. When I saw a install.py I just ran it.
- Update : The user will still have the notification. But I will be cleaner to redownload the the image rather that updating the current container. Since the configuration and other important data are persisted in a docker volume, nothing will be lost (even with a new image/container).
- Static source for FFMPEG : I use this because it was very simple a very quick to install ffmpeg, but I can install it differently if you want
- Repo: It's up do you. We can easily take files from other repo (and branch/tags) when we build the image. The only downside is that we must release this repo and then the docker repos.

gaetancollaud:
I changed the installation process to use the one described in the README. The service now run with the _octoprint_ user.

@foosel Have you decided if you want to merge this in this repository or put it in another ?

harakiro:
+1 imo for this as it adds a clean way to support multiple printers. I was looking at running docker with a similar approach on my raspberry pi and ran across this.

taxilian:
Docker support sounds awesome, though I'll stick with using a raspberry pi per printer.

Those interested in this ticket may find an example of a) how to set up a nginx vhost for all printers on the same domain and b) how to put basic printer stats and webcams on a single "dashboard" page useful.  https://gist.github.com/taxilian/87e83fd9e4d9c1a4408f2fb67524a516

If not, forgive the spam =] just finished it and was reading through issues. My nginx is running via nginx-proxy on a docker host, if that helps :-P

foosel:
@gaetancollaud I was on a long needed vacation for the past three weeks, with strict OctoPrint abstinence, so I only now get around to replying, sorry.

I'm still of the opinion that a separate repository probably makes way more sense to keep this repository to be overrun with configurations for every build system and/or container solution under the sun, especially since I want to go towards a general distribution on pypi anyhow. The init script currently bundled with OctoPrint's sources will be moved away in the longer run for this very reason too.

My offer with regards to a repository for this under the OctoPrint Github organization stands - what would make sense here, having a dedicated repo called something like "Dockerfile"? As mentioned before, I have next to no experience with Docker and also no time to look into maintenance for this on top of everything else already on my plate sadly, so I'd also need you to maintain this down the road (or find another volunteer to do that), otherwise it won't make much sense to have it officially hosted. Hope you understand were I'm coming from here - I really appreciate the work, I just have to think about maintenance and the long term as well!

Also I realized I still owe you an answer with regards to the ffmpeg static stuff. I didn't realize earlier that the URL you use there is actually the one for static builds linked from ffmpeg.org - so probably as official as it gets. In that case ignore my complaint ;)

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:2.7.12
```
hypergig: no alpine love?

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:2.7.12
+EXPOSE 5000
+VOLUME /user/octoprint/.octoprint
+
+WORKDIR /opt/octoprint
+
+#install ffmpeg
+RUN  wget -O ffmpeg.tar.xz https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-32bit-static.tar.xz \
```
hypergig: Can you checksum the artifact before moving forward? Just to confirm the integrity of the download? See here for an example of this:
https://github.com/docker-library/docker/blob/05678f7f3198d0787d78de9bdc084dcffccf7276/1.12/Dockerfile#L14

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  octoprint:
+    build: .
+    image: gaetancollaud/octoprint
+    container_name: octoprint
+    ports:
+      - 5000:5000
+#    devices:
+#      - /dev/ttyACM0:/dev/ttyS0
```
hypergig: nice!

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+	&& mkdir -p /opt/ffmpeg \
+	&& tar xvf ffmpeg.tar.xz -C /opt/ffmpeg --strip-components=1
+
+#Copy repo to /opt/octoprint
+ADD ./ /opt/octoprint/
+
+#Perform installation
+RUN virtualenv venv \
+	&& ./venv/bin/python setup.py install
```
hypergig: Are there any post cleanup steps that can be taken here to slim down the image?

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+services:
+  octoprint:
+    build: .
+    image: gaetancollaud/octoprint
```
hypergig: Why not just have the build directive?

gaetancollaud:
@foosel NP everybody deserve vacation.

I think that having another repos is the best things to do. This way the main repo remain clean on any installation/configuration files. If you make a repos. call it something like _Octoprint-docker_ or simply _docker_ but not _Dockerfile_ (which is the name of the file used to define a container).

I understand that it's something you will have to maintain but I think it's worth it. Docker is becoming more and more popular and for people which just want to give Octoprint a try it will be setup in no time. I'm volonteer to help you and maintain the docker repo. It's not tomorrow that I will throw my 3D printer away. And if it's okay for you I was thinking about using a CI system like Travis to build the image. This way when you make a new release you just have to press a button to release the docker image as well. 

I will try to install ffmpeg and CuraEngine using the official installation manual.

foosel:
@gaetancollaud You should have gotten an invite for the OctoPrint/docker repository. What else do you need?

gaetancollaud:
@foosel  Thanks ! I commited my progress so far. 

About the automation : I'm didn't use Travis a lot but there is a good documentation, I'm more used to Jenkins. I see that you create a new tag for each realease is that correct ? A good starting point would be to create one docker image per tag (with the tag _:latest_ pointing to the latest realease). Later we could also create an image each time something is commited on develop.

I will first make some tries with Travis and Docker on [one of my project](https://github.com/gaetancollaud/fablab-manager) to avoid making to much useless commits on your repo.

gaetancollaud:
@foosel I hope you will read this otherwise I will try to contact you some other way. 

I requested the access to travis to for the repo https://github.com/OctoPrint/docker (you shall have receive something)

I wanted to create an _octoprint_ organization in dockerhub, but unfortunatelly someone already took it. Maybe it's you ? https://hub.docker.com/u/octoprint/

I made some progress regarding the automation. I think we can easily add an `after_success` script to the travis file of this repos to trigger the docker repo build. This way a docker image can be build on each commit on develop/master or on each tag creation.

foosel:
Quick feedback: should have granted access already for Travis, hope that worked out. Dockerhub organisation named octoprint is in fact mine, can do whatever is necessary to get that stuff setup to access images. Only problem right now is I'm out sick, so I can't do much at the moment until I'm back on my feet :/

gaetancollaud:
I cannot enable travis since I'm not an administrator... Maybe you can [do it for me](https://travis-ci.org/profile/OctoPrint)
> Sorry, but we can't find any repositories you have admin access to.
> You require admin rights to enable these repositories

It's great if it's yours !

I received the message on Patreon about you behing sick. We're not in a hury. Take our time to get better we can sort things out later.

Do you want to have another mean of cummunication for this "administrative" stuff ? For instance [email](mailto:gaetancollaud@gmail.com) ?

foosel:
Gave you admin access and also shot you an email :)

PR #1654, opened 2016-12-14
Include LSB above first use in init.d script

tullo-x86:
Fixes a late inclusion of the LSB functions in Octoprint's `init.d` script so that the include is executed before the first call to "log_warning_msg"

Follow the repro steps described in [Issue 1653](https://github.com/foosel/OctoPrint/issues/1653).

Not sure why the init.d script only barfs on the most recent BBB image; OctoPrint v1.3.0 worked fine on an old image it was running from 2 years ago (which I only updated because the GPG keys of the old APT repository expired).

[Issue 1653](https://github.com/foosel/OctoPrint/issues/1653)


foosel:
Thanks for that. Weird that this didn't produce issues earlier. I also pushed the fix on the `maintenance` branch so this will go into 1.3.1

PR #1665, opened 2016-12-19
Octoprint /dev/ttyS* addition

jammi:
OctoPrint wasn't looking for /dev/ttyS* devices, so I added them. Tested on CubieBoard vs a Melzi-based printer.


Salandora:
You can easily add additional ports through the settings page :-)

jammi:
Sure, but why have defaults at all if you don't include all the common serial ports? /dev/ttyS* is the the common one on Linux, just like the already included /dev/cu.* ones are on BSD's.

foosel:
Fine for me, generic enough to warrant an inclusion by default. The additional port list is rather for things like `/dev/myCustomWeirdDeviceName`.

Thanks!

PR #1673, opened 2016-12-24
from __future__ imports must occur at the beginning of the file

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/src/octoprint/util/__init__.py
+++ b/src/octoprint/util/__init__.py
@@ -1,10 +1,10 @@
 # coding=utf-8
 from __future__ import absolute_import
 
-"""
-This module bundles commonly used utility methods or helper classes that are used in multiple places withing
-OctoPrint's source code.
-"""
+
+# This module bundles commonly used utility methods or helper classes
+# that are used in multiple places withing OctoPrint's source code.
+
 from __future__ import absolute_import, division, print_function
 
 __author__ = "Gina Häußge <osd@foosel.net>"


foosel:
While the *duplication* of `__future__` import lines here indeed is a mistake that should be remedied, in general this [should not be causing an error as is](https://docs.python.org/2/reference/simple_stmts.html#future):

> A future statement must appear near the top of the module. The only lines that can appear before a future statement are:
>
>   * the module docstring (if any),
>   * comments,
>   * blank lines, and
>   * other future statements.

Did you run into any errors with this or was this PR triggered by purely aesthetical reasons? I'm asking because there are a bunch of lines where the module docstring (which the string you turned into a comment here is) preceeds the ``__future__`` import throughout the OctoPrint source, which is legal according to the above section of the python docs and is causing no problems at all for me. But if that's causing issues with any Python versions out there I'd like to know to be able to look into it instead of just changing this one occurrence and leaving everything else as is.

patrickelectric:
Python 3 don't allow code before import, I am running octoprint with some changes in a raspberry pi with archlinux.

foosel:
A module docstring (the thing that you moved over to be a comment in this PR, which is not the correct way to handle that) is not code that should cause issues with future imports. Python's own docs as quoted above say so. I also just checked and the same statement is present in the python 3 docs. The part your PR adjusted here should not have produced an error according to those docs since the future imports in question were only preceded by comments, white space, the module docstring and other future imports. So please, could you show me the exact error you are seeing? I only want to understand it, because if the current implementation raises an error with your python version, that version is either not adhering to python's own documentation, the documentation is wrong or I'm misunderstanding it completely and I'd like to clarify which of those it is without a doubt to avoid similar issues in the future.

I'm btw surprised that you can get OctoPrint to run at all with python 3, it explicitly requires python 2.7 at the moment. I strongly suggest to adhere to that requirement, I cannot provide support for python 3 incompatibilities at this point, and stuff will probably be broken one way or the other.

foosel:
@patrickelectric Still waiting for the full error you were seeing here (and also please the exact Python 3 version because - again - according to Python's official documentation this should NOT be producing an error as is, see the quoted bits above).

patrickelectric:
@foosel if you agree I will close this PR because I am without time to confirm this problem.

foosel:
Fine by me :)

PR #1702, opened 2017-01-10
Fix issue in getting file metadata after slicing

derpicknicker1:
This fixes the issue that there were no informations about filament
usage in metadata after slicing with cura plugin. Trying to call
profile.get_float("filament_diameter") ended in en exception with
message " 'module' object has no attribute 'get_float' ". So i defined
profile before using profile and now it works.
See issue #1685
Also inserted a check to determine if filament usage is > 0 to exclude tools with no filament usage in metadata.

### DIFF
--- a/src/octoprint/plugins/cura/__init__.py
+++ b/src/octoprint/plugins/cura/__init__.py
@@ -372,10 +372,12 @@ class CuraPlugin(octoprint.plugin.SlicerPlugin,
 									analysis = dict()
 								if not "filament" in analysis:
 									analysis["filament"] = dict()
-								if not tool_key in analysis["filament"]:
+								if not tool_key in analysis["filament"] and filament > 0:
 									analysis["filament"][tool_key] = dict()
 
-								if profile.get_float("filament_diameter") != None:
+								profile = Profile(self._load_profile(profile_path), printer_profile, posX, posY)
+
+								if profile.get_float("filament_diameter") != None and filament > 0:
 									if profile.get("gcode_flavor") == GcodeFlavors.ULTIGCODE or profile.get("gcode_flavor") == GcodeFlavors.REPRAP_VOLUME:
 										analysis["filament"][tool_key] = _get_usage_from_volume(filament, profile.get_float("filament_diameter"))
 									else:
--- a/src/octoprint/plugins/cura/__init__.py
+++ b/src/octoprint/plugins/cura/__init__.py
@@ -372,12 +372,10 @@ class CuraPlugin(octoprint.plugin.SlicerPlugin,
 									analysis = dict()
 								if not "filament" in analysis:
 									analysis["filament"] = dict()
-								if not tool_key in analysis["filament"] and filament > 0:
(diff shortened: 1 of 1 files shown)


derpicknicker1:
Now it worked. Sorry again :sweat_smile:

foosel:
Ignore the bot, I don't want you to have to do it yet again. But please don't push anything else to your version of `devel` until I've merged this (because it would become part of this PR) :)

derpicknicker1:
Instead of loading the profile to the variable ``profile`` you can use ``engine_settings`` which is loaded with the printer profile in line [269 ](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/plugins/cura/__init__.py#L269)
```python
269 |     engine_settings = self._convert_to_engine(profile_path, printer_profile, posX, posY)
```

Then the code in the line [378](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/plugins/cura/__init__.py#L378) (line [380](https://github.com/foosel/OctoPrint/pull/1702/files#diff-1a6055cacd6d027c47bb6e3fc15e6278R380) in this pull request) is the following:
```python
if engine_settings["filamentDiameter"] != None and filament > 0:
	if engine_settings["gcodeFlavor"] == GcodeFlavors.ULTIGCODE or engine_settings["gcodeFlavor"] == GcodeFlavors.REPRAP_VOLUME:
		analysis["filament"][tool_key] = _get_usage_from_volume(filament, float(engine_settings["filamentDiameter"])/1000)
	else:
		analysis["filament"][tool_key] = _get_usage_from_length(filament, float(engine_settings["filamentDiameter"])/1000)
```

foosel:
I've squashed and merged this unto `maintenance`(since it's a bug fix it definitely should go into the next release ASAP, hence the backport), see 58f934e2c1c891d2c579d116e1a19c56a1ac93b6. That means though that Github doesn't detect that this was in fact merged, so please don't be alarmed that you probably got a "closed" instead of "merged" notification for this PR.

I removed the 0-check for filament length again (and renamed `profile` to `slicing_profile`, to stop the var from shadowing an import, see f20e985d0da3d197dabe350cda67ead4cff0117d), because with the check present the data parsed from the slicing process would differ in structure from data for the same file when processed through the internal gcode analysis implementation.

Thanks for spotting this btw - that was a bug introduced by the removal of some code that was outdated and I apparently removed a bit too much ;)

derpicknicker1:
The metadata differs from internal analysis if you NOT do the check. Without the check, tool1 will always be included in metadata, even it is not used. 
Normal analysis of uploaded files will only include tool0 when tool1 is not used in the gcode.

foosel:
Ah, you are right. I was thinking more along the lines of "if the tool is selected but nothing gets ever extruded or we only have retractions, that should be reflected properly". But of course, if the tool never even gets selected, it won't show up in the internal analysis. 

I'm a bit undecided now how to handle it, both leaving it in AND leaving it out if the value is 0 is wrong in some way. Hmpf.

PR #1703, opened 2017-01-11
Changed old comments to not claim a method does more than it really does

noahsmartin:
Because comments are there to clear up confusion, might as well start here.

I just started looking through OctoPrint source and noticed this comment seemed to not be true anymore so I thought I'd fix it.  poll_temperature and poll_sd_status are called by a RepeatedTimer now, the methods don't re-enqueue themselves.


foosel:
Good point, thanks! :+1:

PR #1705, opened 2017-01-11
Simplified install process on Mac

noahsmartin:
I was trying to install OctoPrint on macOS by following the steps in the README when I ran into the problem "No module named AppKit", the same as here: https://github.com/foosel/OctoPrint/issues/1565.  It turns out OctoPrint was using NSSearchPathForDirectoriesInDomains from AppKit to get the Application Support directory.  This required an extra step of installing pyobjc.

However, pyobjc isn't that easy to install now because of System Integrity Protection (http://apple.stackexchange.com/questions/209572/how-to-use-pip-after-the-os-x-el-capitan-upgrade).  Hopefully Mac users could get OctoPrint set up without this extra step at all.  So I decided to simplify and use appdirs (https://pypi.python.org/pypi/appdirs) instead.  This can be installed easily in setup.py and does exactly what we need by finding the same Application Support directory.

After this change I was able to install and run OctoPrint on my Mac.

### DIFF
--- a/THIRDPARTYLICENSES.md
+++ b/THIRDPARTYLICENSES.md
@@ -55,3 +55,4 @@
   * [SockJS-Tornado](http://github.com/mrjoes/sockjs-tornado/): MIT
   * [Tornado](http://www.tornadoweb.org/): Apache License 2.0
   * [watchdog](http://github.com/gorakhargosh/watchdog): Apache License 2.0
+  * [appdirs](http://github.com/ActiveState/appdirs): MIT
--- a/docs/development/index.rst
+++ b/docs/development/index.rst
@@ -145,7 +145,6 @@ You'll need a user account with administrator privileges.
        cd OctoPrint
        virtualenv venv
        source venv/bin/activate
-       pip install -U pyobjc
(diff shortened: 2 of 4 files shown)


foosel:
Hm.. Can we wrap this additional dependency into a platform specific check so it only gets installed on Macs? Most common OS for running OctoPrint on is Linux, so I'd rather avoid always pulling in a Mac-specific dependency.

Something like a simple `if sys.platform == "darwin"` in `setup.py` extending the requirements should probably already do it. I can't test that myself though since I don't have access to a Mac anymore.

noahsmartin:
Yeah I like that idea, updated the commit.  I also tested this on 10.12.1 and 10.11.6

foosel:
Great! I've cherry picked this on `maintenance`, I think it merits a fast release to the public :) It will be part of 1.3.1.

Since Github won't pick up on this cherry-pick-based merge, you'll get a "closed" message instead of a "merged" message, please don't be alarmed by this.

I also took the liberty and added your name to the list of authors.

Thanks!

PR #1706, opened 2017-01-12
Added a mask icon for Safari pinned tab and touchbar

noahsmartin:
This change adds the mask-icon to OctoPrint.  Safari uses this icon, as well as the color, to display an image in pinned tabs and the touchbar.  Apple's docs explaining this are here: https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html

With this change it just looks a little nicer to have OctoPrint as a pinned tab or a favorite site in the touchbar.

I made this icon just by tracing a png because the new icon needs to be a black and transparent svg.  If anyone already has an svg of the OctoPrint icon I'd be happy to replace it. 

Before as a pinned tab:
screen shot 2017-01-11 at 6 14 57 pm

After as a pinned tab:
screen shot 2017-01-11 at 1 52 12 pm

On the left is the new touchbar icon for favorites, and the right is the old one.  As you can see, the new icon matches more the expected style of other sites.
touch bar shot 2017-01-11 at 6 12 45 pm


foosel:
Nice!

Regarding the SVG, I did [this one](https://drive.google.com/open?id=0ByuNfeP3KbSsOFpxRzZNbFVJcW8) a while ago for a different use case, but I fear the thin line for the differently colored part of the tentacle doesn't really work at that size:

image

[Filling in that part](https://drive.google.com/open?id=0ByuNfeP3KbSsUzljSXNBTFhoQk0) looks ok though:

image

Both those screenshots are from Safari 9.1 under El Capitan (on BrowserStack, no actual machine access here) btw. Testing against Safari 10 on Sierra didn't show an icon, neither with one of mine nor your's embedded - any ideas what could be up here?

image

noahsmartin:
Thanks! I updated the SVG, here is how the new icon looks on the touch bar:

touch bar shot 2017-01-12 at 10 54 36 am

Safari has a separate cache for template icons that needs to be cleared, try deleting the contents of `~/Library/Safari/Template Icons` if that is something BrowserStack will let you do.  There is also a file, CacheSettings.plist, in that directory that can tell you exactly what image and color is cached.

I tested this on 10.12.2 with Safari 10.0.2 and on 10.11.6 with Safari 10.0.1

foosel:
I still can't get the Safari 10 on Browser Stack to show the icon and sadly as you already feared BrowserStack won't let me clear that cache. I decided to just trust you on this ;) and have cherry-picked the commit to `maintenance` to get it among people ASAP. It will be part of the 1.3.1 release.

As already mentioned in #1705, Github doesn't know how to handle cherry picks and will therefore report this PR as "closed" instead of "merged". Please don't be alarmed by this.

PR #1741, opened 2017-01-31
Update index.jinja2

Vanclec:
Many plugins will add informations in the navbar, resulting of a degraded appearance. Using .container-fluid instead of .container would allow to show more data into the navbar, since many screens nowadays have a larger width.

### DIFF
--- a/src/octoprint/templates/index.jinja2
+++ b/src/octoprint/templates/index.jinja2
@@ -19,7 +19,7 @@
         <div class="page-container">
             <div id="navbar" class="navbar navbar-static-top">
                 <div class="navbar-inner" data-bind="css: appearanceClasses">
-                    <div class="container">
+                    <div class="container-fluid">
                         <a class="brand" href="#"> <span data-bind="text: appearance.brand">OctoPrint</span></a>
                         <div class="nav-collapse">
                             <!-- Navbar -->


foosel:
Sorry for the late reply, I was down sick the past two weeks and am still working through the backlog from that time.

I see two problems with this change:

  1. The logo + title and the default controls on the right no longer line up with the left and right side of the actual app - that looks just completely inconsistent and bad to me:

     image

     vs
     
     image

  2. If there are notifications, they will now *always* cover the navbar elements on the right side. So those at least need to be moved down:

     image

The second point is an issue that should be solved in any case.

The first one however I don't see solved easily without basically refactoring the whole frontend to be fully responsive. Which admittedly is something that I'm thinking about from time to time, but it is very very disruptive and might even be an outright breaking change for other plugins, depending on how they manipulate the interface.

I agree that there needs to be a solution to multiple navbar plugins each cramping something more or less large into the limited space up there, and I know what you mean with "degraded appearance". But I'm not happy with this particular solution to this problem either since it just makes the interface look inconsistent and basically makes it semi-responsive.

For the tab bar I solved the issue of too many plugins via the little "overflow" menu you also see in the above screenshots. Maybe something like that would also be an option here? I can also look into making the text of the default menus in there smaller, that will free up a bit of space. But plugin authors also need to limit themselves a bit with regards to the real estate they take up there. I've seen some plugins that really have a huge foot print, and I can't really enforce anything there.

If you don't mind, I'll put this to a vote among OctoPrint's Patrons for a couple of days - I don't want to make this decision on my own.

Vanclec:
Hi Gina, thanks for your reply! :) 

The notifications can be moved a little bit down to avoid superposition with the navbar, yes. For the lining up, I honestly don't see the problem. I'm not a web integrator so maybe I don't know some basic rules, but since the user has a widescreen display that can show him more information at a glance, why not let him use it? The burger menu on the tab bar is great but since these plugins want to show quick information to the user, it can't apply there. 

Please have a look to my current navbar:

image
This one looks very disgracious to me.

I hope you're going better with the sickness. :) 

Regards,
Chris.

foosel:
Hey there,

one thing that I picked up over the years is that with design in general you want to stay consistent with regards to margins and alignment and such. Your proposal sadly doesn't do that. 

The obvious solution for that would be to redesign the app so that it overall makes better use of widescreen interfaces, but that is not something that is done overnight - it would need a serious rewrite of huge parts of OctoPrint's frontend and potentially also introduce compatibility issues with existing plugins that depend on the layout being as it is today. Backwards compatibility is something that I want to guarantee. 

I want to do a revamp of the current UI, as it is dated, has a lot of disadvantages (not scaling well to either larger or smaller screens being one of them) and simply at this point is one of the oldest parts in OctoPrint that hasn't aged well. But due to the aforementioned reasons that is not something I can now quickly do at the snap of a finger. It takes time to prepare and will probably involve a migration phase over a bunch of versions in order to allow plugin authors to adjust the plugins to whatever will be the next default UI of OctoPrint.

I also agree that a hamburger menu or something similar that arbitrarily hides "overflowing" plugins is not the right solution to this problem either, and I've still not come to a good idea to solve this problem either, but making the current layout inconsistent for everyone is not the solution and also something that my patrons so far seem to mostly agree on being the wrong approach as well.

So here's an idea to improve the current state until a better solution for the plugin overflow problem can be found and - longer term - a new UI is in place that makes proper use of available space:

  * I'll reduce the current menu entries in the navbar (Settings, System, Login when not logged in) to icons. That will free up some space. 
  * I'll also take a look at if something can be done about the big footprint of brand + instance name on the left side. 
  * I take care of the notification positioning. 
  * And we turn this change of your's into a *plugin*, listed on the official repository, that people can install on their instances if they want to. How does that sound? It's perfectly possible, the only question is whether you want to do that yourself or if you feel more comfortable having me or someone else do it (if you so far haven't looked into plugin development - in this specific case it would only consist of a very small handful of lines though injecting one line of JavaScript into the page which applies the change).

BTW, taking a look at your current navbar I see a prime example of a plugin with a very very huge footprint that could be optimized. The navbar plugin concept in OctoPrint was introduced to provide easy access to place *small* amounts of information or menu entries. In the specific case of the plugin up there, utilizing icons instead of text and maybe even getting rid of the target temperature (or only displaying that in a popout) plus allowing the user to maybe get rid of the Raspi temperature (which while nice to have you usually shouldn't need for operation) would already go a long way in reducing the footprint there.

Vanclec:
Hi Gina. :)

Thanks for the reply. I agree that such a revamp would take some time to implement. But your idea to free up some space on the current navbar sounds really good to me.

For the plugin development, I just made some changes to a plugin displaying data on the LCD screen but nothing to do with the interface. I'll first contact the temperatures' plugin maintainer to ask him to reduce the plugin's footprint, it'll free up some space as well. If he doesn't agree to, am I allowed to fork, edit and republish as an alternative? I'll also take a more serious look into plugins development to make one that could "materialize" my idea. For now I was making the modification with a userscript (Greasemonkey).

Thank you for the attention, by the way! That's very cool to see such a motivated and involved developer! :)

foosel:
So, I did the following on the `maintenance` branch (to be included in 1.3.2 and soon to also be made available on the `devel` branch):

  * switched "Settings" and "System" to icons only, with tool tip
  * removed "OctoPrint: " prefix when the instance has a name (but left it in the tooltip)
  * limited the title display to 250px max, limited the user name/login button to 100px max - if larger, text will get the ellipsis-treatment ("Some very long na...")
  * moved the notifications to not overlap with the navbar

Things now look like this:

image

I also quickly whipped up a very small proof of concept of how you'd go about turning the navbar resize into a plugin, you can find that [here](https://gist.github.com/foosel/098ea5aaa63b35001b37170b3363e790). It's nothing fancy (can't update itself, can't be installed through the plugin manager, really just the absolute bare minimum), but it works.

I think it's a good idea to address the footprint of the plugin in your screenshot with the author. And if the author doesn't want any changes you might be proposing there, I also don't see a problem with an alternative version on the plugin repository - and in general there's nothing keeping you from forking, adjusting and republishing open source software as long as you adhere to the license in question.

Anyway... if I didn't misunderstand you we seem to agree that providing the above steps together with a more general redesign down the road is a good compromise :) Any objections to closing this PR?

Vanclec:
No objections at all, thank you for the feedback! :)

PR #1748, opened 2017-02-02
Add sameDevice to api/slicing API

eyal0:
Adds sameDevice to the slicing API so that the web client can know if the configured slicer runs locally or not.

A web client could display if slicing is currently possible only if it knows whether slicing is configured, if we're are printing/paused, and if the slicer is on the same device as the printer.  Of those, only the last one is unavailable.  With this PR, all are available.

Reload octoprint and watch network traffic and watch the JSON output from REST API to api/slicing

This is the plugin that initiated the change: https://github.com/kennethjiang/OctoPrint-Slicer/pull/29

### DIFF
--- a/src/octoprint/server/api/slicing.py
+++ b/src/octoprint/server/api/slicing.py
@@ -72,6 +72,7 @@ def slicingListAll():
 			result[slicer] = dict(
 				key=slicer,
 				displayName=slicer_impl.get_slicer_properties()["name"],
+				sameDevice=slicer_impl.get_slicer_properties()["same_device"],
 				default=default_slicer == slicer,
 				configured=slicer_impl.is_slicer_configured(),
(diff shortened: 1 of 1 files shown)


eyal0:
Any further action needed from me?

foosel:
As the bot pointed out above, it's against the wrong branch. Additionally I'm down with a nasty cold and can't look into it at the moment (brain mushy, not going to touch code in this state).

foosel:
Ah, correction, it's no longer against the wrong branch :+1: Second problem (on my side) still applies though, sorry. Trying to get back on my feet as far as I can.

foosel:
Now, on second look that looks actually like such a simple change that I dared to merge it even in this state. Thank again! I'll also try to remember to backport it.

PR #1768, opened 2017-02-12
add sameDevice to SlicingViewModel

eyal0:

Adds sameDevice to the SlicingViewModel.  It was already added to the API in https://github.com/foosel/OctoPrint/pull/1748 but now it needs to be part of the view model.

I listed the SlicingViewModel.slicers() in Chrome's development console before and after this change and saw that sameDevice is now listed for each slicer.


foosel:
👍

Issue #1780, opened 2017-02-18
Video Stream gets added each time you navigate to the Control tab


ghulands:
#### What were you doing?

Everytime you navigate to the Control tab, it adds a new stream to the web page. This can be see in the web inspector and also in the bandwidth on the network interface - jumping about 2MB/s for each stream.

#### What did you expect to happen?
Only use the single stream.

#### What happened instead?
See above.

#### Branch & Commit or Version of OctoPrint
1.3.1

#### Printer model & used firmware incl. version
N/A

#### Browser and Version of Browser, Operating System running Browser
Safari 10.0.3, macOS Sierra 10.12.3

#### Link to octoprint.log

N/A

#### Link to contents of terminal tab or serial.log

N/A

#### Link to contents of Javascript console in the browser

N/A

#### Screenshot(s) showing the problem:

N/A

I have read the FAQ.

ntoff:
works normally in Chrome on windows. Navigating to the control page opens up the stream, navigating away from the control tab closes the stream after a few seconds. Navigating to it, away from it, to it, away from it in quick succession doesn't open up any new streams.

OssenFoss:
I have noticed the same thing. I tried Chrome (Version 56.0.2924.87 (64-bit)) and Firefox (51.0.1 (64-bit), both work fine. It only affects Safari (Version 10.0.3 (11602.4.8.0.1))

ghulands:
Just another thing to add...

It will even duplicate the stream if you leave the control tab selected, but navigate away and back with the safari tabs.

foosel:
Looks like a Webkit bug. [First report I found](https://lists.apple.com/archives/web-dev/2006/Oct/msg00138.html) is from 2006. And [here's another ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=73395) about the same bug in Chromium, since fixed. Apparently no fix for Safari.

[...]

PR #1786, opened 2017-02-20
Issue #1780: Safari Webcam Fix.

ghulands:
What does this PR do and why is it necessary?
Disables the disabling/enabling of the webcam if running in Safari as it doesn't disassociate the img src of the stream and thus multiples bandwidth unnecessarily.

How was it tested? How can it be tested by the reviewer?
In Safari.

Any background context you want to provide?

What are the relevant tickets if any?
#1780
Screenshots (if appropriate)

Further notes

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -362,9 +362,21 @@ $(function() {
 
         self.onSettingsBeforeSave = self.updateRotatorWidth;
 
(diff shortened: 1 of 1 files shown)


foosel:
Thank you so much for this :+1:! That really made my day - it doesn't happen that often that a user reporting a problem sends in a PR to fix it themselves :D

I'll also backport this to the `maintenance` branch so that it makes it into the 1.3.2 release.

PR #1803, opened 2017-03-06
Timelapse deletion confirmation dialog

ntoff:
Fulfils a feature request: https://github.com/foosel/OctoPrint/issues/1798 to add a confirmation dialog when deleting a timelapse to prevent accidental deletion.

Create a timelapse video via the usual methods, or an empty text file named "test.mpg" and hit the delete icon in the timelapse tab, observe there is a dialog confirming the deletion and that pressing "yes" deletes the file while pressing "no" does not.

I've seen this request a couple of times and been a victim of accidental deletion myself, figured I'd have at it.

https://github.com/foosel/OctoPrint/issues/1798

image

#### Further notes
Should it be optional like the print cancel confirmation?


ntoff:
eegads the commit list

Issue #1798, opened 2017-03-04
[Request] Confirm delete of Timelapse video


scorpydude:
For the love of all that is right and just, can we please have a confirmation of delete of a timelapse? I've lost many a great timelapse when trying to view to a dodgy mouse and few too many beers and quick clicks! (Or a recycle bin or something)

foosel:
Just realized this is a duplicate of #748. Also see discussion in #1807 though, we are nearing a solution.

PR #1804, opened 2017-03-06
Add confirm dialog to timelapse deletion

ntoff:
**What does this PR do and why is it necessary?**

Fulfils a feature request: #1798 to add a confirmation dialog when deleting a timelapse to prevent accidental deletion.

**How was it tested? How can it be tested by the reviewer?**

Create a timelapse video via the usual methods, or an empty text file named "test.mpg" and hit the delete icon in the timelapse tab, observe there is a dialog confirming the deletion and that pressing "yes" deletes the file while pressing "no" does not.

**Any background context you want to provide?**

I've seen this request a couple of times and been a victim of accidental deletion myself, figured I'd have at it.

**What are the relevant tickets if any?**

#1798

**Screenshots (if appropriate)**

image

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/timelapse.js
+++ b/src/octoprint/static/js/app/viewmodels/timelapse.js
@@ -207,7 +207,7 @@ $(function() {
  					.done(self.requestData);
             } else {
                 showConfirmationDialog({
-                    message: gettext("This will cancel your print."),
+                    message: gettext("Delete selected timelapse?"),
                     cancel: gettext("No"),
                     proceed: gettext("Yes"),
                     onproceed: function() {
--- a/src/octoprint/static/js/app/viewmodels/timelapse.js
+++ b/src/octoprint/static/js/app/viewmodels/timelapse.js
@@ -18,6 +18,7 @@ $(function() {
         self.timelapseFps = ko.observable(self.defaultFps);
         self.timelapseRetractionZHop = ko.observable(self.defaultRetractionZHop);
         self.timelapseCapturePostRoll = ko.observable(self.defaultCapturePostRoll);
+        self.timelapseDeleteConfirm = ko.observable(true);
 
         self.persist = ko.observable(false);
         self.isDirty = ko.observable(false);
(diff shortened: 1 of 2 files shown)


foosel:
Considering the reaction the [last time we added a confirmation dialog where previously there was none](https://github.com/foosel/OctoPrint/issues/1638) I think this needs to be controlled by a setting. Especially since otherwise deleting a larger number of timelapses through the UI will become quite the hassle indeed.

I'm actually thinking about if this shouldn't be a browser specific setting (written to local storage) directly toggleable on the list instead of something in the Settings dialog... e.g. (mock-up):

image

Thoughts? :)

ntoff:
I wondered about a setting but then also wondered if mass deletion shouldn't be handled better all together with checkboxes. For me I made a script in the system menu section that just "rm *.mpg" for mass deletion.

Still though, cancelling a print can be a more urgent matter given there's no always visible emergency stop button so a confirmation dialog might get in the way of a speedy print halt, while filesystem maintenance is a bit less of an urgent matter and mass deletion of things needs a "checkbox to select" kind of deal.

Unfortunately I don't know how to do the mass select checkbox thing, otherwise I'd have put that in there too with this PR.

>I'm actually thinking about if this shouldn't be a browser specific setting (written to local storage) directly toggleable on the list instead of something in the Settings dialog... e.g. (mock-up):

That's a good idea, with it being on by default maybe?

ntoff:
Updated with temporary confirmation disable boxy thing. It doesn't persist across a page refresh (on purpose) so it automatically re-enables the confirmation on each page visit.

gege2b:
I agree with @ntoff about checkboxes for multiple deletion of timelapse (even if I didn't do timelapse myself)
Something similar to what webmails does (thinking about gmail specifically)

And maybe this could be extended to gcode files ? :D 

Agreed too about reset the confirmation checkbox on page refresh

ntoff:
I'm reconsidering the temporary nature of the setting, people might think something is broken when the setting doesn't stick.

Also wondering if it's worth having maybe another category or at least a group within "Features" just for safety checks? Maybe have a setting for the confirmation dialog for deleting, shutting down, cancelling a print, etc, and have all those settings grouped together? Is that going too far?

gege2b:
I think it would add much overhead on the settings dialog.
on the other hand, it will give users a true control over when to be "safe" and when not

Personnaly I'm not against this kind of options (I love options, options are life)

ntoff:
Ok, I think for now I like @foosel 's idea of the tick box for enabling it or disabling it but maybe not saving it through local browser storage but having the "save defaults" option save whether or not the tick box stays ticked when it saves all the other options listed on that page, otherwise people might think the save defaults option is broken or something (in a similar way my "it's only temporary" idea might cause confusion).

foosel:
@ntoff why you close? O_O

ntoff:
too many random commits like messing up the dialog text, I want to clean that up and I'm currently trying to do a crash course in copy/paste school of learning to try to put select check boxes for mass deletion (which I admit I'll probably never figure out).

PR #1860, opened 2017-04-11
added missing period and space so works

ParkerK:
Just fixing documentation so it's accurate. Won't run as is, needs the . and space as shown in requirements.txt


foosel:
Thank you for spotting this. I've manually merged the PR on the `maintenance` branch (for reasons, see above).

PR #1867, opened 2017-04-13
Fix dragndrop upload

eyal0:
    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

Fixes https://github.com/foosel/OctoPrint/issues/202 so that the drag and drop upload doesn't disappear when the mouse stops moving.

I was able to reproduce the original issue and also see that this solution solves the issue.  **The reviewer should test that this doesn't break drag and drop on platforms that were already working.**

Fixes https://github.com/foosel/OctoPrint/issues/202

#### Further notes
Another easier solution is to simply change the timeout from 100 to 400 as described in the issue.  it's a smaller fix but a little clunky in that the page becomes less responsive.

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/files.js
+++ b/src/octoprint/static/js/app/viewmodels/files.js
@@ -810,6 +810,7 @@ $(function() {
                 self.listHelper.removeFilter('sd');
             }
             self.sdTarget = $("#drop_sd");
+            $("#drop_overlay").on('drop', self._forceEndDragNDrop);
 
             function evaluateDropzones() {
                 var enableLocal = self.loginState.isUser();
@@ -1070,7 +1071,6 @@ $(function() {
             var dropZoneLocalBackground = $("#drop_locally_background");
             var dropZoneSdBackground = $("#drop_sd_background");
             dropOverlay.addClass('in');
-            dropOverlay.on('drop', self._forceEndDragNDrop);
 
             var foundLocal = false;
             var foundSd = false;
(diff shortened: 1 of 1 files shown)


foosel:
I might be overlooking something here, but with the changes I'm getting this behaviour in Chrome 57.0.2987.133 and Firefox 52.0.2 under Windows 10:

pr_1867

It appears that the drop is never detected properly and hence the overlay is not removed. It works fine when just moving the dragged file out of the window, but actually dropping it blocks the UI for good. I assume you didn't see that in your own tests and we have a case of different behaviour across operating systems here?

eyal0:
Oy, I completely forgot to test that!  I only tested dragging away from the screen and not actually uploading!  Sorry.

I tested uploading, both to local and to a non-exsistant SD card.  In both cases, the overlay goes away.

Thanks for testing!

eyal0:
I fixed it, by the way.  Try again?

foosel:
Thanks, that looks better :) I'll squash this and cherry pick it to the `maintenance` branch - since it's a bug fix it makes sense to push it out as fast as possible and that will ensure it will be in 1.3.3. However that will mean that Github won't detect this PR as merged. I hope that's ok for you nevertheless.

foosel:
This is now merged on `maintenance` as 358a252, will be in the 1.3.3 release (and soon also merged up to `devel`). Thanks! 👍 Closing this manually since Github doesn't handle cherry-pick style merges sadly, don't be alarmed by that please.

PR #1868, opened 2017-04-13
allow multiple upload

eyal0:
highly appreciated!

checklist:

    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

Add the ability to upload multiple files.

I tried uploading multiple files and it worked.

#### Further notes
This might be a big change.  I'm not sure what tests I should be running.  But it basically works.  Multiple files were uploaded.  I didn't test drag n drop for multiple files (not sure how).


foosel:
I like the idea, the current implementation makes the upload progress bar hiccup hard though... First file reports progress just fine, second one however only reports once it's fully uploaded, until then it sits at `. I tried sequential uploading on the upload widget (by setting [`sequentialUploads`](https://github.com/blueimp/jQuery-File-Upload/wiki/Options#sequentialuploads) to `false`) but that didn't improve behaviour either. 

If this is to be added, that needs to be solved and I have to admit, I'm a bit unsure as to how right now :/

eyal0:
I tried to reproduce what you're seeing but it looks okay to me.  Maybe because the saving was very quick for me?  Was it repeatable for you?

foosel:
Yesterday I had this very much repeatable. Today it doesn't reproduce O_O Exact same code.

What I do still see - logically, since that part of the code is simply not suited for multiple uploads as of now - is the highlighting in the refreshed list only highlighting one of the uploaded files, and also triggering an expensive refresh of the file list on all individual files uploaded (of which only the first one actually does get send out though thanks to [this line](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/js/app/viewmodels/files.js#L230)). In short, `_handleUploadDone` needs to be able to handle multiple uploads. Maybe also `_handleUploadFail` (both mess around with the progress bar currently, they should only touch that when ALL uploads are done/have failed). The `focus` parameter to `requestData` also needs to be adjusted to handle focusing multiple files. And I'm very unsure how to solve the ["scroll to uploaded file" feature](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/js/app/viewmodels/files.js#L285-L289) for multi file uploads - only scroll to the first one in the list maybe? 

Here's a screenshot of the requests that currently get generated due to one triple file upload:

image

1, 2 and 3 are the actual file uploads. 4 is the file refresh triggered by 1 finishing uploading, 5 is triggered by an "files changed" event that I think is triggered by either 2 or 3 (would normally be suppressed, the [suppression](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/js/app/viewmodels/files.js#L988) [mechanism](https://github.com/foosel/OctoPrint/blob/master/src/octoprint/static/js/app/viewmodels/files.js#L1031) here fails), 6, 7 and 8 are triggered due to metadata analysis finishing on 1, 2 and 3. File listings are expensive. We'd like to only see 1, 2, 3, 4, 6, 7, 8 here - 6, 7  and 8 can't be avoided because metadata analysis can take a long time, so we don't want to bundle those, but we definitely only want exactly one file refresh after all uploads complete.

I think the issue is that we handle this like individual file uploads due to `_handleUpload*` triggering per file. What would be needed here is a way to trigger handlers for both individual uploads (that handle error messages & tracking which files need to be highlighted) as well as the overall upload process (that also suppresses unnecessary updates & updates the progress bar & finally refreshes the list). I think that could be achieved through the [start](https://github.com/blueimp/jQuery-File-Upload/wiki/Options#start) and [stop](https://github.com/blueimp/jQuery-File-Upload/wiki/Options#stop) callbacks on the fileupload library, but I might be mistaken.

eyal0:
I understand what you're saying, but I'm not sure what the right thing to do is.

Imagine that I have a very slow network and I'm uploading a few files.  I might want to get the "files changed" update after every file because I can start slicing and printing that file even before the rest of the files complete their upload.  Why should I wait for other files to complete?  With this argument, maybe files changed after each file is a good idea?

As a user, I would expect to get a warning about each one that fails.  And I'd expect the progress bar to hit 100% after they're all done, success or not.

Let's think about what the right behavior should be and then see what needs to be changed to do it.  As I see it, the only problem is too many files changed events.  Three ideas:

1. Only download files changed at the end.  Least traffic but maybe users don't need to wait.
2. Files changed per file.  Very responsive but too much network traffic.
3. Coalescing events.  Only download the file list once per X seconds at maximum.

Other than too many file list downloads, any other issues?

foosel:
The problem with the file list requests is actually not the traffic caused by them, but the caused CPU load. Each uncached file listing necessitates a crawl through he whole hierarchy in the uploads folder. And that kind of crawling can spike the CPU on the Pi easily, depending on how many files the user has there. So what I fear here is a user uploading a bunch of files at once and then bringing their ongoing print to a crawl due to that (uploading is not disabled while printing, and doing that would impede existing workflows). That can happen now as well if you have a huge number of files in your uploads folder, but multiple requests handled sequentially will certainly increase the chance of it happening. Starving the print thread for 1s is less bad than starving it for 3s.

There are currently three reasons for refreshing the file list:

  1. an upload request finishing - we know something changed, so let's fetch changes and highlight/scroll to the file that we just uploaded
  2. a "files changed" event from the backend - we didn't know something changed, but the server just told us, so let's fetch changes and leave everything else as is
  3. a "metadata changed" event from the backend - like (2)

I've tried to visualize that in a sequence diagram:

```
+-----------+                 +-----+               +---------+           +-----------+
| Frontend  |                 | API |               | Storage |           | Analysis  |
+-----------+                 +-----+               +---------+           +-----------+
      |                          |                       |                      |
      | upload [file]            |                       |                      |
      |------------------------->|                       |                      |
      |                          |                       |                      |
      |                          | store [file]          |                      |
      |                          |---------------------->|                      |
      |                          |                       |                      |
      |                         signal file list changed |                      |
      |<-------------------------------------------------|                      |
      |  (2)                     |                       |                      |
      |                          |                [path] |                      |
      |                          |<----------------------|                      |
      |                          |                       |                      |
      |                          | start analysis of [path]                     |
      |                          |--------------------------------------------->|
      |                          |                       |                      |
      |    upload of [path] done |                       |                      |
      |<-------------------------|                       |                      |
      |  (1)                     |                       |                      |
      |                          |                       |                      |
      | get file list            |                       |                      |
      |------------------------->|                       |                      |
      |                          |                       |                      |
      |                          | create file list      |                      |
      |                          |---------------------->|                      |
      |                          |                       |                      |
      |                          |                       | recursive scan       |
      |                          |                       |---------------       |
      |                          |                       |              |       |
      |                          |                       |<--------------       |
      |                          |                       |                      |
      |                          |             file list |                      |
      |                          |<----------------------|                      |
      |                          |                       |                      |
      |                file list |                       |                      |
      |<-------------------------|                       |                      |
      |                          |                       |                      |
      | highlight [path]         |                       |                      |
      |-----------------         |                       |                      |
      |                |         |                       |                      |
      |<----------------         |                       |                      |
      |                          |                       |                      |
      |                          |                       |                      |
      |                          |              signal analysis ready of [path] |
      |<------------------------------------------------------------------------|
      |  (3)                     |                       |                      |
      |                          |                       |                      |
      | get file list            |                       |                      |
      |------------------------->|                       |                      |
      |                          |                       |                      |
      |                          | create file list      |                      |
      |                          |---------------------->|                      |
      |                          |                       |                      |
      |                          |                       | recursive scan       |
      |                          |                       |---------------       |
      |                          |                       |              |       |
      |                          |                       |<--------------       |
      |                          |                       |                      |
      |                          |             file list |                      |
      |                          |<----------------------|                      |
      |                          |                       |                      |
      |                file list |                       |                      |
      |<-------------------------|                       |                      |
      |                          |                       |                      |
```

Since the "files changed" event fires within the server-side upload handler, every upload would trigger both case (1) and (2). To prevent this, upload request handling currently sets a flag on start of the upload of a file that disables (2) - we can't just leave (1) away because we need the information from the server what file we need to actually highlight (name got sanitized) and we only get that information after (2) got triggered, at the end of (1). It also clears that flag on `done` or `fail` of the upload. This mechanism doesn't work with multiple uploads because one uploaded file might clear a flag set by another upload, causing the check for the flag in (2) to not trigger. Hence, unnecessary file list refresh that gets overruled just seconds later thanks to (1). That is a problem that needs to be solved regardless of whether we talk about refreshing the file list at the end of (1) for every single file in an upload or only once all files are uploaded and finished processing.

The other problem is that every single file upload individually triggers a file refresh as well which as mentioned above can cause issues with potentially ongoing prints. That indeed might best be solved by introducing something like a timeout ("refresh 5s after `done`"), so basically what you propose in your third option if I understand you correctly.

> As a user, I would expect to get a warning about each one that fails. And I'd expect the progress bar to hit 100% after they're all done, success or not.

Agreed. That's yet another problem. Right now the progress bar might go all over the place - depending on request ordering and race conditions - because it gets manipulated from within the `done` and `fail` handlers attached to the individual uploads instead of the overall upload task. The warning per failed upload should already work however, but if e.g. the second upload of three fails, you'll get a progress bar first showing maybe 33%, then jumping to 0%, then back to maybe 66%, then 100%. Likewise if the file list request triggered by a successful upload of one file gets completed before the other uploads complete. The progress bar handling must be decoupled from the individual upload handling.

eyal0:
I have a new argument to convince you why to accept this PR:

OctoPrint already supports multiple file upload.

If I select multiple files and click-drag them from my desktop on to OctoPrint, they will all upload.  So if OctoPrint is already doing it like that, might as well also allow mutiple upload from the upload button.

Convincing?  :-)

foosel:
Just to clarify, I don't have a problem at all with accepting this PR/the functionality it adds in general, I have a problem with accepting it in its current state due to the issues stated above. 

If OctoPrint is already allowing you to do that via drag-n-drop that doesn't necessarily mean it's supported functionality, and it in fact will have the same problems, which actually would mean I'd currently clarify it as a low prio bug. By accepting this PR I'd make this a supported feature. I'd rather have it thought through in that case and properly implemented, without any progress bar hiccups thanks to race conditions or unnecessary and preventable requests to the server.

I'll happily merge this once these two issues are solved (either by you or - which will take longer due to my other tasks - by me). But please understand that I do not want to "quickly add something" when I already know that in its current form it will cause issues and hence additional maintenance overhead. There are already enough places in OctoPrint that need improvement, let's not add yet another one when simply adjusting it to take care of the identified problems can be done fairly easy.

foosel:

eyal0:
That's okay, it's complicated to get the multiple-file-upload to look good in the GUI.

Issue #1842, opened 2017-03-27
[Brainstorming] FontAwesome in OctoPrint is really outdated


Depaulicious:
As a plugin developer I sometimes struggle to add the right icon for each feature. Very often all icons I need are in latest FontAwesome, but not in the one included in OctoPrint.

I'm opening this issue to learn if you guys are okay with an eventual pull request in which I'll try to update it , trying not to break everything else (e.g. making sure the icons that are currently provided are also available as `icon-*` in addition to `fa-*`, or other similar solutions).

I'm not willing to work on a fork that won't be accepted upstream; plugin compatibility comes first IMO.

foosel:
If it's possible to stay backwards compatible (icon-* works, existing plugins don't break), that would be **great**. The backwards compatibility situation is actually the reason why this stuff is so outdated (well, that and the fact that I didn't think about doing an update before rolling out plugin support with 1.2.0). The last time I looked it sounded like being quite a huge ordeal though and hence I haven't yet found the time. Any help there would be very welcome indeed!

I'd target that for 1.4.0 rather than 1.3.x though - that way we'd get more time to iron out any compatibility kinks before it hits mainstream, and allow to collect feedback from people running the  `devel` branch.

Depaulicious:
Awesome! I'll try to do it as soon as I have time, then.
I think, however, that the icons changed their look a little bit with releases. I'll see how they look like.

foosel:
Definitely keep me updated! If there are some changes, that's a death we'll have to die I guess, and it will not cause as much interruption as a completely missing icon ;)

ntoff:
wait, if I'm using fa-* now, and it works, what's up with that? Does the current octoprint version of font awesome also have fa- fonts in it? Or have I somehow got a client side thing going on meaning only I can see the icons I'm using?

foosel:
@ntoff might be a plugin that concurrently installed a newer version of font awesome.

Come to think of it, that might actually be a good solution too (if it doesn't cause conflicts), marking old icons as deprecated but keeping them available for a couple more versions to give plugin authors time to migrate.

ntoff:
ruh roh, there's uh... some place I need to be, I think I see my friend over there

*runs*

foosel:
Well, at least that mystery is solved, the TouchUI plugin bundles a newer Font Awesome (4.4.0).

[...]

PR #1915, opened 2017-05-13
Update FontAwesome to 4.7

Depaulicious:
Okay, so as I promised in #1842 I updated FontAwesome in OctoPrint.
This PR includes two commits: one that updates the fonts and the main CSS (with some patches to make it continue to work with old `icon-` classes) and one that updates the templates in order to use the new `fa-` classes.

Note that `octoprint.css` looks rewritten because I regenerated `octoprint.less`, I had to add a class.


foosel:
Currently fighting with 1.3.3 / some fixes, will take a look and merge (or even backport to `maintenance`) once I've won ;)

foosel:
@Depaulicious Instead of completely replacing the old Font Awesome with the new version, could you change this so that the old version is still available and the new one added concurrently? I know you patched the `css` to allow the `icon-` prefixes to still work, but I'm worried that we nevertheless might introduce regressions that way, plus it will make updating to newer versions of "NewFontAwesome" more tricky since we'd need to manually merge with your compatibility layer everytime. And as discussed in #1842 it might in general be less disruptive to do a soft-rollout.

BillyBlaze:
Is it perhaps a good idea if we would put this in a jinja2 helper?

e.g. something like:
```{{ 'circle' | icon }}``` that would generate into ```<i class="fa fa-circle"></i>```

Depaulicious:
@foosel I think so, as long as CSS allows to rename fonts so I can use the new one with `fa-` and the old one with `icon-`. Is it okay for you if I `rebase -i` the branch and delete the CSS commit, then add the new one without stacking up changes?

@BillyBlaze I don't think so. While many icons are just `fa fa-something`, some include other classes to implement different functionality (stack icons, spinning icons, etc). So that would lead to ugly workarounds IMO.

foosel:
@Depaulicious you could use sub folders for the two versions to keep css and font files disjunct, the css already uses relative paths if I'm not mistaken? I'd be fine with moving the font-awesome stuff into its own folders (e.g. `assets/vendor/fontawesome3.2.1` and `..../fontawesome4`) if that would help things along.

Rebasing is totally fine!

@BillyBlaze What @Depaulicious said, I fear that while that would help in a lot of "normal" cases, there are a lot of locations even within stock OctoPrint where I used some spinning and such that would then not be captured (or the custom filter would become too complicated). I mean, yes, it would allow easier switching to newer versions if future FA installments change the syntax yet again, but I fear the cost for that would be too high to make it really a 100% solution.

Depaulicious:
Alright. I'll rebase then and try to do that. I'll also rebase on top of the new `devel` changes.

foosel:
I'll probably try to get it into 1.3.4 (1.4.0/`devel` is still far off), so if you go against `maintenance` instead of `devel`, that might be even better.

Depaulicious:
Okay, I'll try. Hopefully it won't conflict against _everything_. I think I have to resubmit the PR for github to allow merging, though, unless you're going to merge it manually.

foosel:
Just saw that you changed the base branch, so apparently a re-submit isn't necessary (anymore)? Conflicts should hopefully be minimal, there haven't been that many changes to the UI stuff so far on `devel` as far as I can remember.

Depaulicious:
Yes I changed it. I just finished cherry-picking my commit on top of `maintenance` and I'm currently adding the two FontAwesome versions.

Depaulicious:
I'm kind of worried that keeping the directory structure for css won't be a good thing when OctoPrint squashes all the stylesheets together. I think I'll keep it for the fonts only.

foosel:
Well, it could be done with a CSS rewrite routine in the webassets setup, but yes, for now let's not make it overly convoluted :D

Depaulicious:
Okay this should do it.

foosel:
:+1:

PR #1981, opened 2017-06-23
Update safemode.rst

TylonHH:
I guess writing in the yaml is fix, isn't it?


foosel:
Actually no, the yaml setting gets reset after starting once in safe mode. However, it should probably rather say "the first and the last" (I think I moved things around a bit), starting with `--safe` will always start in safe mode, you'll need to strip that parameter away in order to start in normal mode.

PR #2022, opened 2017-07-21
Add "stl_transformation_matrix" to slicing profile

markwal:
CuraEngine has functionality to allow a 3x3 matrix transform to be applied to input stl's to achieve scaling and rotation.  It may be helpful to be able to set an intended transform in either the slicing profile or in the overrides. For example, if you're slicing for an interesting printer kinematics or to accommodate a simple UI to allow scale or rotation, but perform the operation as part of slicing rather than re-uploading a transformed stl.

Currently the matrix transform is only available via the CuraEngine "-m" command line setting and not as a "-s" setting.  In addition, CuraEngine changed it in the new cura to be expressed as an array of arrays instead of a flat list of numbers like 15.04.  So this PR proposes an abstraction that stores the matrix as a list of list in the profile and flattens it for the old cura command line.

### DIFF
--- a/src/octoprint/plugins/cura/__init__.py
+++ b/src/octoprint/plugins/cura/__init__.py
@@ -286,7 +286,7 @@ class CuraPlugin(octoprint.plugin.SlicerPlugin,
 				# The extruder count is needed to decide which start/end gcode will be used from the Cura profile.
 				# Stock Cura implementation counts the number of objects in the scene for this (and also takes a look
 				# at the support usage, like the engine conversion here does). We only ever have one object.
-				engine_settings = self._convert_to_engine(profile_path, printer_profile,
+				engine_settings, matrix = self._convert_to_engine(profile_path, printer_profile,
 				                                          pos_x=pos_x, pos_y=pos_y,
 				                                          used_extruders=1)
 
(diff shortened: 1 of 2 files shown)


foosel:
Sorry for not getting around to do a review yet. 

I have a more general question first... I'm not sure if storing that matrix in the printer profile is the best way to go. Wouldn't it be better - like you also hinted at - to extend the slicing API (both REST and internal) to allow providing it as a parameter, extend the slicer engine properties to report on "supports scaling", "supports rotating" and then use all that to at least add six input fields (rotate around X, Y, Z, scaling factor in X, Y, Z) to the slicing dialog if the slicer supports that stuff?

It feels like model transformation is something that really is usually too model specific to be stored in the printer profile. But I might be overlooking something here ;)

markwal:
I understand what you mean.  I also thought of plumbing it that way, more like posx and posy.  However, one thing that leaned me over to printer profile is the use for something like the PrintrBot PrinterBelt since it is more like part of slicing to tilt everything to match the tilted conveyor.

So it is something that attaches to both (either?) the model and the profile so I thought a property would be nice because then you could use it either way.  Either in the profile or as an override or both.

dnewman-polar3d:
FWIW, note that the next pre-production revision of the printrbelt will have an (manually) adjustable angle.  In practice, I don't see printer owners changing the angle often owing to the need to fine tune the angle on the slicing-software end.  (Having the angle off may impact a physically long print.)  Other than people just experimenting, I suspect most operators will either leave the angle alone or set it very infrequently.

foosel:
@markwal Sorry again for the delay (flu of doom 😕). I just took another look at your suggested changes and now I'm even more confused :) In our conversation up there we talked about the up and down sides of having this in the printer profile, but reading your code and your PR text I now realize I must have been a bit blind since you are explicitly talking about the *slicing* profile.

FWIW, for something like belt based printers I fully agree, a transformation matrix indeed makes sense. That one I would actually see as part of the printer profile though, not the slicing profile, considering that it should be quite static.

How about actually moving that as a new field to the printer profile, then using that in the cura plugin?

foosel:
@markwal ping? :)

markwal:
I like your idea of having it be in the printer profile and having the cura plugin look it up.  I'll take a run at doing it that way.

PR #2041, opened 2017-07-27
Singlenozzle Honend Temp.

veneficusoz:
First tanks for the great Software... Vielen Dank ;)

It would be great if the Temp. is always tracked from the T0 Tool. At the moment the Graph sucks on the last value when changing the Tool.


foosel:
If you want to report a bug, please open [a bug report](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md#how-to-file-a-bug-report), not a pull request ;)

PR #2139, opened 2017-10-03
Add shutdown warning about SD printing and serial reset

ntoff:

As per request: https://github.com/foosel/OctoPrint/issues/2131
Add an extra warning text to the shutdown dialog about potentially disrupting
any ongoing prints, including those running off an SD card attached to
the printer, in case certain control boards are sensitive to random
serial activity causing them to be reset.

Also added a bold tag to the action being performed (shutdown vs restart) to draw a users eye to that section to indicate this text may be worth reading.

Access the system shutdown / restart menu, additionally activate or de-activate a plugin that requires a restart, observe the shutdown dialog now contains the new warning.

As per request: https://github.com/foosel/OctoPrint/issues/2131

image

#### Further notes

I think it's worth mentioning *why* certain boards happen to reset on serial activity, but perhaps not in this dialog box. Maybe that's best left to the FAQ / wiki page?


ntoff:
re-doing it against maintenance (branch nuked because I thought I broke it so it was easier to just start over)

PR #2152, opened 2017-10-13
Add print interruption warning to auto update dialog

ntoff:

Adds a warning to the auto updater dialog about possibly resetting your printer. This typically only applies when using octoprint to *monitor* a printer that is printing off an SD card, which was initiated via the LCD panel in which case OctoPrint has no way of knowing the printer is busy and therefore won't automatically block the installation of updates.

Install an old version of a  plugin or octoprint, allow it to notify you of an update, observe the new dialog box warns about a possible reset 

Fills in a missing piece of a feature request that was missed on the first PR

https://github.com/foosel/OctoPrint/issues/2131

image

### DIFF
--- a/src/octoprint/plugins/softwareupdate/templates/softwareupdate.jinja2
+++ b/src/octoprint/plugins/softwareupdate/templates/softwareupdate.jinja2
@@ -19,6 +19,12 @@
             {{ _('Be sure to read through any linked release notes, especially those for OctoPrint since they might contain important information you need to know <strong>before</strong> upgrading.') }}
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


foosel:
Yay! 👍

ntoff:
note/question

Should the h5 be changed to a bold? or just removed entirely?

It's been driving me insane ever since the discussion. I thought about the big scary warning being the next "section", first section being release notes, but since the first "section" doesn't have a bold heading, should I just remove the bold/heading all together and make it normal text? Or do you want to just take care of it on your end and I should stop obsessing over it?

foosel:
I'll take care of it, no need to obsess ;)

PR #2162, opened 2017-10-20
Update gettingstarted.rst

jcbelanger:
This a quick pull request to fix the issue described in https://github.com/foosel/OctoPrint/issues/2160


foosel:
Thanks! I'm still going to up the version of requests though - looks like a simple update of requests is safe based on their changelog, but the update to the docs still definitely makes sense!

PR #2163, opened 2017-10-21
Use valid css instead of less for example

jcbelanger:
Just changing the less to css in the css file used in the example


foosel:
Oops, that must have been a left-over. Thanks!

PR #2176, opened 2017-10-25
Fix sphinx warnings during OctoPrint doc generation

pbackx:
There were a number of warnings while generating the OctoPrint documentation (about 20). Most warnings caused an incorrect layout, but some hid information.

"make clean html" now runs without warnings. All information is now correctly displayed.

I know this is not a large problem, but it bothered me when I was writing the virtual printer documentation. With the broken windows theory in mind, I thought I'd take a bit of time and fix it.

none

#### Further notes
There is one issue that I'm not sure I completely fixed to my satisfaction. In some instances, JSON responses were abbreviated with an ellipsis. The HTTP and JSON lexer did not like that. So I have removed the ellipsis and kept the important parts. 
The only other option is to use no lexer ("none") and no nice highlighting. If you prefer that, just let me know and I can change it.


docs/jsclientlib/index.rst
```diff
@@ -14,7 +14,7 @@ If you are using it from a web page hosted on OctoPrint as a Jinja2 template, yo
 methods to embed it instead of manually entering the URL, in order to have OctoPrint take care of setting the
 correct URL prefix:
 
-.. code-block:: html
+.. code-block:: html+jinja
```
foosel: 😱 I had no idea this was possible! Thanks for showing me!

docs/bundledplugins/pluginmanager.rst
```diff
@@ -94,15 +94,15 @@ octoprint.plugin.pluginmanager.reconnect_hooks
 
 .. py:function:: reconnect_hooks_hook(*args, **kwargs)
 
-   Returns additional hooks defined by the plugin for which the plugin manager
-   should display the "You should reconnect to your printer" message on plugin
-   install/uninstall/enabling/disabling.
+Returns additional hooks defined by the plugin for which the plugin manager
```
foosel: Doesn't this de-indentation change how that is rendered? 🤔

foosel:
Thank you so much for this! ❤️ 

I've been meaning to take a look at this for ages now because it's been bothering me as well but never found the time. 

About the ellipsis, I think just leaving them out is fine. It's not optimal, but not having any syntax highlighting is the worse choice I think. It might be possible to write a custom lexer to work around this, but that would probably be overkill.

I think it would make sense to do this PR against `maintenance` instead of `devel` - it certainly is more of an improvement instead of a new feature, so nothing speaks against an inclusion in the next maintenance release 1.3.6 :) Either you rebase your commit against `maintenance` (just changing the target branch on Github will most likely not work as hoped, last time I tried that I ended up with a PR with all me `devel` commits), or I cherry pick your commit on `maintenance` (that won't register as "merged PR" though). What would work better for you here? :)

docs/bundledplugins/pluginmanager.rst
pbackx: Actually it doesn't. I tested this to be sure. With the indentation, the Python code-block below was not rendering correctly. I don't know why, maybe because there was double indentation and then there was more indentation in the Python code. I decided to remove the indentation for the entire section and that worked just fine.

docs/jsclientlib/index.rst
pbackx: It took me some searching through the huge Pygments lexers page, but yes, those things exist :)
http://pygments.org/docs/lexers/

pbackx:
I guess a cherry pick is the easy thing to do, so :+1: for cherry picking

foosel:
Cherry picked to `maintenance` as a0df4e78b57d61df9748e4580f6222509dced476. I also found the solution for [this comment of mine](https://github.com/foosel/OctoPrint/pull/2176#pullrequestreview-72113851). Issue was the the ``code-block`` was improperly indented, and you deindentation of the whole `py:func` body fixed that wrong indent. 2fcbf49348431c03fbe251cef0e4444719d135cf moved things back at consistent indentation levels and from the looks of the [rendered page](http://docs.octoprint.org/en/maintenance/bundledplugins/pluginmanager.html#reconnect_hooks_hook) the example still works fine. While in the general vicinity anyhow I also forced a higher version of the `sphinxcontrib-mermaid` dependency since 0.2.1 was giving me a deprecation warning in Sphinx (see 5f34d5df27d827bafaf8b47484956022f4e51213).

Thanks again for the PR 👍, compiling the docs locally now really is more fun than before 😊

Closing this manually since Github doesn't detect cherry-picks - don't be alarmed by that please :)

Issue #1110, opened 2015-11-06
[Request] Create a more granular permission system


foosel:
At the moment OctoPrint only distinguishes between three user roles if ACLs are enabled:
- **Guests/Anonymous**: May view the current status, including progress, file being printed, gcode preview, webcam, terminal output. May also download timelapses and uploaded files.
- **Users**: All of the above, plus may control the printer, upload files, start/pause/cancel jobs, send commands to the printer.
- **Admin**: All of the above, plus can view and manage settiings, update the installation, install plugins.

As various tickets (e.g. #950, #1109 and #640 to name a view) and posts on mailinglist etc have shown, that is not granular enough for many usage scenarios.

Instead there should be a more granular permission system that allows:
- giving or revoking access to individual functionality per user group (that means both in the UI and on the API endpoints)
- possibly allowing adding custom user groups for more complex usage scenarios

In order to do this we need to
- model user groups vs. permissions on a per-feature-level (roles)
  - determine which roles need to be recognized for vanilla OctoPrint and model them
  - add a way for plugins to define individual roles that can be offered for the user to attach to groups
- create some way to store user groups vs. associated permissions (currently that stuff is hard coded)
- add a permission editor
- document permissions in API documentation ("needs user group with role X")
- clearly communicate the limits of the permissions in the interface - we cannot prevent people from viewing the webcam if the URL is known since access to the webcam is provided by the webcam server, OctoPrint merely embeds it, so it cannot control access to it, that needs to be done by the webcam server
- Not only protect the existing API endpoints this way, but also the static endpoints as served by Tornado (e.g. timelapse and printable downloads)

foosel:
Documentation of Flask Prinicple: https://pythonhosted.org/Flask-Principal/

UM2EXTENDED:
This would be very useful improvement.

For beginner users it will be necessary to print files without the risk of damaging the printer by using the movement controls or the Terminal. Simple guest users don't provide this.

Especially for third parties this will be the minimum security requirement.

foosel:
Fun fact: movement controls are disabled while printing for that exact reason.

UM2EXTENDED:
Sure I know while printing it's off, but not before and after.

In my case I granted access to a beginner user and he watched the printer head moving trough the camera (top view position), but he didn't noticed the heater bed was at Z0 and he moved it down (up arrow) several times. Now the Z motor is damaged and need to be replaced on UM2 Extended.

foosel:
[...]

PR #2198, opened 2017-11-02
[WIP] More granular permission system

Salandora:
Alright, as mentioned in the IRC chat, I wanted to do a PR for this little project... Erm, huge project.

I tackled the more Granular Permission request (#1110).
It's not done yet, this here shall serve as a discussion platform, for a better overview of what has been implemented by now and what still needs to get implemented.

So to some informative stuff:
1. Implemented:
    - Permissions
    - Groups
    - Possibility to create new Groups through the API and therefore through the Web Interface
    - API checks for permissions
    - Web Interface checks for permissions
    - Plugins can define their own permissions
    - Admin and Guest groups are predefined non deletable groups
       - Admin group is static, admins will have all permissions
       - Guest group can be changed
    - Migration from old role system to new more granular permission system 
(still based on roles, but many more)

2. What is missing:
    - A bunch of useful automatic Tests
    - Some more hardening in the backend (Informations required)
      - How to protect the dataupdater?
      - How to protect the plugin messages?
      - Any way to generally protect the SockJS interface?
    - Some sort of fullscreen login (like the ForceLogin plugin) if a guest has absolutely no rights.
(Needs a bit of brainstoming)

Hopefully I haven't forgoten any important information.
Right now the only thing left for me is to wish everyone who wants to take a look at the code much fun and good luck :-)

Greetings
Salandora


src/octoprint/users.py
```diff
@@ -1,647 +1,28 @@
...
-
-##~~ Apiuser object to use when global api key is used to access the API
-
-
-class ApiUser(User):
-	def __init__(self):
-		User.__init__(self, "_api", "", True, UserManager.valid_roles)
+		from octoprint.server import groupManager
+		groups = [groupManager.guests_group]
+		if "user" is None:
```
foosel: I think you mean ``if "user" in roles``? Because that will never become true ;)

foosel:
Just for the record:

> Some sort of fullscreen login (like the ForceLogin plugin) if a guest has absolutely no rights. (Needs a bit of brainstoming)

This at least should be covered as of 6d1f610. Currently it tests if the user has at least the `STATUS` and if not shows the dialog. We could of course also switch that to no permissions at all, but I figured, not having the `STATUS` permission would actually leave a pretty non-functional interface.

foosel:
> How to protect the plugin messages?

How about changing `PluginManager.send_plugin_message(self, plugin, data)` into something like `PluginManager.send_plugin_message(self, plugin, data, permissions=None)`? That way plugins could provide additional limitations for messages they send, which the registered `PrinterStateConnection` instance (which I'm btw starting to think needs a rename 🤔) could then use to check against the authenticated user (about that, see [this gist](https://gist.github.com/foosel/68dfdbd9abeec4feb4a431d010c75bcf) of my initial idea how to accomplish that). Taking this a step further - we could adjust the `_emit` method to take additional permissions to check and have the other `on_*` methods adjusted accordingly (mostly to check against `Permissions.STATUS`).

Which brings me to think... do we maybe want to declare `STATUS` as something of a minimal limit and ALWAYS check plugin messages at least against that? Or do we maybe need something even more basic than this?

foosel:
I did a first throw of what we talked about on IRC and here with regards to the socket lock down. Also added propagation of changes (with funny results - removing or adding the admin permission on a user now has the wrench pop up and out in a different browser). I still see some issues with the propagation of the group changes though, and a recent merge of devel broke some tests. I'll investigate that further.

foosel:
At the risk of only talking to myself here ;) - Group changes now propagate properly, tests are fixed as well and I did a bit of refactoring to make the group/permission dialogs easier to take in. Some CSS stuff is still misbehaving though (of course they are).

jneilliii:
How is this going to effect plugins and simpleapi mixin calls?

foosel:
@jneilliii we'll make sure that `user_permission` and `admin_permission` stay there (albeit with a deprecation warning), and map to the new admin and user groups. So if you use the old mechanisms anywhere in your plugin to restrict APIs of any kind (simple or blueprint) to specific classes of users, that will continue to work. Still, long term goal should be that once 1.4.0 hits with the new permissions in place, plugins will utilize it.

foosel:
Some visual changes. 

User list:

image

Group list:

image

User editor:

image

image

Also several bugs fixed again that crept in. I have the feeling that some permissions are not properly applied, switched a user from guest to user and the printer profile list and file list stayed empty. Also noticed:

  * [ ] API key generation in user manager works, but UI doesn't properly refresh. Probably some minor JS error.
  * [ ] Active flag changes so far do not log out users - this should be changed now that we also allow live changes when permissions change.
  * [ ] `user_permission` compatibility layer is still missing. Do we simply wrap a check of `not user.is_anonymous` here?
  * [ ] I'm still unsure whether we should enforce the "Users" group on all logged in users or not. Right now we don't.

DongerZone:
Wondering if there is work to be done on this? I was about to spend my holidays working on this but saw its already in progress!

foosel:
Nope, it's merged to `devel`

DongerZone:
Exciting, can't wait to see it in effect!

PR #2217, opened 2017-11-12
Fix __future__ import in util

Depaulicious:

It fixes a syntax issue that CPython implementations simply ignore, but makes PyPy crash.

Start OctoPrint with PyPy. Without this commit you'll have this error:

```
  File "./octoprint", line 11, in <module>
    load_entry_point('OctoPrint==1.3.6.dev130+g79581c39', 'console_scripts', 'octoprint')()
  File "/home/octoprint/pypyvenv/site-packages/OctoPrint-1.3.6.dev130+g79581c39-py2.7.egg/octoprint/__init__.py", line 557, in main
    from octoprint.cli import octo
  File "/home/octoprint/pypyvenv/site-packages/OctoPrint-1.3.6.dev130+g79581c39-py2.7.egg/octoprint/cli/__init__.py", line 136, in <module>
    from .dev import dev_commands
  File "/home/octoprint/pypyvenv/site-packages/OctoPrint-1.3.6.dev130+g79581c39-py2.7.egg/octoprint/cli/dev.py", line 234, in <module>
    @dev_commands.group(name="dev", cls=OctoPrintDevelCommands)
  File "/home/octoprint/pypyvenv/site-packages/click-6.2-py2.7.egg/click/core.py", line 1169, in decorator
    cmd = group(*args, **kwargs)(f)
  File "/home/octoprint/pypyvenv/site-packages/click-6.2-py2.7.egg/click/decorators.py", line 115, in decorator
    cmd = _make_command(f, name, attrs, cls)
  File "/home/octoprint/pypyvenv/site-packages/click-6.2-py2.7.egg/click/decorators.py", line 89, in _make_command
    callback=f, params=params, **attrs)
  File "/home/octoprint/pypyvenv/site-packages/OctoPrint-1.3.6.dev130+g79581c39-py2.7.egg/octoprint/cli/dev.py", line 26, in __init__
    from octoprint.util.commandline import CommandlineCaller
  File "/home/octoprint/pypyvenv/site-packages/OctoPrint-1.3.6.dev130+g79581c39-py2.7.egg/octoprint/util/__init__.py", line 8
SyntaxError: __future__ statements must appear at beginning of file
```

#### Further notes

(Sorry I've had very little time to check the other PRs and issues I've been contributing to)
By the way, OctoPrint is a lot faster on PyPy. You should try it. Install python dependencies + `pypy` and `pypy-dev` for headers.

### DIFF
--- a/src/octoprint/util/__init__.py
+++ b/src/octoprint/util/__init__.py
@@ -1,11 +1,10 @@
 # coding=utf-8
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function
 
 """
 This module bundles commonly used utility methods or helper classes that are used in multiple places withing
 OctoPrint's source code.
 """
-from __future__ import absolute_import, division, print_function
 
 __author__ = "Gina Häußge <osd@foosel.net>"
 __license__ = 'GNU Affero General Public License http://www.gnu.org/licenses/agpl.html'


foosel:
Thank you for specifying HOW you are able to trigger this problem. I had someone report this a while ago, never got an answer how to reproduce (or a stack trace for that matter), couldn't myself and hence was baffled. Especially since the Python Language References [makes it sound like this perfectly valid](https://docs.python.org/2/reference/simple_stmts.html#future):

> A future statement must appear near the top of the module. The only lines that can appear before a future statement are:
>
>  * the module docstring (if any),
>  * comments,
>  * blank lines, and
>  * other future statements.

I'm btw surprised that OctoPrint even runs on PyPy. I would have expected issues with some dependencies.

Note that I merged this against `maintenance`. Master is the current stable release and usually only updated on releases - I do make exceptions for the occasional documentation update, but this is an actual code change, so it will have to wait until 1.3.6.

Depaulicious:
> Especially since the Python Language Reference makes it sound like this perfectly valid

In fact it is, PyPy seems stricter than CPython. It's not a big change, in fact there was even a duplicated import.

> I'm btw surprised that OctoPrint even runs on PyPy. I would have expected issues with some dependencies.

Me too, I was very surprised that "syntax" issue was the only issue. I had issues with plugins though: the Telegram plugin depends on Pillow, which doesn't really like the ARM+PyPy combo. It compiles and installs without warnings, but some extension module doesn't get pulled in. I reported issues to both projects, https://github.com/fabianonline/OctoPrint-Telegram/issues/96 and https://github.com/python-pillow/Pillow/issues/2850.

Aside of that, you can really feel the speed up on PyPy. On CPython it takes something like 10-20 seconds to start, on PyPy it took less than 10 seconds. Also everything is a bit more responsive. In fact, most of the code is pure python and PyPy does a great job with it.

> Note that I merged this against maintenance. Master is the current stable release and usually only updated on releases - I do make exceptions for the occasional documentation update, but this is an actual code change, so it will have to wait until 1.3.6.

I know and I always forget about it, I'll try to remember to use that branch as the base next time.

Depaulicious:
I filed an issue to PyPy: https://bitbucket.org/pypy/pypy/issues/2701/future-imports-not-allowed-after

PR #2233, opened 2017-11-22
Respond with 411 Length Required when missing

gregersrygg:

Changes HTTP response code for file upload to 411 (Length Required) instead of 400 (Bad Request) when Content-Length is missing. It is very hard for the client to guess that length is required by the server when the response doesn't say anything about why it's a bad request, and it's perfectly valid to do a file upload without a Content-Length in the header. Using 411 response code tells the client _why_ it's a bad request.

I don't know python, and the documentation is not good enough to explain how to setup and run the unit tests. The only thing this change will possibly break is an already broken client, or a unit test that is hard-coded to expect 400 (which is essentially the same as 411 - just less informative).

https://tools.ietf.org/html/rfc7231#section-6.5.10

### DIFF
--- a/src/octoprint/server/util/tornado.py
+++ b/src/octoprint/server/util/tornado.py
@@ -184,7 +184,7 @@ class UploadStorageFallbackHandler(tornado.web.RequestHandler):
 			if self.is_multipart():
 				if not self._bytes_left:
 					# we don't support requests without a content-length
-					raise tornado.web.HTTPError(400, log_message="No Content-Length supplied")
+					raise tornado.web.HTTPError(411, log_message="No Content-Length supplied")
 
 				# extract the multipart boundary
 				fields = self._content_type.split(";")


foosel:
Very good point. I took the liberty to cherry-pick this to the `maintenance` branch since it's not a new feature but rather an improvement or even a bug fix and should hence go with the next stable release 1.3.6. Github doesn't properly detect cherry picks, so I'm closing this PR manually. Your commit is merged as b002e41a005f69af855f53ef25d35983a9869429.

> the documentation is not good enough to explain how to setup and run the unit tests

I've also fixed that on the `maintenance` branch, already visible [on that branch on docs.octoprint.org](http://docs.octoprint.org/en/maintenance/development/environment.html). The [contribution guidelines](https://github.com/foosel/OctoPrint/blob/maintenance/CONTRIBUTING.md#pull-requests) have also been updated accordingly on `maintenance` and that will go live with the next release.

PR #2234, opened 2017-11-22
Doc/suse linux installation

caikn:
add installation steps for development under Suse Linux

After running the commands as described in doc, the octoprint development environment is up and running in my SLES 12 SP2


foosel:
Thanks!

PR #2239, opened 2017-11-23
Improve the checksumming speed.

alex-gh:
This PR improves the checksumming speed by a factor of 1.4-3.4. It does so by avoiding the overhead of calling ord() for every single character.

It has been tested by providing the same data to the old and the new algorithm and getting the same result back.


foosel:
\o/

Issue #1545, opened 2016-10-14
[Request] Include username of user who started print in "State" pane


twilde:
When access control is turned on, it would be really helpful to have the username of the user who started a print (or loaded the file for printing, or maybe both?) in the "State" pane, either all the time or as an option. Thanks for the awesome software!

somebugis:
Hello @twilde 
I'll try do to it this week

somebugis:
What do you mean by have the username ? Just the user who inquire his name before the print or rather a login page ?

twilde:
I was referring to the username the person logged in with, when access control is turned on.

somebugis:
![login](https://cloud.githubusercontent.com/assets/23096933/20007577/3691dd32-a29d-11e6-98b2-d20f861d8eb5.png)


This one ?

twilde:
Yup, that's the one!

cmtec:
+1 
When running a multi user environment this would help a lot.

foosel:
This is available on `maintenance` and `devel` and will be released as part of 1.3.9

Issue #652, opened 2014-11-15
[Request] log filament usage of every user


n-dim:
Having the user management in OcotoPrint it would be great to track the filament usage on a per user base. This way in our fablab we could account the filament usage of different people very easily. The problem is often, if you stop a print midway, you don't know how much filament was used so far. But OctoPrint could track this by constantly logging the executed extrusion length, right?

stv0g:
Just voting for this feature :+1: 

And maybe we could a cost estimation for filament too? Its just a rough estimation. But taking the filament length and width we could calculate the volume/weight.

JoseAlcerreca:
+1, but no need to estimate cost in my opinion.

coniferous1:
+1 Cost would eventually be nice, but I think just keeping track of what every user is using is primary. Cost could come later.

n-dim:
I think once the usage is tracked a cost calculation is just a single input field in the settings and a simple multiplication with the usage. I think the usage calculation is what is complicated about this feature. Once this is done the cost calculation would not be hard to implement. But overall I'm just guessing.

miton18:
+1 very interesting feature, just lenght filement per user

thirtythreeforty:
I would like this feature as well, mainly just the per-user logging.  In addition, I think it would be nice to have an option to store the cost of filament (like @n-dim said -- it should not be hard), so that the filament estimate metric could include the cost of the print, like this:

![index4](https://cloud.githubusercontent.com/assets/3952613/7507587/84635a92-f437-11e4-8bbf-288c3b957e6f.png)

I know that my users don't care about how much filament a print will take, but they might care about how much it costs, especially if their approach is "measure once, print twice."

thirtythreeforty:
@foosel I'm interested in writing a logging system like this one.  Do you have an opinion on how it should store logged data on the backend?  It would be nice to back everything with a database so we could run queries such as "who printed things on 9 May 2015?" or "how much filament has user *X* used?".

arekm:
Looks like this requests improved version of https://github.com/imrahil/OctoPrint-PrintHistory (missing per user info and actual filament usage)

foosel:
User who started a print will be available starting with 1.3.9. Everything else is will be up to plugin authors 
 utilizing that information.

foosel:
Forgot to close this on release of 1.3.9 which added the data needed for plugins to implement this.

PR #2257, opened 2017-12-02
[WIP] Add user informations in print events

gillg:
Add user which launch print and make current event in all print events

checklist:

    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

* Add user which launch print in all print events : "print_user"
* Add current event user on all print events (eg. canceller user) : "event_user"
These informations have no effects for retrocompatibility, and could be used for new plugins.
In my case I think use them in "history plugin" to set automatically the good user.

#1545 
#652

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -1222,10 +1222,13 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 			_, name = self._fileManager.split_path(FileDestinations.LOCAL, path)
 			origin = FileDestinations.LOCAL
 
+		from flask.ext.login import current_user
+
 		result= dict(name=name,
 		             path=path,
 		             origin=origin,
 		             size=print_job_size,
+					 event_user=current_user,
 
 		             # TODO deprecated, remove in 1.4.0
 		             file=full_path,
@@ -1233,6 +1236,8 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 
 		if position is not None:
 			result["position"] = position
+
+		result["print_user"] = self._comm.getPrintUser()
 
 		return result
 
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -1228,7 +1228,7 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 		             path=path,
 		             origin=origin,
 		             size=print_job_size,
-					 event_user=current_user,
+		             event_user=current_user,
 
 		             # TODO deprecated, remove in 1.4.0
 		             file=full_path,
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -624,6 +624,9 @@ class MachineCom(object):
 			return None
 		return self._currentFile.getFilepos()
 
+	def getPrintUser(self):
+		return self._currentPrintUser
+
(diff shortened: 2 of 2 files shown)


src/octoprint/printer/standard.py
```diff
@@ -1222,10 +1222,13 @@ def _payload_for_print_job_event(self, location=None, print_job_file=None, print
 			_, name = self._fileManager.split_path(FileDestinations.LOCAL, path)
 			origin = FileDestinations.LOCAL
 
+		from flask.ext.login import current_user
+
 		result= dict(name=name,
 		             path=path,
 		             origin=origin,
 		             size=print_job_size,
+		             event_user=current_user,
```
foosel: This doesn't make much sense - `Printer` operates outside of the context of any active flask requests, so attaching users to events triggered internally doesn't make sense.

src/octoprint/util/comm.py
```diff
@@ -840,6 +843,9 @@ def startPrint(self, pos=None):
 
 				self.resetLineNumbers()
 
+				from flask.ext.login import current_user
```
foosel: See my other comment - this won't work like that in all cases. Instead you should add a new parameter to the the start print methods that allows to forward a user id (which should however stay optional for backwards compatibility) which should then be attached to the currently selected file information in printer and comm layer.

``current_user`` won't always be set since you won't always have a request context here. In any case, nothing past the REST API implementation should ever depend on flask stuff if at all possible. Backend modules shouldn't need to know about frontend facing API implementations.

gillg: Thanks for your reply. I thinked about api connections, but I not found how always get the good user wich lunch printing (by api, flask or anything else). The goal of this data in events objects is to capture the "initiator" user of the print action. It usefull for history plugin or a future printing service management (eg fablab). The canceler user could be also a good information. In my pr user field is potentially different of print_user

foosel: I think I didn't explain myself well.

I wholeheartedly agree that adding the user who started (or cancelled or otherwise interacted with) a job to the corresponding events is a good idea! The problem here isn't adding the user, it's HOW you add it, by using the `current_user` implemented from flask. This field is only available in the context of an active request, NOT when an event gets triggered asynchronously through the system.

My point is, you'll have to add a userid parameter to the methods in question and populate that accordingly throughout the system, not just rely on `current_user` being available because it won't be in a lot of cases where print related events will be generated. 

Take this example.

User "foo" starts a print job. `current_user` is available when he does that and so `PrintStarted` gets fired with "foo" attached as user. The print continues for a while. Then the firmware sends a [pause action command](), making the print pause. A `PrintPaused` event is triggered, but `current_user` is not set (no request active). So now you have a problem.

Additionally you are tightly coupling `PrinterInterface`, `StandardPrinter` and even the `MachineComm` class to the technology of the *frontend facing API*. That's a strong code smell. It increases maintenance overhead, makes the code more tricky to test and so on. The printer object and even more so the communication layer shouldn't have to know that it gets called from a flask based API and that their might be a user stored in flask's `current_user`. It should simply get told "hey, user with id 'foo' started/paused/resumed/cancelled" this print, not try to determine that itself based on implementation details of other system components it should have no deeper knowledge of.

foosel:

PR #2258, opened 2017-12-02
[WIP] Setup file cursror location during file analyzis

gillg:
checklist:

    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

----

Analyze gcode file (filament usage etc.) regarding the end position of file cursor.
For exemple, if print is canceled, analysis should read GCODE from begining to cancel point in file, instead of the full file.
Add also multi extruder support (more than 2...)

### DIFF
--- a/src/octoprint/filemanager/analysis.py
+++ b/src/octoprint/filemanager/analysis.py
@@ -249,6 +249,7 @@ class AbstractAnalysisQueue(object):
 			                                                       "path": entry.path,
 			                                                       "origin": entry.location,
 			                                                       "type": entry.type,
+																   "position": entry.position,
 
 			                                                       # TODO deprecated, remove in 1.4.0
 			                                                       "file": entry.path})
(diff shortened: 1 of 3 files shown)


src/octoprint/filemanager/analysis.py
```diff
@@ -384,7 +387,7 @@ def _do_analysis(self, high_priority=False):
 			if analysis["extrusion_length"]:
 				result["filament"] = dict()
 				for i in range(len(analysis["extrusion_length"])):
-					result["filament"]["tool%d" % i] = {
+					result["filament"]["tools"][i] = {
```
foosel: Please do not change the internal format used here. It might already be used by some plugins and by changing this you are making things backwards incompatible. It's also used in the frontend as-is, so this change already breaks filament usage display there.

gillg: I understand the retrocompatibility issue, but you can't iterrate safely on its. maybe we could add (instead to replace) the new entry "tools" and set old "tool%i" as deprecated.

This is used by history plugin for example (it use only "tool1" and "tool2"...), but I made a PR too.

gillg:
The motivation behind this PR is to handle correcly the filament volume used at printing end.
This is used for exemple by history plugin and filament manager plugin.

If you cancel print, le full théorical volume is used instead of real volume used.

src/octoprint/filemanager/analysis.py
foosel: Huh? You can iterate perfectly fine over the keys in this dict. I don't understand why you want to change this, there is no reason (especially not within the context of this PR) and it only produces compatibility issues.

foosel:
So the goal of this PR is to enhance the analysis sub command thus that plugins might call it with a position parameter to only get a partial usage?

foosel:

PR #2290, opened 2017-12-15
disable connection state change button while printing

ntoff:

Disables the connection state change button while a print is in progress

* Connect to a printer (virtual printer was used for testing)
* Start a print
* Note that the connect / disconnect button is now unable to be clicked while a print is in progress or paused

Fullfills a feature request, but in a slightly different way. The feature request was to throw up a confirmation dialog, I believe simply disabling the  button is probably safer.

ref: https://github.com/foosel/OctoPrint/issues/2287

image

#### Further notes

None

### DIFF
--- a/src/octoprint/templates/sidebar/connection.jinja2
+++ b/src/octoprint/templates/sidebar/connection.jinja2
@@ -10,4 +10,4 @@
 <label class="checkbox">
     <input type="checkbox" id="connection_autoconnect" data-bind="checked: settings.serial_autoconnect, css: {disabled: !isErrorOrClosed()}, enable: isErrorOrClosed() && loginState.isUser()"> {{ _('Auto-connect on server startup') }}
 </label>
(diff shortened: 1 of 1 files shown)


foosel:
Hm... To be honest, I think a confirmation would be better. There might be legitimate reasons to want to disconnect while printing (e.g. you started an SD print, now want to switch your instance over to another printer connected to the machine, something with the comms got severly stuck and you want to trigger a restart of the controller, ...). A confirmation prevents accidents but still allows that to happen if you are positive that this is what you want.

ntoff:
well if that's the case, perhaps a command line option for "--expert mode" to run OctoPrint in a mode that disables any and all popups asking if we want to confirm our actions so those of us who hate confirmations, can go about our business without having to click multiple times to perform simple actions. Unless we want the settings dialog to be filled with settings for...

* confirm on cancel print
* confirm on shutdown
* confirm on disconnect
* confirm on starting a print (I can see someone wanting a confirmation on this soon)
* confirm on installing plugins
* confirm on confirming my confirmation
* confirm on saving settings

Issue #2297, opened 2017-12-17
flask-login returns 401 by default, messing with basic authentication from proxies


bzed:
#### What were you doing?

I'm using a haproxy (as documented somewhere in the wiki ;)) with basic authentication.
Basically this works fine as long as I use autologin on octoprint side.
If not, I'm getting - mainly from plugins - some extra basic auth requests while not being logged in as the default behaviour of flask-login is to return 401 for requests (a plugin trying to talk to the api triggers that for example - I think....).

According to the flask-login documentation in my opinion the proper thing to do would be to add a login view that does nothing but returning a 403 (instead of the 401, which is the default).

```
unauthorized()
This is called when the user is required to log in. If you register a callback with LoginManager.unauthorized_handler(), then it will be called. Otherwise, it will take the following actions:

Flash LoginManager.login_message to the user.
If the app is using blueprints find the login view for the current blueprint using blueprint_login_views. If the app is not using blueprints or the login view for the current blueprint is not specified use the value of login_view.
Redirect the user to the login view. (The page they were attempting to access will be passed in the next query string variable, so you can redirect there if present instead of the homepage. Alternatively, it will be added to the session as next if USE_SESSION_FOR_NEXT is set.)
If LoginManager.login_view is not defined, then it will simply raise a HTTP 401 (Unauthorized) error instead.
```
#### What did you expect to happen?

Getting 403 errors instead of 401 from flask-login.

#### What happened instead?

401 errors triggered basic auth again as the browser can't make a difference between OctoPrint beeing funny and a broken BA.

#### Did the same happen when running OctoPrint in safe mode?

Actually I don't think so as nothing is shown which would trigger the issue. Didn't test it, though :P If plugins can trigger it, its buggy anyway ;)

#### Version of OctoPrint

OctoPrint 1.4.0.dev652+gbad3612e.dirty (yes, dirty, pull request incoming ;))

#### Operating System running OctoPrint

deb stretch

#### Printer model & used firmware incl. version

doesn't matter

#### Browser and Version of Browser, Operating System running Browser

doesn't matter (latest chrome)

#### Link to octoprint.log

doesn't matter

#### Link to contents of terminal tab or serial.log

doesn't matter

#### Link to contents of Javascript console in the browser

doesn't matter

#### Screenshot(s)/video(s) showing the problem:

doesn't matter

I have read the FAQ.

foosel:
As I'm wrapping things up for 2017, I'll take a look at this come January! 🎄

[...]

PR #2304, opened 2017-12-19
Trust basic authentication headers for user login.

bzed:
This is my suggestion on how to handle basic authentication for users
mananaged by OctoPrint's UserManager.

In my opinion OctoPrint should not be doing Basic Authentication on its
own. The commen case is that it runs on some slow hardware which its not
able to handle stupid bots trying out thousands of passwords in short
time. Also you want to protect it as much as possible from the outside
to ensure that the basd internet can't break your 72h print after 71
hours by brute-forcing passwords.

So I'm introducing an
  accessControl: trustBasicAuthentication
config option. If enabled, it will take the user name from the
Authorization header, check if the user exists and is enabled - and if
so, just login the user.

OctoPrint works well for me with my haproxy basic auth setup - with disabled autologin.
Although I'm not sure if the settings should not me merged somehow into the autologin settings as it is something similar.

Also anonymous logins for a disabled user manager is not implemented. I'm not exactly sure if it is necessary at all, but it should not be hard to add.

highly appreciated!

checklist:

    Brainstorming ticket
  * [X] Your PR targets OctoPrint's devel branch, or maintenance if it's
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

The more details the better!

----

Using basic auth together with the normal login has some issues (#2297 for example), lets trust basic auth and avoid the double login part!

Setup a haproxy with basic auth as documented in the wiki and add a user with the same name in Octoprint and in the haproxy config.

none :)

not necessary.

#### Further notes
Comments welcome ;)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -87,6 +87,7 @@ date of first contribution):
   * [Peter Backx](https://github.com/pbackx)
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
+  * [Bernd Zeimetz](https://github.com/bzed)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/server/__init__.py
+++ b/src/octoprint/server/__init__.py
@@ -147,6 +147,19 @@ def load_user(id):
 
 	return None
 
+def load_user_from_request(request):
+	header = request.headers.get('Authorization')
+	if header:
+		header = header.replace('Basic ', '', 1)
+		try:
+			header = base64.b64decode(header)
+		except TypeError:
+			return None
+		id = header.split(':')[0]
+		user = userManager.findUser(userid=id)
+		if user:
+			return user
+	return None
 
 #~~ startup code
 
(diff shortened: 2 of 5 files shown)


foosel:
👍 I think this should be included in 1.3.7 already, would you mind changing that PR over to `maintenance`?

bzed:
I'll figure out what the difference was for maintenance. I *think* https://github.com/bzed/OctoPrint/commits/basic_auth_autologin worked okay-ish, but I'm not sure about the api login anymore. Will figure it out, but due to the principal stuff it needs a bit different code.

bzed:
@foosel https://github.com/bzed/OctoPrint/commits/basic_auth_autologin is rebased on the maintenance branch. basically the same code +- the merge conflict. Works for me, not really well tested, though.

foosel:
I've created #2405 based on that. If you could give that a test spin against your setup that would help a lot. I tested it against a local setup, but just in case... ;) 

Heads-up: I added an option that is enabled by default to not only fetch the user name from the header but also check if the password matches. This can be disabled however, just set `accessControl.checkBasicAuthenticationPassword` to false in `config.yaml`. Reason: I wanted to make it a bit more tricky for users to shoot themselves in the foot by having this enabled but their instance still also accessible without basic auth in front of it.

Closing this one since it's superseded by #2405.

PR #2327, opened 2017-12-28
grammatical fix in text

tedder:
[It's a common error](https://brians.wsu.edu/2016/05/31/setup-set-up/); I'm not sure if it needs to be fixed in all three places or perhaps only in the .pot file.

highly appreciated!

Have not tested.


tedder:
eh, didn't change branches cleanly. closing.

PR #2328, opened 2017-12-28
grammatical fix in text

tedder:
[It's a common error](https://brians.wsu.edu/2016/05/31/setup-set-up/); I'm not sure if it needs to be fixed in all three places or perhaps only in the .pot file.

Have not tested.


foosel:
The pot file gets generated regularly from those places, so that is great like it is, thanks! Also learned something that will hopefully stick 😄

PR #2346, opened 2018-01-07
software xon/xoff workaround

adriancuzman:
Yes, transferring to sdcard is very slow... with Marlin firmware, but with the new SERIAL_XON_XOFF implementation the speed can be increased.

I used Marlin firmware with SERIAL_XON_XOFF set. The host for Octoprint used was Linux with the software flow control bug. (I have the CP2102 USB-Serial chip)

Set to true in config file in the serial section "useSoftwareFlowWorkaround". If not set, the code is not active and will wait for 'ok' after each command. Example:

serial:
  autoconnect: true
  baudrate: 500000
  port: /dev/ttyUSB0
  useSoftwareFlowWorkaround: true   

After the config file is set. Transfer a file to sdcard. Check if file on sdcard is the same with the one sent. Transfer time should be less than before :)  

Information about the software flow control bug:  https://github.com/neundorf/CuteCom/issues/22

https://github.com/foosel/OctoPrint/issues/2261

#### Further notes
- had to increase deque to 1000 lines to have lines in history when resending commands
- there are cases when resend is triggered, but it works and the file in the end is complete without missing parts
- I don't know Python, I just learned a bit Python to make this PR, any feedback to make this code correct will be appreciated :)

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -313,6 +313,7 @@ class MachineCom(object):
 	STATE_TRANSFERING_FILE = 11
 
 	CAPABILITY_AUTOREPORT_TEMP = "AUTOREPORT_TEMP"
+	CAPABILITY_SERIAL_XON_XOFF = "SERIAL_XON_XOFF"
 
 	def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfileManager=None):
 		self._logger = logging.getLogger(__name__)
@@ -387,7 +388,19 @@ class MachineCom(object):
 		self._currentLine = 1
 		self._line_mutex = threading.RLock()
 		self._resendDelta = None
-		self._lastLines = deque([], 50)
+
+		self._use_xonxoff_workaround = settings().get(["serial", "useSoftwareFlowWorkaround"])
+		self._xonxoff_stopChar = chr(19)
+		self._xonxoff_resumeChar = chr(17)
+
+		maxLinesHistory = settings().getInt(["serial", "maxLinesHistory"])
+		if maxLinesHistory is None:
+			if self._use_xonxoff_workaround:
+				maxLinesHistory = 150
+			else:
+				maxLinesHistory = 50
+
+		self._lastLines = deque([], maxLinesHistory)
 		self._lastCommError = None
 		self._lastResendNumber = None
 		self._currentResendCount = 0
@@ -425,6 +438,8 @@ class MachineCom(object):
 		self._send_queue = SendQueue()
 		self._temperature_timer = None
 		self._sd_status_timer = None
+		self._pause_transmission = False
+
 
 		# hooks
 		self._pluginManager = octoprint.plugin.plugin_manager()
@@ -582,6 +597,22 @@ class MachineCom(object):
 		if state == self.STATE_TRANSFERING_FILE:
 			return "Transferring file to SD"
 		return "Unknown State (%d)" % (self._state)
+
+	def isSendingFileToSDWithSoftwareFlow(self):
+		#first determine if sending file to sd and printer supports xon_xoff
+		if 	self._state == self.STATE_PRINTING and self.isStreaming() and \
+			self._firmware_capabilities.get(self.CAPABILITY_SERIAL_XON_XOFF, False) and \
+			self._currentLine > 1:
+				# currentLine > 1 is to enable flow control only and ignore ok only after the first ok is sent
+				# if not handle_ok will not set clear_to_send and in send_loop it will wait forever
+				if self._serial != None and getattr(self._serial, "xonxoff", False):
+					#software flow control is enabled and handled in the usb-serial chip/driver
+					return True
+				else:
+					#software flow is handled in octoprint if workaround is enabled
+					if self._use_xonxoff_workaround != None and self._use_xonxoff_workaround:
+						return True
+		return False
 
 	def getErrorString(self):
 		return self._errorValue
@@ -892,6 +923,7 @@ class MachineCom(object):
 			self._currentFile.start()
 
 			self.sendCommand("M28 %s" % remoteFilename)
+			self._pause_transmission = False
 			eventManager().fire(Events.TRANSFER_STARTED, {"local": localFilename, "remote": remoteFilename})
 			self._callback.on_comm_file_transfer_started(remoteFilename, self._currentFile.getFilesize())
 
@@ -1609,7 +1641,9 @@ class MachineCom(object):
 		self._log("Connection closed, closing down monitor")
 
 	def _handle_ok(self):
-		self._clear_to_send.set()
+		can_send = not self.isSendingFileToSDWithSoftwareFlow()
+		if can_send:
+			self._clear_to_send.set()
 
 		# reset long running commands, persisted current tools and heatup counters on ok
 
@@ -1630,8 +1664,8 @@ class MachineCom(object):
 			self._resendNextCommand()
 		else:
 			self._resendActive = False
-			self._continue_sending()
-
+			if can_send:
+				self._continue_sending()
 		return
 
 	def _handle_timeout(self):
@@ -1912,13 +1946,17 @@ class MachineCom(object):
 					self._log("Failed to autodetect serial port, please set it manually.")
 					return None
 
+			software_xonxoff = settings().get(["serial", "xonxoff"])
+			if software_xonxoff is None:
+				software_xonxoff = False
+
 			# connect to regular serial port
 			self._log("Connecting to: %s" % port)
 			if baudrate == 0:
 				baudrates = baudrateList()
-				serial_obj = serial.Serial(str(port), 115200 if 115200 in baudrates else baudrates[0], timeout=read_timeout, writeTimeout=10000, parity=serial.PARITY_ODD)
+				serial_obj = serial.Serial(str(port), 115200 if 115200 in baudrates else baudrates[0], timeout=read_timeout, writeTimeout=10000, xonxoff=software_xonxoff, parity=serial.PARITY_ODD)
 			else:
-				serial_obj = serial.Serial(str(port), baudrate, timeout=read_timeout, writeTimeout=10000, parity=serial.PARITY_ODD)
+				serial_obj = serial.Serial(str(port), baudrate, timeout=read_timeout, writeTimeout=10000, xonxoff=software_xonxoff, parity=serial.PARITY_ODD)
 			serial_obj.close()
 			serial_obj.parity = serial.PARITY_NONE
 			serial_obj.open()
@@ -2017,6 +2055,17 @@ class MachineCom(object):
 		# finally return the line
 		return line
 
+	def _filterXonXoffCharacters(self, ret):
+		if self._xonxoff_stopChar in ret:
+			self._pause_transmission = True
+			ret = ret.translate(None, self._xonxoff_stopChar)
+		if self._xonxoff_resumeChar in ret:
+			self._pause_transmission = False
+			self._continue_sending()
+			self._clear_to_send.set()
+			ret = ret.translate(None, self._xonxoff_resumeChar)
+		return ret
+
 	def _readline(self):
 		if self._serial is None:
 			return None
@@ -2036,6 +2085,9 @@ class MachineCom(object):
 
 		if ret != "":
 			try:
+				if self.isSendingFileToSDWithSoftwareFlow() and self._use_xonxoff_workaround:
+					ret = self._filterXonXoffCharacters(ret)
+
 				self._log("Recv: " + sanitize_ascii(ret))
 			except ValueError as e:
 				self._log("WARN: While reading last line: %s" % e)
@@ -2392,6 +2444,9 @@ class MachineCom(object):
 						# definitely do not want that, so we tickle the queue manually here
 						self._continue_sending()
 
+					if self.isSendingFileToSDWithSoftwareFlow() and not self._pause_transmission:
+							self._continue_sending()
+							self._clear_to_send.set()
 				finally:
 					# no matter _how_ we exit this block, we signal that we
 					# are done processing the last fetched queue entry
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -2420,7 +2420,7 @@ class MachineCom(object):
 						command_requiring_checksum = gcode is not None and gcode in self._checksum_requiring_commands
 						command_allowing_checksum = gcode is not None or self._sendChecksumWithUnknownCommands
 						checksum_enabled = not self._neverSendChecksum and ((self.isPrinting() and self._currentFile and self._currentFile.checksum) or
-						                                                    self._alwaysSendChecksum or
+						                                                    self._alwaysSendChecksum or self._use_xonxoff_workaround or
 						                                                    not self._firmware_info_received)
 
 						command_to_send = command.encode("ascii", errors="replace")
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -392,6 +392,8 @@ class MachineCom(object):
 		self._use_xonxoff_workaround = settings().get(["serial", "useSoftwareFlowWorkaround"])
 		self._xonxoff_stopChar = chr(19)
 		self._xonxoff_resumeChar = chr(17)
+		self._waitingForM28Confirmation = False
+		self._startedSendingFileToSD = False
 
 		maxLinesHistory = settings().getInt(["serial", "maxLinesHistory"])
 		if maxLinesHistory is None:
@@ -600,9 +602,8 @@ class MachineCom(object):
 
 	def isSendingFileToSDWithSoftwareFlow(self):
 		#first determine if sending file to sd and printer supports xon_xoff
-		if 	self._state == self.STATE_PRINTING and self.isStreaming() and \
-			self._firmware_capabilities.get(self.CAPABILITY_SERIAL_XON_XOFF, False) and \
-			self._currentLine > 1:
+		if 	self._startedSendingFileToSD and \
+			self._firmware_capabilities.get(self.CAPABILITY_SERIAL_XON_XOFF, False):
 				# currentLine > 1 is to enable flow control only and ignore ok only after the first ok is sent
 				# if not handle_ok will not set clear_to_send and in send_loop it will wait forever
 				if self._serial != None and getattr(self._serial, "xonxoff", False):
@@ -1642,6 +1643,11 @@ class MachineCom(object):
 
 	def _handle_ok(self):
 		can_send = not self.isSendingFileToSDWithSoftwareFlow()
+		if self._waitingForM28Confirmation:
+			self.__waitingForM28Confirmation = False
+			self._startedSendingFileToSD = True
+			can_send = True
+
 		if can_send:
 			self._clear_to_send.set()
 
@@ -2663,11 +2669,13 @@ class MachineCom(object):
 			self.setPause(True)
 
 	def _gcode_M28_sent(self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs):
+		self._waitingForM28Confirmation = True
 		if not self.isStreaming():
 			self._log("Detected manual streaming. Disabling temperature polling. Finish writing with M29. Do NOT attempt to print while manually streaming!")
 			self._manualStreaming = True
 
 	def _gcode_M29_sent(self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs):
+		self._startedSendingFileToSD = False
 		if self._manualStreaming:
 			self._log("Manual streaming done. Re-enabling temperature polling. All is well.")
 			self._manualStreaming = False
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -86,6 +86,7 @@ date of first contribution):
   * [Peter Backx](https://github.com/pbackx)
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
+  * ["Adrian Cuzman"](https://github.com/adriancuzman)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -86,7 +86,6 @@ date of first contribution):
   * [Peter Backx](https://github.com/pbackx)
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
-  * ["Adrian Cuzman"](https://github.com/adriancuzman)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and


src/octoprint/util/comm.py
```diff
@@ -387,7 +387,7 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
 		self._currentLine = 1
 		self._line_mutex = threading.RLock()
 		self._resendDelta = None
-		self._lastLines = deque([], 50)
+		self._lastLines = deque([], 1000)
```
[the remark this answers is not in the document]
adriancuzman: Yes, when enabled in Marlin, software flow control is used all the time, but as a workaround and experimental feature, I wanted to be enabled just for SD transfers... Is there a benefit for this in other cases? For example when printing directly from serial?

It was more like 100 didn't work, 200 nope... ok 1000 it is :) I'll try to figure out which lower value works. But I think this is system dependent (3D Printer + PC) I think this should be configurable from the config file...

src/octoprint/util/comm.py
```diff
@@ -425,6 +425,10 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
 		self._send_queue = SendQueue()
 		self._temperature_timer = None
 		self._sd_status_timer = None
+		self._pause_transmission = False
+		self._use_xonxoff_workaround = settings().get(["serial", "useSoftwareFlowWorkaround"])
```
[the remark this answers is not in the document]
adriancuzman: When I worked on this I had  Marlin 1.1.6  and it didn't report it, I've upgraded to 1.1.7 and now it's reported. I'll update to use the response to M115.

src/octoprint/util/comm.py
```diff
@@ -583,6 +587,16 @@ def getStateString(self, state=None):
 			return "Transferring file to SD"
 		return "Unknown State (%d)" % (self._state)
 
+	def isSendingFileToSDWithSoftwareFlow(self):
+		# self._currentLine > 1 - active only after the first ok is received, if not, it will lock...
```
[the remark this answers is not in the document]
adriancuzman: Yes, this could be a problem.. the check is that to switch to software flow control after the first line is sent. If not, it will lock, if the ok is received, contine_transfer  from handle_ok will not be executed and in send_loop it will not be executed too... so it locks. I think the issue here is the order in time how the flags are set and how handle_ok and send_loop are executed in time.. any idea on how to fix this?

src/octoprint/util/comm.py
```diff
@@ -583,6 +587,16 @@ def getStateString(self, state=None):
...
+		if (self._use_xonxoff_workaround != None and self._use_xonxoff_workaround) and \
+				(self._serial != None and not self._serial.xonxoff) and \
```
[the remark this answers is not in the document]
adriancuzman: Thank you for mentioning this :) I will change it, as I said, I haven't worked with Python before... I'm more used to C/C++/Java languages :))

src/octoprint/util/comm.py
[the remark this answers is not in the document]
foosel: Configuration option sounds good. Maybe also just increase the size once XONXOFF support is detected?

src/octoprint/util/comm.py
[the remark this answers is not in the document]
foosel: 👍

src/octoprint/util/comm.py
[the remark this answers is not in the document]
foosel: I have to admit, I'm not entirely sure what the core problem is that line number check is attempting to fix. I understand it locked up if no `ok` had been received yet? But why/where?

In any case, instead of checking for a specific line number which as pointed out can change mid-print or really at any time thanks to an `M110`, maybe a global flag after initial handshake might make more sense. But that depends a bit on the original issue which sadly - as I said - I can't fully wrap my head around right now.

Any clarification would be highly appreciated 😅

adriancuzman:
Hi,
  I managed to update the PR, I've added "xonxoff" configuration switch in the serial group and tested on Windows (the workaround is not needed on Windows, because the USB-Serial supports software flow control, on Linux the driver does not support it, so the workaround is needed, in my case I have a Trigorillaboard with cp2102 chip). So when the system supports software flow control, in config xonxoff can be set to true and xonxoff will be set to true in PySerial. If the system does not support software flow control (for example Linux) then xonxoff must be set to false (if set to true, the xonxoff characters will be filtered by the linux driver, but the flow will not be stopped) and useSoftwareFlowWorkaround should be set to true (also in the serial group of the config file). The resend history length depends on the system and baud rate, on my laptop, a queue length of 150 is ok, but on my OrangePi at 500000 baud is not. The length can be configured also with the maxLinesHistory in the config file (also in the serial group). 

I've tried to use software control flow, but I couldn't make it work... it locks and it's beyond me, so for now I would leave it only for SD transfers. One issue is when the logic changes to flow control from ok confirmation. For SD transfer this is fixed with line number >1, but if enabled all the time, this change will happen after M115 and the connection will lock and timeout...  

Also I've tried to transfer a file with line reset in the middle of the file (M110), the file transfer will fail with and without the workaround. This could be an issue, but do slicers add M110? 

I will use this branch on my OrangePi connected to the printer to see how it will work with everyday work.
Please review this PR when you have time and any feedback will be appreciated. Thank you :)

foosel:
Just a quick heads up, I'm busy sneezing my guts out, will take a look when I'm back on my feet.

adriancuzman:
Oh, sorry to hear that :( ... yes, it's the flu season... and this year it's a nasty version. It's ok, the PR will be here, no need to hurry, health is more important :)

foosel:
As a quick follow-up to the resend shenanigans observed there - I just saw that there appears to be an off by one error at work here, at least the firmware is requesting 42 but OctoPrint sends 43 (after going through its own send queue). And the rest of endless "send 42 - nope" is caused by OctoPrint's workaround for some firmwares out there not clearing their rx buffers after a resend request, meaning anything that was sent before the resend request was triggered will still be processed but not fit the order of things, causing consecutive resend requests. OctoPrint hence ignores these. Usually that's fine since stuff quickly rectifies itself, but here we are handling large histories and also have this off by one situation, and that leads to this huge segment of requests.

adriancuzman:
I'm glad that you're feeling better :)

Sorry for the late response, I was focused on fixing this issue, I found out this by using the printer a few times..The problem is that I had one configuration that worked, tested a 500k file and after a 8Mbyte file (with getting the sd card and comparing the content on the pc), after that I changed the firmware, installed a new probe, made changes and after that it didn't work anymore, tried to remake the firmware settings, but still turned with random missing parts. I'm sorry, I can't explain it. In the mean time I connected from the OrangePI (a Pi3 clone) to the printer using UART directly (using a voltage divider and it works at 500k baud). This is just to get around the faulty USB-Serial driver. The UART driver supports XON/XOFF and it works. I made also a small python program that reads a file and sends it over serial. The file is sent 100% every time without errors, no resend errors. But using Octoprint, after ~10k lines I get resend line errors, each time (which is over 200kbytes). I've tried a lot of things but without a solution, it's driving me crazy, compared to the simple program, the read and write are on different threads, but it's not that, tried to limit how many lines are sent ahead against how many OKs received, but then it's synced against the OKs and no speedup. I also found the resend + 1 issue, when debugging the issue, but I think this is specific to this case, or it's for all cases, I had never seen a resend when using the printer normally...  Your firmware configuration is correct, I've tried with TX buffer size 32, but didn't made a difference, tried 500k, 250k , the same, there is no way to be sure the file is 100% not modified on the sd.

I've attached the program I used to test the connection {sorry, it look horrible, like a quick hack and the CRC code is from Octoprint :) } Any ideas on why this one works 100% and in Octoprint after 200k the firmare fails? The clear to send, continue sending logic is not correct? I'm out of ideas.. :(

Also, the XONXOFF drivers should be ok on Windows, so to make a quick test and see the speeds when XON/XOFF is working, on OrangePI I get ~1.5Mybtes/min with 500kbaud, I would be super happy with this if it only worked :))

[pytestCRC.txt](https://github.com/foosel/OctoPrint/files/1729612/pytestCRC.txt)

adriancuzman:
One more thing,  in the Marlin firmware, they send the stop send char when the RX buffer is 12.5% full (RX_BUF_SIZE / 8), they consider this should be enough time for the sender to receive the char and stop sending without overflowing the buffer. So it's very important to respond as quickly as possible and stop sending. This is one thing that I was unable to fix, I think it takes too much time to stop sending data. But still this shouldn't be an issue when XON/XOFF is implemented in the driver, not 100% sure, I don't know how it's implemented if too much data is send (buffer, how much? or it blocks the write method?). And I don't understand why with the small program it works and when using Octoprint the firmware sends resend errors, it should not happen, the driver should stop sending too much data... I really have no explanation.. :/

adriancuzman:
I think I found a solution, the problem was how Marlin working when requesting a resend, on serial if a resend is requested, the rx buffer is flushed and everything that comes for some time are not proper lines and incomplete lines that have '\n' will be received, but if they don't start with 'N' they will be consider simple commands (without CRC) and "will" be added in the queue, that's why in the sd file we have more data :)) So the problem is when a resend is requested. On lower speeds 38-57k bauds the resend is not triggered and it works, but over 57k some resends can occur and that makes the file not usable. 

The firmware I used on my printer is: https://github.com/adriancuzman/Marlin/tree/1.8_octoprint_test
The last commit has the possible fix for the resend issue.

Also this is the Octoprint that I used, I've added +1 to resend delta,even if it sends more lines, it should be ok.
https://github.com/adriancuzman/OctoPrint/tree/softwareFlowControlWorkaround2

I'll make proper pull requests when I get back,  I'll be away from home until next week. This is just a quick update on the issue. Also thank you for taking time to help with this issue :)
Have to run, bye.

foosel:
> the problem was how Marlin working when requesting a resend, on serial if a resend is requested, the rx buffer is flushed and everything that comes for some time are not proper lines and incomplete lines that have '\n' will be received

So instead of throwing away any garbage that still might be in there, it's happily processing it? Great. So just like what I already feared:

> some firmwares out there not clearing their rx buffers after a resend request, meaning anything that was sent before the resend request was triggered will still be processed

The problem I see is that while your fix may work on first glance, it will cause quite some trouble if another communication issue arises on the line while a resend is already active. Imagine the following:

  * Host sends line 23
  * Firmware receives line 23 but the checksum doesn't match
  * Firmware requests a resend of line 23 and sets error flag
  * Host sends line 23 again
  * Firmware receives line 23 but the checksum doesn't match yet again
  * Firmware doesn't do anything because error flag is already set
  * Host doesn't do anything because firmware is not replying
  * => communication is now stalled

What the firmware should do instead here in my eyes is *first* clear out anything still in the RX buffer and simply throw it away (maybe with the exception of `M112`), *then* send the resend request to the host. What it does instead is send a resend request to the host and then get confused by what's still lying around unprocessed in its RX buffer from before it even noticed something was off itself, let alone notified the host about it.

> I've added +1 to resend delta,even if it sends more lines, it should be ok.

Hm, I'd rather figure out why the resend is off by one instead of intentionally getting it wrong. Because that reeks of something that will bite me in the arse down the road. The resend handling is a fickle beast and must be treated with caution.

adriancuzman:
I've worked more this week on this issue and have a solution, more or less... but at least I know the problems that stops it from working properly :)) 

1. Marlin does flushes the rx buffer when an error occurs, the problem is with the 'state machine', when receiving commands it can receive commands with checksum or without, so one solution is to force to accept only commands with checksum when streaming to sd file.
 
The modifications are in a new branch (for the marlin firmware): https://github.com/adriancuzman/Marlin/commits/1.8_sd_only_CRC 

(I'll try to cleanup this and make a PR to Marlin, but I don't think they will accept the solution... )

In Octoprint I've forced "alwaysSendChecksum" to true just to test it. My main problem was with "_finishFileTransfer" executed before the transfer was completed (it is executed, but in the sendQueue there are a lot of commands to be sent, I've tried to execute some parts of the function when M29 was sent ( in "_gcode_M28_sent" ) but I couldn't make it work properly. If the alwaysSendChecksum is not set, then all the commands that are in the sendQueue after the finishFileTransfer is executed will be sent without checksum and they will be rejected in Marlin :)    Just moving self._changeState(self.STATE_OPERATIONAL) when M29 is actually sent from queue, didn't fix it, can't remember 100% but it will lock and not send additional commands from queue... 

Also, when alwaysSendChecksum is on, the weird logic "self._currentLine > 1" is not working anymore... from "isSendingFileToSDWithSoftwareFlow"... a workaround is when M28 is sent to set a flag so that in handle_ok to clear to send the send queue and ignore all other 'ok's when streaming... so one issue is to properly delimit when is streaming to sd and not... :/ 

2. Using these 2 branches I managed to make transfers 100% accurate (5x a file of 8.5Mbytes) from  laptop to the printer at 500k bauds. But not on the OrangePi (quad arm at 1.2Ghz, similar to Raspberry 3), to make it 100% accurate a speed of 57k bauds should be used, at 115k it's 2/3 files will not have a few bytes changed :)) {the 8.5Mbyte file}. This is because everything is executed on one core, at 250k it's 100% full, at 115k it's 60-70%, but it seems that is not enough to make it 100% stable.  Is it possible to use Process instead of Thread? So that it can be executed on different cores, maybe just receiving and sending, everything else should be fine on one core :)  Tried to rename threading with multiprocess, but it didn't worked :)) Here is the commit, maybe will help: https://github.com/adriancuzman/OctoPrint/commit/b829ebb290f18b1a7a56bb5a861e7edd8e1603a6

Also, in Marlin, had to use a 2k RX buffer and lower the limits (from RX_BUF/8 to RX_BUF/16 when sending the stop send char and from RX_BUF/10 to RX_BUF/32 when start send will be used) so that it's enough time for the XON/XOFF flow to stop. 

In the end, I will cleanup and a solution I see is that with alwaysSendChecksum on when the workaround for XON/XOFF is used {without it, the finishFileTransfer issue will be needed to be fixed, which will be more complicated and more code will need to be changed... }

So, any thoughts? :/

adriancuzman:
Update:

I've cleaned up the Marlin modifications, it's just one commit on this branch:
https://github.com/adriancuzman/Marlin/tree/force_checksum_on_sd_writing
(also this will be used for a new PR on Marlin)

Also I've cleaned up the modifications needed for Octoprint, which I pushed on this branch, I didn't squashed the commits just to be easier to review what's new.  Forced checksum when workaround is enabled, found a fix for currentLinenumber > 1 and resendLine + 1 'fix' is not needed :)

adriancuzman:
Update2: I was using pySerial 3.4 - updated the package and forgot about it :(
I'll redo the tests to see if it makes a difference.
Any particular reason why <2.8 is used? 3.4 worked fine :D

adriancuzman:
Update3: reverted to pySerial 2.7 and tested with a big file (8.5Mbytes), transferred 3 times without a single character changed. So the version of pySerial didn't make any difference :)

By the way I saw that there is an async pySerial... maybe this will fix the cpu usage problem..

adriancuzman:
In the weekend I've made some speed tests between different configurations (from laptop x86 and from OrangePi arm, using the USB->Serial or direct UART from OrangePI to serial 3 of the board, with XON/XOFF workaround Octoprint and with XON/XOFF enabled in the UART driver).

For me, the XON/XOFF enabled in the UART driver didn't work properly, I don't know how much buffering has the driver, but any transfer at 115k baud or more would fai because too many resend errors. Also a direct connection because of the wires (~15cm) will introduce more communication errors ( it won't work over 500k baud, but using the USB->serial chip, 2M baud will work)

From laptop the fastest it will work is when using a 2M baud connection through USB->Serial chip and XON/XOFF handled in Octoprint. A 548k file was transferred in 24 seconds and in 21 seconds if using the latest SdFat library (I made a PR with this but was rejected..) anyway I think this is the maximum practical speed for Atmega2560 boards.. witch is ~25kbytes/sec  (which is ok, an 8Mbyte file is transferred in ~5min, less than the hotbed is ready) 

From OrangePi, the fastest speed was with a 115.2k baud connection and using a RX buffer of 2k, the 548k file was transferred in ~62 seconds, which is practical for smaller files, but the problem is that I can't afford a 2k RX buffer, with it after calibration and bed leveling, I was down to less than 200 bytes of free memory and I'm not comfortable with this :))

Also the 2k RX buffer was also enabled when using the laptop, to make it without errors. But using the serialTest program that only transfers a file it will work with a 1k RX buffer.

In conclusion, using a direct serial connection is not an improvement, I'm back to using the USB to serial chip (also serial3 used 1 of the min stop pins, that was another downside :D ) And performance is really important in software flow control,  serial RX/TX should be on different threads on separate cores.. it would be very complicated to switch from threads to process for send/receive loops? I tried to change from thread to process and changed the event queues to multiprocess.queue, but it stalled and I can't figure it out why.

src/octoprint/util/comm.py
```diff
@@ -2608,11 +2669,13 @@ def _gcode_M25_queuing(self, cmd, cmd_type=None, gcode=None, subcode=None, *args
 			self.setPause(True)
 
 	def _gcode_M28_sent(self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs):
+		self._waitingForM28Confirmation = True
```
foosel: I fear a race condition here. Since sending and receiving are handled in separate threads, if the stars align just right the `_sent` handler might only be triggered after the monitor thread has already received and parsed the accompanying `ok`. Which means the `ok` for the `M28` won't trigger setting the flag, only the next ok will. Not sure if that is a problem in actual operation.

And no, there is sadly no way at the moment to match up sent commands with received oks (long story short: bug in several firmware variants out there makes commands not get acknowledged in the order they were sent in).

src/octoprint/util/comm.py
```diff
@@ -2017,6 +2061,17 @@ def _handle_errors(self, line):
 		# finally return the line
 		return line
 
+	def _filterXonXoffCharacters(self, ret):
```
foosel: Let's use `serial.XON` and `serial.XOFF` instead of redefining those constants.

src/octoprint/util/comm.py
```diff
@@ -2036,6 +2091,9 @@ def _readline(self):
 
 		if ret != "":
 			try:
+				if self.isSendingFileToSDWithSoftwareFlow() and self._use_xonxoff_workaround:
+					ret = self._filterXonXoffCharacters(ret)
```
foosel: In my tests I noticed that even when NOT using the workaround but native driver mode, I got the control chars in the stream, making it necessary to always filter. Can you confirm or deny that?

src/octoprint/util/comm.py
```diff
@@ -583,6 +600,21 @@ def getStateString(self, state=None):
 			return "Transferring file to SD"
 		return "Unknown State (%d)" % (self._state)
 
+	def isSendingFileToSDWithSoftwareFlow(self):
+		#first determine if sending file to sd and printer supports xon_xoff
+		if 	self._startedSendingFileToSD and \
+			self._firmware_capabilities.get(self.CAPABILITY_SERIAL_XON_XOFF, False):
+				# currentLine > 1 is to enable flow control only and ignore ok only after the first ok is sent
```
foosel: This comment is now outdated

src/octoprint/util/comm.py
```diff
@@ -387,7 +388,21 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
 		self._currentLine = 1
 		self._line_mutex = threading.RLock()
 		self._resendDelta = None
-		self._lastLines = deque([], 50)
+
+		self._use_xonxoff_workaround = settings().get(["serial", "useSoftwareFlowWorkaround"])
+		self._xonxoff_stopChar = chr(19)
+		self._xonxoff_resumeChar = chr(17)
+		self._waitingForM28Confirmation = False
```
foosel: `self._waiting_for_m28_ok`

src/octoprint/util/comm.py
```diff
@@ -387,7 +388,21 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
...
+		self._startedSendingFileToSD = False
```
foosel: Can you explain what this flag is accomplishing? I understand it's your solution to the line number test, but I'd like to understand your reasoning here and figure out if there is a better approach. I'm not too happy about introducing so many one shot variables here for one feature, and this is the most puzzling of all.

Why is just using the state like your previously did not working here?

Also, if kept this should be changed to snake_case instead of camelCase. Yes, lot's of old vars in there still in camelCase, but I'm slowly migrating that and am also trying not to introduce new ones.

adriancuzman:
First thank you for responding, it's ok, I understand, a project like this takes a lot of your time... please, it is not necessary to apologize for a late response :)

So, this PR is turning more in a research/try to make flow control work. But in the end I think it will be more to learn on how to implement it in the future. 

**These are me main problems that are stopping flow control to work:**
- the time needed for the host to stop sending data when a stop char is received
- the marlin firmaware accepting mixed lines with checksum and without in the same time
- the comm layer logic that waits for an ok to send another line and change to send more lines without ok confirmation (this is the reason why I use a lot of bool switches to get around the logic, but still use the clear_to_send sync logic to stop sending lines as soon as possible)

other problems: 
 - the linux driver doesn't handle XON/XOFF flow control (for the USB chips) this is why it's needed to be treated in the application level (if it's enabled, no error will be generated ('not implemented would be nice'), it will filter the chars, but will not stop the flow.. it's a silent bug/feature :D ) 
 - the UART driver implements XON/XOFF flow control, but I think when sending a lot of data the buffers in the driver get full and something goes wrong and data is lost, normally the driver should block the in the write function when the buffers get full, but I'm not sure if this happens.. but from my testing, it seems that having flow control at application level is better and the performance impact is very small, almost none measurable. 

Also I'm not using this in the end.. so if I can't use it, it's not ready for other people too :))

**I'll try to summarize what is minimum needed to make it work:**

  For Octoprint:
         - when sending a file on SD, disable the waiting for 'ok' responses and ignore them
         - in Octoprint filter XON/XOFF chars and stop sending lines
        (these two from above are in this PR + enabling XON/XOFF flow in the serial driver)

  In the printer firmware (Marlin):
         - accept only lines with checksum when saving a file to SD

  To make it work for everyone and not as an experiment it will be needed:
          - to offload the comm layer logic from running on the same core with everything else
          - the comm layer logic to support sending data and without waiting for ok confirmations

So it can work, but in certain conditions...

Best scenario is:  Octoprint running on a PC with a CPU speed over 3Ghz per core and workaround switch set to True (from this branch) and Marlin set with a 2k RX buffer, XON/XOFF enabled and the fix to force CRC checking of all the lines when saving to SD (this commit https://github.com/adriancuzman/Marlin/commit/8f1b3ba6f4570bed29399f2211e2f1b1476d15a5) With this I'm getting 21-23kbytes/sec writing speed on SD. The baud rate was 500k, it should work on lower baud rates. Also I've tested with the small serial test program in python (the https://github.com/adriancuzman/marlinSendFileTest ), with this the RX buffer in Marlin can be 1024bytes (maybe it will work with 512byte, but I haven't tested it). This indicates how slowing down the response from the host influences the need for a bigger RX buffer on the printer (a small program that is just reading a file and sending to serial vs a program that does this and more on the same core)

Octoprint running on ARM (a 1.2Ghz cpu from orangepi) a 2k buffer in Marlin is need to reliably send data at only 115k baud (reliably - send at least 3 times a BIG.g file from the malinSendFileTest repository).

In the end I think this solution for now it's not ready. At first when I discovered the sd speed issue I thought it should be a quick fix, 1-2 weeks max... but after 3 months, I've changed my mind that this is not a 'quick fix' :))  The good think is that it can work, the bad thing is that in only certain conditions.. as I've said, I'm not using it (first of all, I can't afford a 2k RX buffer on my board, 1k is maximum and what's the point of having a full pc/laptop to replace the job of a raspberry? :)) )  

I think it's ok to close this PR, the workaround code it's turning into something I don't like too :))

src/octoprint/util/comm.py
```diff
@@ -1609,7 +1642,14 @@ def convert_line(line):
 		self._log("Connection closed, closing down monitor")
 
 	def _handle_ok(self):
-		self._clear_to_send.set()
+		can_send = not self.isSendingFileToSDWithSoftwareFlow()
+		if self._waitingForM28Confirmation:
+			self.__waitingForM28Confirmation = False
```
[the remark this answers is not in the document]
adriancuzman: yes it is... oops :))

src/octoprint/util/comm.py
adriancuzman: thank you, I didn't knew they where defined in serial..

src/octoprint/util/comm.py
adriancuzman: if in the firmware XON/XOFF is enabled, whenever the buffer is at 12% full it will send the chars, blindly, there is no handshake to confirm the host will interpret the chars. So if flow control is enabled in the firmware and the workaround is "not" active (the filter function) or flow control is "not" enabled from the driver, the chars will be delivered to the application level...

src/octoprint/util/comm.py
adriancuzman: yes, it's not a good solution to use switches like this... I try to explain more in the comment where the variable is defined.

src/octoprint/util/comm.py
adriancuzman: At first I tried using the state, but it locked. I'll try to explain: the send_loop blocks until a clear to send it's sent, in handle_ok, if I ignore the ok handling using the state will not send a clear_to_send needed to unblock send_loop. The idea is that, the ok_handle should ignore handling the ok only after the confirmation of M28 is received. The state does not catch this... that's why I use another switch :/ (the  _waitingForM28Confirmation). Another problem is that the state also is off when the file is actually finished. When it's waiting for ok confirmation, the send queue, will be filled with 1 item (line). But when the workaround is active, the send queue will be filled with more lines, there could be  hundreds of lines in the queue and the state is changed when all the lines from the file are set in the queue, but not removed. So for big files, the progress bar will be 100% done, the popup will show the transfer is finished, and the state is changed, but the file is still transferring (what's left in the send queue). So I needed another switch to detect when the file is actually finished (and this is when M29 is 'actually' sent). And this switch _startedSendingFileToSD  is used to actually catch this and to filter the flow control, if I use the state, what's left in the queue will not be filtered properly and it could lead to issues (the printer sends a stop and it will be ignored). 

I hope I've explained it well and will help in the future when updating the comm layer :)

fiveangle:
@ejtagle - did you ever find a way around the upstream Linux serial driver XON/XOFF issues ?

ejtagle:
This goes extremely slow... The only viable and deterministic implementation is kernel-side:

foosel:
Since @adriancuzman themselves said that this solution isn't ready yet due to too many limiting factors in firmware, OS and OctoPrint, closing this. I could leave it on hold for a long time now, but that will only make it become stale. I would have preferred it otherwise, but if it doesn't work reliably with the current constraints, it's simply not merge-able 😕

DanielJoyce:
Python has pretty good support for simple process forking. Overhead can be avoided via using shared_memory or even a local socket to communicate.

It doesn't surprise me that XON/XOFF support is spotty for things that aren't true serial ports. Usually the intention is it to be handled at the software layer.

Another problem is the CH430/CH431 chips used on cheap printers have excedingly super tiny buffers.

DanielJoyce:
I think one danger is some driver's handle xon xoff as a nicety and others don't. So I suspect you should probably turn off that handling at the try level and then handle it all in your software stack because that's always what xon xoff implied, you saying you'll do it yourself. And doing it properly requires scanning the buffer for xon/xoff and doing it all yourself 

I guess the uarts in printers don't properly support hardware flow control? Should be possible to stream a file to sd card and use hardware flow control.

PR #2385, opened 2018-01-26
API: add 'True' as a string as a boolean

tedder:
API: add 'True' as a string as a boolean

Easy to test:
curl -H "X-Api-Key: x" http://octopi.local/api/printer?history=true | grep history
curl -H "X-Api-Key: x" http://octopi.local/api/printer?history=True | grep history

The second test previously failed, with this code it succeeds.

This allows pythonic API clients to pass boolean values in:

requests.get('/api/printer', params={'history': True})

Previously this didn't work, as Octoprint doesn't see True as a boolean, it gets a sting. So it never matches True, it only matches "True".

na

na

#### Further notes
Hope you get feeling better, Gina.

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -405,7 +405,7 @@ default_settings = {
 }
 """The default settings of the core application."""
 
-valid_boolean_trues = [True, "true", "yes", "y", "1"]
+valid_boolean_trues = [True, "True", "true", "yes", "y", "1"]
 """ Values that are considered to be equivalent to the boolean ``True`` value, used for type conversion in various places."""
 
 


foosel:
Hey, thanks for that, good idea! Since I thought it would make sense to have this in the next maintenance release I cherry picked your commit to the `maintenance` branch as b3b19dc136fa5708d9978c44a4a778083e103971. Then I thought it maybe might make sense to have `valid_boolean_trues` be case insensitive in general and introduced `CaseInsensitiveSet` to take care of that in the future (see ec12bd7cd410378059a495e8819d2de25b02032c) so from now on also `TrUE`, `trUe` and so on will work ;)

Since Github doesn't detect that I cherry picked your commit, I'll have to close this PR manually. Rest assured though that your change *has* been merged and in fact been expanded further!

PS: Feeling better, thanks :)

Issue #1600, opened 2016-11-25
[Request] display long filenames from SD card - use M33 to get long filenames from printer


doughadfield:
For printers that support returning the full filename of SD card files (Marlin) via the M33 command, please can the full filename be displayed in the Octoprint file list. Octoprint uses the M20 command to return a list of files in the SD card directory. These are returned in Microsoft short form, and are usually almost completely un-human-readable. The M33 command can then be used to obtain the full filename, but this needs to be iterated for each file name separately. Please can Octoprint be enhanced to do this after issuing the M20 command, and display the full filenames on the web page.

Thank you
Doug

foosel:
Sadly the M33 command is pretty much useless for that. The problem is that I'd have to call it once for every single file (including those in sub folders and sub sub folders...) on the SD. Imagine a SD card with only 20 files and you are already looking at 21 commands to be able to populate the file list. That simply doesn't scale at all.

There was a discussion a while back on the marlin firmware bug tracker where it was talked about enhancing the M20 output to directly include the long file names in the response but I think that didn't end in any firmware implementation.

Additionally the long file names are not guaranteed to be ASCII, whereas the communication over serial is actually supposed to be only ASCII (even though a lot of firmware variants out there simply don't care). So the long file names would also need to be transported in some encoded format.

I'd be happy to add long file names from the sd to OctoPrint once there is an *efficient* way to retrieve that kind of information. But sorry I refuse to implement (or merge) a solution utilising M33, which would effectively nuke the whole printer communication for an indeterminate amount of time (depending on the number of files on the card) every time the file list needs to be refreshed. This will just lead to people opening tickets because "their printer locked up" because they had hundreds of files on the attached sd card.

Richie76:
would it be possible to get the long filename just for the one file printed ? so it can be used for the filename of the timelapse ?

And sorting of the sd-card files in newst first order would also be quite helpful (at least for me: I´m printing from and flashair sdcard and using octoprint only for controlling and timelapse due to the slow serial interface of my printer and the artifacts this cuases).

tia, rgds, richard

drdelaney:
What about getting the long filename for the file currently loaded / printing file?

Andrei-Pozolotin:
perhaps M33 could be made salable by lazy running it in background along with periodic M105 polls
this way, you show short names first, and long names will "eventually" show up

MStylesMS:
[...]

PR #2407, opened 2018-02-06
Display long filenames on SD card with Marlin 1.1

cbxbiker61:

This is a basic approach to adding long filenames support with Marlin 1.1.  Lazy lookup could be implemented quite easily by looking at the sdFileLongname for None periodically until all sdFileLongname's are populated.  As far as ASCII goes I don't know why we can't just filter the M33 response.

I feel that this support is going to become more important as printers start to support support power fail recovery as well as missed steps recovery (TMC2130 Einsy board).  These new features are going to require printing from SD card.

It was tested with a Marlin 1.1.8 and and Einsy board.  With 25 files it is plenty fast.

https://github.com/foosel/OctoPrint/issues/1600

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -87,6 +87,7 @@ date of first contribution):
   * [Peter Backx](https://github.com/pbackx)
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
+  * ["Kelly Anderson"](https://github.com/cbxbiker61)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -582,7 +582,7 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 	def get_sd_files(self):
 		if self._comm is None or not self._comm.isSdReady():
 			return []
-		return map(lambda x: (x[0][1:], x[1]), self._comm.getSdFiles())
+		return map(lambda x: (x[0][1:], x[1], x[2]), self._comm.getSdFiles())
 
 	def add_sd_file(self, filename, absolutePath, on_success=None, on_failure=None):
 		if not self._comm or self._comm.isBusy() or not self._comm.isSdReady():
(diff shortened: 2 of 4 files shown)


foosel:
Sorry, but I don't want to merge this.

As explained in [this comment](https://github.com/foosel/OctoPrint/issues/1600#issuecomment-262955148) on #1600, iterating over the whole file list returned by `M20` and calling `M33` for each entry simply does not scale. The host has no control over the number of files on the printer's SD and hence the connection to the printer might be saturated with `M33` calls for a while and nothing else in terms of command and control will be possible during that time. And if something goes wrong in some way during that long exchange, the SD listing command will become stuck for good since the counter won't reach 0. The cost usefulness ratio here is simply skewed.

The `M33` command is nice and all for getting the long name of *one* file, but it simply is entirely unsuited for getting a full list of long file names from the firmware at scale. Now, if the long names were to be included in the output of `M20`, that would be something that might work (encoding issues aside). But as far as I know that was never implemented despite long discussions about it.

How about this: Let's add a hook instead that allows expanding the SD file list through plugins (and make the entries of that be a proper data structure with something like `name`, `size`, `displayName`). Then extract this functionality into a plugin. If people want to wait for their printer connection to become responsive until all long names for the files on the SD have been fetched, they can install it. If not they don't have this forced on them.

cbxbiker61:
I don't disagree with you.  The plugin option is probably the way to go.  I just wanted to get something tangible in the works.  I may talk to thinkyhead about extending M20.  On the surface it doesn't seem like it would be too terribly hard to get M20 to return the short and long filenames.  As well as adding an identifier at startup to know the feature is available.

foosel:
I sadly can't remember the ticket number anymore, but we had a lengthy discussion about this in the marlin tracker about 1.5 to 2 years ago. Maybe it originated in the "let's add size information to `M20`"? I'm unsure. In any case I think no consense was achieved. And what's still an entirely onsolved issue is the whole encoding situation - this stuff is supposed to be ASCII, but I think back then it was stated that `M33` will return whatever encoding was used for the long file name, which could be anything really.

Grogyan:
It would be ideal to have it as a plugin.

feldon30:
A local cache would do quite nicely here. Cache would be a text file or sqlite with each 8 char filename + timestamp and then the full filename. Expire every cache record after 30 or 60 days.

PR #2410, opened 2018-02-08
[Brainstorming] Support both mpeg2 and h264 for timelapse rendering

koenkooi:

This PR adds a setting to use h264 as videocodec for timelapse rendering. This should give a quality boost at the same bitrate setting or allow to drop the bitrate while keeping the same quality.

I tagged it as 'RFC' because there are at least these issues still left:
- [x] ~~Spurious whitespace changes~~
- [x] ~~Missing translation strings~~
- [x] ~~Commit ordering, settings are in before the code~~
- [x] ~~Fix and extend related unit tests~~

Before I spend more time on this, is this the way to go, or should any extra changes to timelapse by done in an octopi style external script or as a module?

Tested by setting _Settings → Webcam & Timelapse → Timelapse Recordings → Advanced settings → Videocodec_ to 'libx264' and starting a print.
See screenshot section for details on the resulting timelapse.

See above, h264 support for timelapses.

https://github.com/foosel/OctoPrint/issues/2388
https://github.com/foosel/OctoPrint/issues/2373
https://github.com/foosel/OctoPrint/issues/1980

schermafbeelding 2018-02-08 om 15 54 53
schermafbeelding 2018-02-08 om 15 58 17

### DIFF
--- a/docs/api/settings.rst
+++ b/docs/api/settings.rst
@@ -254,6 +254,8 @@ mapped from the same fields in ``config.yaml`` unless otherwise noted:
      -
    * - ``webcam.ffmpegThreads``
      -
+   * - ``webcam.ffmpegVideoCodec``
+     -
    * - ``webcam.watermark``
      -
    * - ``webcam.flipH``
--- a/docs/configuration/config_yaml.rst
+++ b/docs/configuration/config_yaml.rst
@@ -1194,6 +1194,9 @@ Use the following settings to configure webcam support:
      # Should be left at 1 for RPi1.
      ffmpegThreads: 1
 
+     # Videocodec to be used for encoding. Defaults to mpeg2video.
+     ffmpegVideoCodec: mpeg2video
+
      # The bitrate to use for rendering the timelapse video. This gets directly passed to ffmpeg.
      bitrate: 5000k
 
--- a/src/octoprint/server/api/settings.py
+++ b/src/octoprint/server/api/settings.py
@@ -108,6 +108,7 @@ def getSettings():
 			"ffmpegPath": s.get(["webcam", "ffmpeg"]),
 			"bitrate": s.get(["webcam", "bitrate"]),
 			"ffmpegThreads": s.get(["webcam", "ffmpegThreads"]),
+			"ffmpegVideoCodec": s.get(["webcam", "ffmpegVideoCodec"]),
 			"watermark": s.getBoolean(["webcam", "watermark"]),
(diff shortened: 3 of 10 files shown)


foosel:
What was your experience with rendering performance on something like a Pi with this? The last time I tried something like this it pretty much brought the poor guy to a grinding halt, that was three years ago or so though.

And this also has an additional TODO ;)

* [ ] Fix and extend related unit tests

koenkooi:
I didn't notice any difference between mpeg2 and h264 when setting 'threads' to '0' on an rpi2 and rpi0-w. No noticable hickups while printing over serial during rendering. I haven't compared rendering times, which I guess is another TODO :)

phreaker0:
@koenkooi Thanks for implementing this, this saves me from reencoding and I now can get my time lapse video faster over my slow wan uplink. I use it on a RPI3 with only one thread, it takes much longer to reencode than with mpeg2. Printing while encoding works fine for me as expected with the one thread limit.

funkycode:
What about container? to have matroska instead of mpeg?

koenkooi:
I've left out container choice in this PR because I felt it might be too much options to add in one go. You can easily patch it locally, octoprint will allow you to download them since #2409 was merged.

koenkooi:
Sweet, the unit tests caught a bug. Here's the change I squashed in:

```
koen@beast:/build/pkg/OctoPrint$ git show dfc3c34
commit dfc3c3406612da1644fa1ff12336297466945139
Author: Koen Kooi <koen@dominion.thruhere.net>
Date:   Sat May 12 17:33:49 2018 +0200

    SQUASH ME: timelapse.py: python 'is' is not a string compare, use ==
    
    Signed-off-by: Koen Kooi <koen@dominion.thruhere.net>

diff --git a/src/octoprint/timelapse.py b/src/octoprint/timelapse.py
index 6d27202..3c0c9f1 100644
--- a/src/octoprint/timelapse.py
+++ b/src/octoprint/timelapse.py
@@ -941,7 +941,7 @@ class TimelapseRenderJob(object):
                logger = logging.getLogger(__name__)
 
                 ### Not all players can handle non-mpeg2 in VOB format
-                if videocodec is 'mpeg2video':
+                if videocodec == "mpeg2video":
                     containerformat = "vob"
                 else:
                     containerformat = "mp4"
```

koenkooi:
@foosel this is ready for proper review now, I've kept the commits seperated to make review easier (IMO), but I think the end result should use the squash-and-merge button to keep the unit tests working for every commit in the upstream tree.

PR #2440, opened 2018-02-22
Added a new "customControls_controlTemplate_command"

electr0sheep:
* [?] Your changes are not possible to do through a plugin and relevant
    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

----

This PR adds the ability to specify custom button classes (btn-primary, btn-danger, etc) to custom commands. This change is necessary because without it, the styling functionality of this plugin will not work: https://github.com/Salandora/octoprint-customControl

The changes in the PR are not achievable by a plugin.

I didn't use unit tests, however I tested it with a command that didn't have the customClass property and the class was set as expected **class=btn**. I also tested it with the plugin referenced above for commands with customClass defined

I don't know if adding custom classes to controls are a highly sought after feature, but I wanted to add a control that applied heat to the printer bed and nozzle, and I wanted to make it red. So, it would be useful to me!

N/A

screen shot 2018-02-22 at 12 27 55 am

#### Further notes
I am new to OctoPrint, as well as Knockout, but my intentions were to check for the customClass property and if it didn't exist, to style the button with the same styling before my changes. This way, my code shouldn't conflict with any other code/plugins that don't use custom classes.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -88,6 +88,7 @@ date of first contribution):
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
   * [Bernd Zeimetz](https://github.com/bzed)
+  * [electr0sheep](https://github.com/electr0sheep)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/templates/tabs/control.jinja2
+++ b/src/octoprint/templates/tabs/control.jinja2
@@ -154,7 +154,8 @@
 <script type="text/html" id="customControls_controlTemplate">
     <form class="form-inline custom_control">
         <!-- ko template: { name: 'customControls_controlTemplate_input', data: $data, if: $data.hasOwnProperty('input') } --><!-- /ko -->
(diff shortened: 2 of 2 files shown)


foosel:
Hm... Instead of adding another template subclass, I'd rather see the current one's template expanded such that it can inject custom (additional?) classes.

I'm also a bit confused as to why that is necessary to get that plugin to work, I thought that was working for a while already including custom colors. @Salandora could you maybe elaborate on this?

In any case, sounds like an improvement of an existing feature, so doing this against `maintenance` would also make more sense I think.

electr0sheep:
> Instead of adding another template subclass, I'd rather see the current one's template expanded

Like I said, I'm pretty new to all of this, but I agree. I'll see if I can't expand the current template instead.

> I'm also a bit confused as to why that is necessary to get that plugin to work, I thought that was working for a while already including custom colors.

I don't think the custom colors did work. According to my information, the drop-down menu that you choose the style from references a function "editStyle":

https://github.com/Salandora/octoprint-customControl/blob/b684644bbe695e6253e41a2802bbc5dcb2a15f90/octoprint_customControl/templates/customControl_hookedsettings.jinja2#L14-L32

which wasn't implemented:

https://github.com/Salandora/octoprint-customControl/blob/b684644bbe695e6253e41a2802bbc5dcb2a15f90/octoprint_customControl/static/js/customControl.js#L556-L557


> doing this against maintenance would also make more sense I think

No problem, I can fix your first suggestion and then submit another PR against maintenance

foosel:
Huh, okay, then I guess my memory is mistaken.

And sounds good!

Salandora:
I never came around to finish the styling part. But I thought I had already implemented the necessary changes to the core... seems like I did not :-/

PR #2444, opened 2018-02-24
Custom classes for custom controls

electr0sheep:
* [?] Your changes are not possible to do through a plugin and relevant
    Brainstorming ticket
    e.g. dev/my_new_feature
    lessc)
    have added unit tests
    nothing broke

----

This PR adds the ability to specify custom button classes (btn-primary, btn-danger, etc) to custom commands. This change is necessary because without it, the styling functionality of this plugin will not work: https://github.com/Salandora/octoprint-customControl

The changes in the PR are not achievable by a plugin.

I didn't use unit tests, however I tested it with a command that didn't have the customClass property and the class was set as expected **class=btn**. I also tested it with the plugin referenced above for commands with customClass defined

I don't know if adding custom classes to controls are a highly sought after feature, but I wanted to add a control that applied heat to the printer bed and nozzle, and I wanted to make it red. So, it would be useful to me!

N/A

screen shot 2018-02-22 at 12 27 55 am

#### Further notes
I am new to OctoPrint, as well as Knockout, but my intentions were to check for the customClass property and if it didn't exist, to style the button with the same styling before my changes. This way, my code shouldn't conflict with any other code/plugins that don't use custom classes.

This PR reflects the conversation with @foosel

### DIFF
--- a/src/octoprint/templates/tabs/control.jinja2
+++ b/src/octoprint/templates/tabs/control.jinja2
@@ -173,6 +173,11 @@
     <label style="cursor: default" data-bind="text: output"></label>
 </script>
 <script type="text/html" id="customControls_controlTemplate_command">
-    <button class="btn" data-bind="text: name, enable: $root.isCustomEnabled($data), click: function() { $root.clickCustom($data) }"></button>
+    <!-- ko if: $data.customClass -->
+        <button data-bind="text: name, attr: { 'class': customClass }, enable: $root.isCustomEnabled($data), click: function() { $root.clickCustom($data) }"></button>
+    <!-- /ko -->
+    <!-- ko ifnot: $data.customClass -->
+        <button class="btn" data-bind="text: name, enable: $root.isCustomEnabled($data), click: function() { $root.clickCustom($data) }"></button>
+    <!-- /ko -->
 </script>
 <!-- End of templates for custom controls -->


foosel:
I've thought about this some more, and I don't like being able to completely redefine the class. What I'd rather like to see is having the option to define *additional* classes to apply (that also matches how for example plugins allow styling of injected templates today). That would reduce the likelihood of a user forgetting to include `btn` next to `btn-danger` and completely breaking their control (and also reduce the overhead of always having to include `btn` anyhow).

Just something like:

``` html
    <button class="btn" data-bind="text: name, enable: $root.isCustomEnabled($data), click: function() { $root.clickCustom($data) }, css: additionalClasses"></button>
```

and having `additionalClasses` default to an empty string. 

As I understand that would require to also adjust the custom control plugin but I just saw you had closed your PR there?

Salandora:
@foosel the closing of the PR was related to a) wrong branch, b) multiple things in one PR.
electr0sheep offered to do a new PR against the devel branch when this PR here is solved, because it is necessary for the changes that needs to be done to the custom control editor plugin

foosel:
Slightly adjusted as mentioned above:

image

Also updated the docs accordingly. Will be part of 1.3.7.

PR #2502, opened 2018-03-21
Fix some invalid HTML

dforsi:
Fixes 4 kinds of errors found with https://validator.w3.org/nu/#textarea
> Bad value javascript for attribute lang on element script
> Bad character after < . Probable cause: Unescaped < . Try escaping it as &amp;lt; .
> The element label must not appear as a descendant of the label element.
> Element legend not allowed as child of element form in this context.

Note: the change related to &lt;legend&gt; seems big because of the increased
indent of the existing code (use "git diff --ignore-all-space" or
"git log --patch --ignore-all-space" to see the lines actually added).

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR fixes some errors that make the HTML invalid, but OctoPrint works and the page is rendered in the same way even without this patch.

Tested comparing the page before and after the patch and running the HTML again through the validator.

A page without errors is more likely to be rendered as expected in more browsers and any regression would be obvious.

#### Further notes

There are still some errors to fix, for which I'm going to open an issue.


foosel:
Thank you!

PR #2530, opened 2018-03-29
fix broken yaml in config docs

tedder:
just a colon


foosel:
Can you do that against `maintenance` instead of `devel` since it's a fix?

foosel:
#2554 was merged.

PR #2533, opened 2018-03-31
Fix buttons not disabled when there isn't an API key available

dforsi:
Bootstrap buttons made from HTML `<a>` tags can only be disabled with CSS.

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

The buttons to copy the API key to the clipboard and to delete the user key aren't disabled when the input box shows "N/A" because the knockout.js code attached to them uses the `disabled` attribute which works only for `<button>` and `<input>` elements, not for '<a>'.

To test the user key, delete it with the "trash" button and see if the buttons change appearance.
To test the global key I edited the HTML by hand.

I would have preferred to change only the HTML from `<a>` to`<button>` but this breaks the layout in many ways.

* Before
image

* After
image

* Before
image

* After
image

#### Further notes

This bug was present in 1.3.6.

Note that I'm only fixing the visual aspect with this PR, but I'd rather change the behavior by keeping always enabled the button to regenerate the global API key, just like it is done for the user key.
The last image shows that if this PR is merged, it wouldn't be possible to manually generate a new global API key if one is missing; I understand that OctoPrint will create a missing key on start, but this user interface doesn't need to depend on that.

### DIFF
--- a/src/octoprint/templates/dialogs/settings/api.jinja2
+++ b/src/octoprint/templates/dialogs/settings/api.jinja2
@@ -18,8 +18,8 @@
         <div class="controls">
             <div class="input-append input-block-level">
                 <input type="text" readonly="readonly" id="settings-apikey" data-bind="value: api_key, attr: {placeholder: '{{ _('N/A') }}'}">
-                <a class="btn add-on" title="Copy API Key to clipboard" data-bind="click: copyApiKey, enable: api_key"><i class="fa fa-copy"></i></a>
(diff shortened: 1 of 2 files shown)


foosel:
Hm... I honestly think that limiting global API key generation availability to there already being one is a bug. As you correctly point out, it's also inconsistent with the user API key. Could you extend this to simply remove the "enabled" dependency on the API key presence? Feels more right that way. Or am I missing something here?

dforsi:
> Could you extend this to simply remove the "enabled" dependency on the API key presence?

done

foosel:
👍 😊

PR #2535, opened 2018-03-31
Pip install with --no-cache-dir to handle plugin installations with M…

ganey:

On older RasperryPi's various plugins will not install with pip and fail with 'MemoryError'.
This occurs particularly with plugins that use `Pillow`

Via current unit tests for breakages.

https://github.com/FormerLurker/Octolapse/issues/107

#### Further notes

Please see ticket for logs detailing previous installation issues. 

Happy to discuss this further, Python isn't my strongest language, so happy to edit this if required. So far haven't had any issues running a fork on AnetA8 running marlin with this PR.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -88,6 +88,7 @@ date of first contribution):
   * [Josh Major](https://github.com/astateofblank)
   * ["alex-gh"](https://github.com/alex-gh)
   * [Bernd Zeimetz](https://github.com/bzed)
+  * [Ganey](https://github.com/ganey)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/plugins/pluginmanager/__init__.py
+++ b/src/octoprint/plugins/pluginmanager/__init__.py
@@ -333,7 +333,7 @@ class PluginManagerPlugin(octoprint.plugin.SimpleApiPlugin,
 			raise ValueError("Either URL or path must be provided")
 
 		self._logger.info("Installing plugin from {}".format(source))
-		pip_args = ["install", sarge.shell_quote(source)]
+		pip_args = ["install", sarge.shell_quote(source), '--no-cache-dir']
 
 		if dependency_links or self._settings.get_boolean(["dependency_links"]):
 			pip_args.append("--process-dependency-links")


foosel:
I'm not entirely sure *why* this is fixing these issues for you... In the ticket you linked you said

> So I tried this which usually helps with other libraries on older RaspberryPi's lacking in memory.

Do you have any kind of information as to *why* that "usually helps"? As far as I understand `--no-cache-dir` it merely disables the local package cache, e.g. if you already downloaded Pillow from PyPI in the past and that download is stored in `~/.cache/pip` it will then be fetched from there instead of doing another network request to fetch the package from PyPI's servers. I don't see how not doing that could prevent an older system running into a `MemoryError` 🤔

ganey:
It seems to do more than just not use the cache dir, 

It really confused me at first too, but it seems to prevent loading the package into memory?

Initial googling for 'pip install' that results in a MemoryError gives this: [pip install matplotlib](https://stackoverflow.com/questions/29466663/memory-error-while-using-pip-install-matplotlib)

Other random suggestings that this works for lower memory devices:
https://chirale.org/2017/01/15/memory-error-on-pip-install-solved/
http://codelectron.com/how-to-solve-memoryerror-during-pip-install/

Based on this result though https://github.com/pypa/pipenv/issues/451 it could be possible to make this env based and just document that lower memory devices should set `--no-cache-dir` as per the [docs](https://pip.pypa.io/en/stable/user_guide/#configuration)? 

As I mentioned previously I'm not too strong in python. Is there a way for me to run the OctoPi tests with say around 128Mb of memory? Then I could try and test this as a pip config value instead.

foosel:
> This error is coming up because, it seems, pip's caching mechanism is trying to read the entire file into memory before caching it… which poses a problem in a limited-memory environment, as matplotlib is ~50mb.

Indeed, *that* explains a lot. Actually the same reason I put some serious work into upload streaming and caching in OctoPrint to avoid this kind of problem with 200MB GCODE files being uploaded.

Ok... so... that is reason enough to make this mandatory for now. Long term we can detect the pip version and switch it off if not needed, but since pypa/pip#2984 is still open I guess it's still present in current versions (and in fact an upstream dependency issue if I understand correctly).

Can you do that against the `maintenance` branch though? Makes sense to get this into 1.3.8 (too late for 1.3.7 sadly).

foosel:
Closing since #2543 replaces it against `maintenance` instead of `devel`

pramodkhandalkar:
pip install --no-cache-dir pyspark
Use  --no-cache-dir  Before library name 
it will directly install and remove the cache after installation

PR #2540, opened 2018-04-02
Add missing documentation for GCODE analysis and other small fixes

### LINKED ISSUE #2463: [Request] Add documentation for dimensions and printingarea in gcodeAnalysis data model
`GET /api/files/[path+file]` returns the `gcodeAnalysis`, as documented in [file_information:files of the common data model](http://docs.octoprint.org/en/master/api/datamodel.html#files). The [gcodeAnalysis](http://docs.octoprint.org/en/master/api/datamodel.html#sec-api-datamodel-files-gcodeanalysis) returns filament info and estimated print time. But it also returns the `printingarea` dict and the `dimensions` dict, neither of which appear to be documented anywhere.

It looks like [foosel added it a year ago](https://github.com/foosel/OctoPrint/commit/639d8d4c5f8432679793431d11461fdfb0531736)- but it isn't really documented. 

(not filling out the issues template, it doesn't apply)

dforsi:
Adds documentation for `dimensions` and `printingArea`; fixes `length`
and `volume`; adds two internal references.
Fixes #2463.

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

It adds some missing information.

> make -C docs html

#2463 

image

#### Further notes
Now the structure of this JSON object is too complex to be rendered in a simple flat table. I'm open to ideas.
I don't think that it is easy to understand that the table above can describe a fragment like this (I may have made a mistake):
````
filament: {tool0: {length: 1407.434510000002, volume: 8.978582902758397}}
````

### DIFF
--- a/docs/api/datamodel.rst
+++ b/docs/api/datamodel.rst
@@ -263,8 +263,11 @@ File information
      - Path to type of file in extension tree. E.g. ``["model", "stl"]`` for ``.stl`` files, or ``["machinecode", "gcode"]``
        for ``.gcode`` files. ``["folder"]`` for folders.
 
-Additional properties depend on ``type``. For a ``type`` value of ``folder``, see "Folders". For any other value
-see "Files".
+Additional properties depend on ``type``.
+For a ``type`` value of ``folder``, see :ref:`Folders <sec-api-datamodel-files-folders>`.
+For any other value see :ref:`Files <sec-api-datamodel-files-files>`.
+
+.. _sec-api-datamodel-files-folders:
 
 Folders
 '''''''
@@ -286,6 +289,8 @@ Folders
      - 0..1
      - Number
      - The size of all files contained in the folder and its subfolders. Not present in non recursive listings!
+
+.. _sec-api-datamodel-files-files:
 
 Files
 '''''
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/api/datamodel.rst [hunk 1]
```diff
@@ -384,14 +389,58 @@ GCODE analysis information
      - 0..1
      - Object
      - The estimated usage of filament
-   * - ``filament.length``
+   * - ``tool{n}.filament.length``
```
foosel: Shouldn't this rather be ``filament.tool{n}.length``? Same with ``filament.tool{n}.volume`` below.

foosel:
I think a flat table should still suffice, as long as the above issue is fixed. 

I would expect people to also take a look at live data which should clarify things entirely. Or we could just include an example?

dforsi:
You were right, I fixed the two `filament.tool{n}...` lines and forced a push.

Including an example would be useful, but only if you can add them in a form that can be tested, such as using unit tests as examples (maybe pulling the code in the .rst file with `literalinclude`), or deriving the json output from running actual code (maybe with `doctest`), so that only the tables are to be maintained manually, but I never used those directives. In some cases it would be better for the reader to see only some fields of a big object, however appending to each chapter an output with all possible fields would be easier to maintain, if you can derive it form the code.

PR #2543, opened 2018-04-03
Pip install with --no-cache-dir to handle plugin installations with MemoryError

ganey:

Fixes installations of pip plugins which throw a MemoryError on low memory devices.

See https://github.com/foosel/OctoPrint/pull/2535 for more info

> This error is coming up because, it seems, pip's caching mechanism is trying to read the entire file into memory before caching it… which poses a problem in a limited-memory environment, as matplotlib is ~50mb.


foosel:
Hah, thanks, I would have merged it manually, but that of course just made my life way easier :)

PR #2554, opened 2018-04-05
docfix: make 'settings' proper yaml

tedder:
See PR #2530; I assumed it should go into 'devel' so it is right in new releases, but Gina asked for it to be done here.

refusing to fill out the template for a docfix :)


foosel:
Thanks!

Issue #485, opened 2014-05-28
Timelapse render bar/status


SelfDestructor:
Is it possible to add a progress bar or status box next to recorded timelapses?
A simple "Rendering" and "Finished" message would really help as right now I'm downloading the movie to see if it's finished rendering to the end.

foosel:
Progress bar would necessitate ffmpeg output parsing, haven't yet looked if that's feasible. Status indicators on the timelapse list might be possible, but aren't implemented yet.

But the "Timelapse rendering" notification is now persistent, even across reloads/client switches. That should make it easier to see that a rendering job is currently in progress. 

Available on the `maintenance` and `devel` branches and going to be included in the 1.2.9 release.

foosel:
This has been implemented by the above PR on `maintenance` and `devel` and will be part of 1.3.9

foosel:
1.4.0 has been released.

PR #2586, opened 2018-04-16
Added timelapse render progress bar

ZachNo:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

It adds a progress bar to the timelapse tab so that users who have created a very large timelapse are able to view the progress of FFMPEG. This was requested in ticket #485 and I feel this is beneficial to all Octoprint users.

By using the virtual printer, I emulated an approximate 5 minute print where a proper webcam was setup and recording. Optionally, fast testing can be done by printing and canceling. Thorough testing was done from a few seconds of recording to several minutes.

Lots of debugging/refactoring took place to minimize the touched files while still keeping in line with the general architecture.

#485

image
https://www.youtube.com/watch?v=ZO0IboxewO8 (ignore the misaligned white text, I've fixed that)

#### Further notes
The reason for the removal of the logging option for FFMPEG, is that error level does not provide the output needed to generate a progress percentage.


src/octoprint/timelapse.py
```diff
@@ -838,7 +839,40 @@ def _render(self):
...
 				self._notify_callback("start", output)
-				p = sarge.run(command_str, stdout=sarge.Capture(), stderr=sarge.Capture())
+				p = sarge.run(command_str, stdout=sarge.Capture(), stderr=sarge.Capture(), async=True)
+
+				self._logger.debug("Parsing ffmpeg output")
+
+				# Compile regex to make it faster
+				duration_regex = re.compile("Duration: (\d{2}):(\d{2}):(\d{2})\.\d{2}")
+				current_regex = re.compile("time=(\d{2}):(\d{2}):(\d{2})\.\d{2}")
+				duration_s = ""
```
foosel: Shouldn't this rather be initialized to `0`? An empty string definitely feels very wrong here, especially since further down we try to parse that into a `float` which would throw a `ValueError`.

src/octoprint/timelapse.py
```diff
@@ -838,7 +839,40 @@ def _render(self):
...
+				# Sarge sometimes doesn't populate commands fast enough, so we check if the command exists first, then poll the thread
+				while len(p.commands) == 0 or p.commands[0].poll() is None:
+					line = p.stderr.readline()
+
+					# We should be checking time more often, so try it first
+					time = current_regex.search(line)
+					if time is not None:
+						current_s = self._parsetime(*time.groups())
+
+						progress = current_s / float(duration_s) * 100
```
foosel: We might not have had a duration yet, so it still could be zero, which would lead to a division by zero here. Needs some checking to make sure this can never happen.

src/octoprint/timelapse.py
```diff
@@ -838,7 +839,40 @@ def _render(self):
...
+
+				self._logger.debug("Parsing ffmpeg output")
+
+				# Compile regex to make it faster
+				duration_regex = re.compile("Duration: (\d{2}):(\d{2}):(\d{2})\.\d{2}")
+				current_regex = re.compile("time=(\d{2}):(\d{2}):(\d{2})\.\d{2}")
+				duration_s = ""
+
+				# Sarge sometimes doesn't populate commands fast enough, so we check if the command exists first, then poll the thread
+				while len(p.commands) == 0 or p.commands[0].poll() is None:
```
foosel: This whole sarge processing should best use the same pattern we use at other places where we fire up other processes with `async=True` and try to parse their output on stdout and stderr, see for example [here](https://github.com/foosel/OctoPrint/blob/c55c568bbddf5112facf6aa3a173d500936b6cbf/src/octoprint/util/commandline.py#L73-L128). While that stuff looks slightly convoluted, there are reasons for all of those steps. 

In fact, it would make a lot of sense to just utilize [`util.commandline.CommandlineCaller`](https://github.com/foosel/OctoPrint/blob/c55c568bbddf5112facf6aa3a173d500936b6cbf/src/octoprint/util/commandline.py#L56) here (with custom logger methods) since it already does all of that stuff for us.

src/octoprint/timelapse.py
```diff
@@ -838,7 +839,40 @@ def _render(self):
 		with self.render_job_lock:
 			try:
 				self._notify_callback("start", output)
-				p = sarge.run(command_str, stdout=sarge.Capture(), stderr=sarge.Capture())
+				p = sarge.run(command_str, stdout=sarge.Capture(), stderr=sarge.Capture(), async=True)
+
+				self._logger.debug("Parsing ffmpeg output")
+
+				# Compile regex to make it faster
```
foosel: Could you add an example output line that this will parse, for future reference? Would help maintenance down the road tremendously :)

src/octoprint/timelapse.py
```diff
@@ -853,6 +887,10 @@ def _render(self):
 			finally:
 				self._notify_callback("always", output)
 
+	@staticmethod
+	def _parsetime(hours, minutes, seconds):
```
foosel: It doesn't really `parse` though but rather `convert`, right? 🤔 Or am I misunderstanding something here?

koenkooi:
Thanks for showing how the tests for timelapse work, now I can make some progress on #2410 !

ZachNo:
I've changed the code to use CommandlineCaller now with a custom logger function.
I've also added a small test for ffmpeg output parsing.

foosel:
I fixed a small issue where at least on Windows ffmpeg's output would only be parsed at the end of the encoding. [This is caused by ffmpeg using `\r` for the progress lines instead of `\n` and Python's popen not parsing those as line endings by default](https://stackoverflow.com/a/35422274).

ManuelMcLure:
How well will this work if we've changed the timelapse rendering script? I've changed my timelapse settings to call a python script that uses ssh to run ffmpeg on a different (more powerful) server so I can render to H.264.

foosel:
@ManuelMcLure Sorry, that question got buried...

If such a rendering script would not produce FFMPEG progress output, the progress bar would just sit there. The rendering would proceed, but there would be no updates.

It might make sense to have the progress bar sit at the "busy" animation and only switch to actual progress mode once the first render progress event is received.

ManuelMcLure:
That works. I just wanted to make sure that it wasn't going to cause a hang or crash. I can live with no progress bar.

PR #2595, opened 2018-04-19
Fix/appearance

dforsi:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

The first 3 commits are bug fixes:
* Fix caret icon always shown in position "closed" even whe details are visible
* Make the API buttons and title in the Settings similar to those in other dialogs
* Fix trivial copy and paste errors

The others are enhancements:
* Replace "Cancel" label with "Abort" on buttons for consistency
* Hide the buttons when the file list is hidden
* Add some icons to the About page

Tested using the page.

For the About page the visible change is as follows.
Before
localhost_5000_ 2

After
localhost_5000_ 5

It would be nice to add a tentacle before the word OctoPrint in the title with something (useless :-) happening when clicked.

### DIFF
--- a/src/octoprint/plugins/pluginmanager/templates/pluginmanager_settings.jinja2
+++ b/src/octoprint/plugins/pluginmanager/templates/pluginmanager_settings.jinja2
@@ -306,8 +306,7 @@
         </form>
     </div>
     <div class="modal-footer">
-        <button class="btn" data-dismiss="modal" aria-hidden="true">{{ _('Cancel') }}</button>
+        <button class="btn" data-dismiss="modal" aria-hidden="true">{{ _('Abort') }}</button>
         <button class="btn btn-primary" data-bind="click: savePluginSettings" aria-hidden="true">{{ _('Save') }}</button>
     </div>
 </div>
-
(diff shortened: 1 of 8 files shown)


foosel:
About the "Cancel" to "Abort" change, I've actually thought about switching over to "Close" instead. There are a bunch of places in the settings dialog where you can modify things without having to hit "Save" to persist things (e.g. user management, api key changing, individual plugin settings dialogs) and it always feels odd modifying something successfully and then hitting "Cancel" in such cases. It would feel even more odd to hit "Abort", so I feel a bit reluctant to do that particular change and would rather see things move over to "Close" or something similar. Thoughts?

dforsi:
I'm not sure anymore about button labels because there are two kinds settings dialogs:
* the small dialogs where the buttons send the changes immediately to the server, so the "Confirm" button is actually a "Save" and I'm not sure which of "Abort", "Cancel" or "Close" would be better to mean the opposite of save
* the big dialog where Cancel/Save apply only to the big dialog and not to the small dialogs

All in all, in both cases the two buttons act the dialog currently open and not for any dialog previously closed, so it would be ok for me if all the dialog had Cancel/Save.

foosel:
I've merged this but reverted the "Abort" commit (and one typo fix that wasn't actually fixing a typo but an intentional choice of name). I've also changed the button for the general settings dialog to "Close" for now but left the sub dialogs as is.

Issue #2539, opened 2018-04-02
Serial port and baud rate not consistently updated/displayed between main page and settings


dforsi:
#### What were you doing?

I was editing connection settings.

First case

1. Start with Port=AUTO and baud rate=AUTO
2. Choose a different port and/or a baud rate in the main window
3. Select "Save connection settings"
4. Click "Connect"
5. Open "OctoPrint Settings" from the wrench icon and see that it still shows AUTO/AUTO
6. Refresh the whole page, open "OctoPrint Settings" again and see that now it shows the updated values

Second case

1. Open "OctoPrint Settings" from the wrench icon
2. Add a serial ports glob `/dev/tty*` (they won't work for printing but they would fill the list nicely) and/or an additional baud rate
3. Click "Save"
4. In the main page see that the new values do not appear in the list until you click the small refresh icon

A different bug is in how the list is ordered in the two views (eg. `tty1` near to `tty10` and sorted ascending) in the main window:
![image](https://user-images.githubusercontent.com/1055635/38191307-8453a004-3668-11e8-9aba-15d1698b61b4.png)

In "OctoPrint Settings" (eg. `tty1` near to `tty2` and sorted descending):
![image](https://user-images.githubusercontent.com/1055635/38191384-ce31bc24-3668-11e8-9626-9ebbbfaf88a8.png)

The same happens for baud rates (eg. add a baud rate of 1 and see it at top or at bottom of the list).

Edit: I don't know  if it's related but the contents of "Additional serial ports" and of "Additional baud rates" is updated only after a refresh if I click Cancel:

1. in "OctoPrint Settings" add something to "Additional serial ports" or to "Additional baud rates"
2. click save
3. reopen the settings and delete the contents
4. click "Cancel"
5. reopen the settings the `textarea` and the `input` are empty
6. refresh the page
7. reopen the settings the `textarea` and the `input` show the old values

#### What did you expect to happen?

Settings are updated in both places of the web UI as soon as they are successfully saved to server.

#### What happened instead?

A refresh is needed to see the updated settings.

#### Version of OctoPrint

1.3.7rc3 didn't test earlier versions

#### Printer model & used firmware incl. version

N/A

#### Browser and version of browser, operating system running browser

Chromium 62.0.3202.89

#### Link to octoprint.log

Nothing new gets logged when I save.

#### Link to contents of terminal tab or serial.log

N/A

#### Link to contents of Javascript console in the browser

Nothing new gets logged when I save.

[...]

PR #2606, opened 2018-04-23
Keep the Connection pane in sync with the Settings

dforsi:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

1. Return the Serial Port and the Baudrate lists without sorting them.
1. Refresh the list of serial ports and baud rates in the Connection pane when the corresponding settings are updated

#2539 

#### Further notes

This PR doesn't fix the another problem reported in #2539 (Settings dialog not updated when "Save connection settings" is checked in the Connection pane until manually refreshed).

### DIFF
--- a/src/octoprint/server/api/connection.py
+++ b/src/octoprint/server/api/connection.py
@@ -78,8 +78,8 @@ def _get_options():
 	default_profile = printerProfileManager.get_default()
 
 	options = dict(
-		ports=sorted(connection_options["ports"]),
-		baudrates=sorted(connection_options["baudrates"], reverse=True),
+		ports=connection_options["ports"],
+		baudrates=connection_options["baudrates"],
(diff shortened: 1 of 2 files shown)


foosel:
Hm... to be honest, I'm rather leaning towards making the settings match the connection tab rather than the other way around. At least the reverse sorting of the baudrate list was intentional (faster baudrate is definitely better than slower and also more likely to be what you are looking for when trying to find the right one for your particular printer). 

So how about modifying [`PrinterInterface.get_connection_options`](https://github.com/foosel/OctoPrint/blob/edd00cfa80ebe15b391aa633a820c67a55c7a7ec/src/octoprint/printer/__init__.py#L54) instead to return a sorted and if necessary reversed list? That would put things at one singular place which is used by both locations.

foosel:
I've added the sorting at the source as described above (incl. natural sorting, so that it goes `tty1`, `tty2`, `tty10`, `tty11` now instead of `tty1`, `tty10`, `tty11`, `tty2`) and merged.

PR #2698, opened 2018-06-24
make favicon color match custom theme color

tedder:
With multiple octopis, the theme/header color is really nice for quickly distinguishing them. But it doesn't help with an ocean of open tabs. This does- simply carry the theme color onto the favicon.

I could do this as a plugin and/or make it a configurable setting, but it's very small change that doesn't need that much overhead, and generally I have opinions about having a plethora of setting flags for nonfunctional things. Thanks to @jneilliii for [helping me through the viewmodels](https://discourse.octoprint.org/t/where-can-i-get-the-theme-color/2354/5).

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/appearance.js
+++ b/src/octoprint/static/js/app/viewmodels/appearance.js
@@ -1,10 +1,57 @@
 $(function() {
+    var save = [];
+
+    // Using the theme color (for the top bar), also recolor the favicon tentacle.
+    function ThemeFavicon(colorname) {
+        save.colorName = colorname;
+
+        // the following is from an Apache licensed snippet:
+        // http://blog.roomanna.com/09-24-2011/dynamically-coloring-a-favicon
+        var link = document.querySelector("link[rel~='icon']");
+        if (!link) {
+          link = document.createElement("link");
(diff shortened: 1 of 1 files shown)


foosel:
I've cherry picked this on `maintenance` aka the future 1.3.10 as f69718b. I've also changed the overlay method slightly (by using a different base image) and added a setting (because I am *absolutely* sure that if I don't people will complain ;)).

Since github doesn't detect cherry picks I'm closing this manually, however it *has* been merged.

tedder:
Can you hop on [the discourse thread](https://discourse.octoprint.org/t/where-can-i-get-the-theme-color/2354/9), @foosel? I don't think it actually *works*.

foosel:
It does, after I fixed some things ;) See 2a67772f4600afd89ccd86ae7828ec3e24e1e0b6. You did in fact use the title instead of the actual color (turning everything black). I fixed that and I also made it use a different base image so it's now a two-tone tentacle like the original one:

image

tedder:
Aha. I saw the change, it just seemed *broken*, like the color it chose was a closure, not a label.

When you push the RC I'll test it on a couple instances.

PR #2701, opened 2018-06-25
Fix add user part of the first run wizard

Salandora:

Fixes the first run wizards add user code

Manually removed the users.yaml and set firstRun to true.
Then run the wizard.


Salandora:
Output of the console before the fix:
https://pastebin.com/Mc5MXchm

Issue #2704, opened 2018-06-26
Timelapse URI test not working on 'devel' branch


koenkooi:
#### What were you doing?
I was trying to change the timelapse URI to point to the proxy instead of mjpg-streamer to fix the conflict between the timelapse and pushover trying to access it at the same time at the end of a build.

#### What did you expect to happen?

The test picture to pop up.

#### What happened instead?

```
Jun 25 14:33:01 blackbox octoprint[28754]: [2018-06-25 14:33:01,700] ERROR in app: Exception on /api/util/test [POST]
Jun 25 14:33:01 blackbox octoprint[28754]: Traceback (most recent call last):
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
Jun 25 14:33:01 blackbox octoprint[28754]:     response = self.full_dispatch_request()
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
Jun 25 14:33:01 blackbox octoprint[28754]:     rv = self.handle_user_exception(e)
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
Jun 25 14:33:01 blackbox octoprint[28754]:     reraise(exc_type, exc_value, tb)
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
Jun 25 14:33:01 blackbox octoprint[28754]:     rv = self.dispatch_request()
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
Jun 25 14:33:01 blackbox octoprint[28754]:     return self.view_functions[rule.endpoint](**req.view_args)
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/octoprint/server/util/flask.py", line 1158, in decorated_view
Jun 25 14:33:01 blackbox octoprint[28754]:     return func(*args, **kwargs)
Jun 25 14:33:01 blackbox octoprint[28754]:   File "/usr/lib/python2.7/site-packages/flask_principal.py", line 199, in _decorated
[...]

PR #2710, opened 2018-07-01
Fix for _test_url

Salandora:

Fixes missing import in _test_url function

Got to settings -> webcam -> Test Snapshot

No Exception got thrown in normal mode, only while debugging it told me it can't find util.parse_mime_type (and everything else in util.)

### DIFF
--- a/src/octoprint/server/api/__init__.py
+++ b/src/octoprint/server/api/__init__.py
@@ -343,6 +343,7 @@ def _test_path(data):
 
 def _test_url(data):
 	import requests
+	from octoprint import util as util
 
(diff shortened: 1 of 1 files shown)


foosel:
Argh. That refactoring error also made it into 1.3.9rc1. Could you recreate that PR against staging/maintenance?

foosel:
Never mind, that refactoring error indeed is only present in `devel`, 1.3.x has the `octoprint.util` import at the very top and hence no such issue.

koenkooi:
Does this fix #2704?

Issue #2143, opened 2017-10-05
[Request] Manipulate GCode from GCode Viewer


HeinerTholen:
Hey there,

I like to change filament and fiddle with the parts during prints. Currently, my solution is to sit next to the printer, wait, and hit pause in the right moment. It does the job.

However, the ideal feature would let me pick a solid line in the GCode viewer tab in OctoPrint and then give me a field to insert code before or after that line. Thereby I could retract, move the hotend away, beep, wait, play "Pour Elise", etc. 

Better even would be a feature that takes me from the solid line in the GCode viewer to the actual GCode line in a text editor, where I could make any modification. 

I guess it would be quite some effort, but it would be a legit feature, no? =)

Cheers and thank you for making OctoPrint! It's very very incredible!
Heiner

foosel:
> I guess it would be quite some effort, but it would be a legit feature, no?

Yes to both. Marked correspondingly ;)

tvi675:
To extend this it would be amazing if you could mark an area to ignore in the gcode viewer, for example if you’re printing multiple parts, one fails but the rest are ok, rather than waste a long print you can just ignore the rest of that part.

bradcfisher:
@tvi675 I've created a plugin that does exactly what you suggested.  There's an open PR to have it listed on the plugins site, but you can install it right before then using this url if you want to try it out:
https://github.com/bradcfisher/OctoPrint-ExcludeRegionPlugin/archive/0.1.0.zip

foosel:
@bradcfisher correction - the PR is no longer open but merged ;)

Closing this since there is now a plugin 🎉

foosel:
Oops... reopening since the original issue was about also inserting lines - my bad :)

LazeMSS:
Possible by plugin:
https://plugins.octoprint.org/plugins/GcodeEditor/

Issue #1973, opened 2017-06-17
GCODE viewer shows incorrect layer number


emgi2008:
For layer 1 & 2, the highest two numbers of the print are shown. 
What is actually layer 3 is displayed as number 1. The layer height is shown correctly. 
Example: 
[octo134-layerissue.txt](https://github.com/foosel/OctoPrint/files/1082567/octo134-layerissue.txt)

This behavior was first seen after upgrading from 1..3.2 to 1.3.4. (skipped 1.3.3)

GitIssueBot:
Hi @emgi2008,

It looks like there is some **information missing** from your bug report that will be needed in order to solve the problem. Read the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) which will provide you with a template to fill out here so that your bug report is ready to be investigated (I promise I'll go away then too!).

If you did not intend to report a bug but wanted to **request a feature or brain storm** about some kind of development, please take special note of the title format to use as described in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md).

**Please do not abuse the bug tracker as a support forum** - if you have a question or otherwise need some kind of help or support refer to the [Mailinglist](https://groups.google.com/group/octoprint) or the [G+ Community](https://plus.google.com/communities/102771308349328485741) instead of here.

Also **make sure you are at the right place** - this is the bug tracker of the official version of OctoPrint, not the Raspberry Pi image OctoPi nor any unbundled third party OctoPrint plugins or unofficial versions. Make sure too that you have **read through the [Frequently Asked Questions](https://github.com/foosel/OctoPrint/wiki/FAQ)** and searched the [**existing tickets**](https://github.com/foosel/OctoPrint/search?q=&ref=cmdform&type=Issues) for your problem - try multiple search terms please.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next two weeks (until 2017-07-01 14:00 UTC) I'll close this ticket so it doesn't clutter the bug tracker. This is nothing personal, so please just be considerate and help the maintainers solve this problem quickly by following the guidelines linked above. Remember, the less time the devs have to spend running after information on tickets, the more time they have to actually solve problems and add awesome new features. Thank you!

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being, so don't expect any replies from me :) Your ticket is read by humans too, I'm just not one of them.

[...]

PR #2717, opened 2018-07-06
Gcode viewer updates

bradcfisher:

This is a set of updates I've made to the gcode viewer for use by my Exclude Region plugin.  I though it'd be nice if I didn't have to hack OctoPrint to get the plugin to work, so thought I'd submit a PR.

Contains the following changes:

1) Rendering speed enhancements.  Speed improvements should be noticeable when compared side-by-side with the original version, even for relatively simple gcode files.  Essentially I applied the following optimizations:
  * y-axis inversion and the zoomFactor scaling applied to the context's transformation matrix instead of being applied to each coordinate value in the JavaScript
  * store commonly used array and object property lookups in local variables to avoid lookup overhead (e.g. var cmd = cmds[i])
  * when determining the last x,y position from the previous layer, exit the loop as soon as both values have been identified instead of scanning every command on the layer
  * compute as much as possible outside the layer rendering loop (colors, sizes, etc)
  * ensure paths are as long as possible before calling stroke(), instead of calling stroke() for every individual move

2) New event hooks:
  * 'onDragStart' - Invoked when the user presses the mouse button down.  Passed the point clicked on (in printer bed coordinates).  Return false to prevent dragging
  * 'onDrag' - Invoked when the user drags/pans the viewer.  Passed the current point the mouse is over (in printer bed coordinates).  Return false to prevent dragging
  * 'onDragStop' - Invoked when the user stops dragging.  Passed the point where the mouse button was released (in printer bed coordinates).
  * 'onViewportChange' - Invoked when a property of the canvas context's transformation matrix changes.  Passed the new transformation matrix.  Can be used to apply the same transformation to a different context (e.g. for layering multiple canvases over each other).  A future update could potentially apply this technique to optimize rendering previous/next layers and further reduce redraw times in certain circumstances by using two or three canvases instead of just one.

3) Consolidated reRender and render methods to removed redundancy (render now calls reRender instead of duplicating code).  This also addresses some potential range-checking issues that reRender had by applying some additional checks that render was already doing.

3) Bugfix to correct refresh() and refresh(0) both causing redraw of last drawn layer instead of refresh(0) drawing layer 0

Just tested by me so far.

To test rendering speed enhancements, select a file and pull up the gcode tab.  Zoom, pan, switch layers, etc.  Side-by-side comparison of the output generated by two instances of OctoPrint (one with the enhancement and one without) are helpful for ensuring both look identical.

The 'onDragStart' and 'onViewportChange' events are utilized by the Exclude Region plugin, so have been tested to the extent that I'm currently satisfied with how they function.

#1973 may be related to the refresh()/refresh(0) bugfix.
#2143 has a comment on it related to the functionality provided by the exclude region plugin, though the original request itself isn't directly relevant

See https://github.com/bradcfisher/OctoPrint-ExcludeRegionPlugin/raw/master/excluderegion-gcode-viewer.png

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -91,6 +91,7 @@ date of first contribution):
   * [Ganey](https://github.com/ganey)
   * [Sven Lohrmann](https://github.com/malnvenshorn)
   * [Andy Castille](https://github.com/klikini)
+  * [Brad Fisher](https://github.com/bradcfisher)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/static/gcodeviewer/js/renderer.js
+++ b/src/octoprint/static/gcodeviewer/js/renderer.js
@@ -10,7 +10,11 @@ GCODE.renderer = (function(){
 // ***** PRIVATE ******
     var canvas;
     var ctx;
-    var zoomFactor= 2.8, zoomFactorDelta = 0.4;
+
+    var viewportChanged = true;
+    var lineWidthFactor = 1 / 2.8;
+
+    var zoomFactorDelta = 0.4;
     var gridStep=10;
     var ctxHeight, ctxWidth;
     var prevX=0, prevY=0;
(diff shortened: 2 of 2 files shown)


foosel:
That's awesome! Could you do that PR against the `maintenance` branch though? I think it definitely merits an inclusion in 1.3.10 and it's something best qualified as "improvement" too, not a new feature, so that branch would make more sense.

I could also cherry pick (sadly a rebase doesn't work), but then this PR won't get detected as merged by Github. 

Tell me what works best for you :)

PR #2730, opened 2018-07-13
Improve octoprint gcode analysis speed

eyal0:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Increase gcode analysis speed by around 30%

Run this command with and without the patch:

time python ./src/octoprint/cli/analysis.py --speed-x=6000 --speed-y=6000 --max-t=10 --throttle=0.0 --throttle-lines=100 BIG_GCODE_FILE.gco

You will need to patch in the two lines that were added to analysis.py so that you can run analysis.py.  I recommended patching those in, too, because they do no harm and make it easier to profile the analysis.  To profile, run:

python -m cProfile -s tottime ./src/octoprint/cli/analysis.py --speed-x=6000 --speed-y=6000 --max-t=10 --throttle=0.0 --throttle-lines=100 BIG_GCODE_FILE.gco | less

Notice that the speed is improved but the output is unaffected.

While working on the analysis factory, I noticed that default analysis is very slow and tried to improve it.  It would run faster if it were in c but that would be a lot more work.  This is a good start.


arhi:
I'm blind or replacing if None with if float(max) cuts 30% of the execution time? I'd report that as python bug :(

eyal0:
Three changes.

1. Don't import math on each call to getFloat
2. Unroll the loop so that we don't need to call getattr, which is doing reflection. Also, loop unrolled so that's faster.
3. We don't need to test against None each time by using infinity.

Check for yourself. I measured 30% on my laptop.

eyal0:
I faked unit tests. The comparison should be: is min greater than max then return None.

foosel:
Looks like you broke one of the existing tests there: [travis report](https://travis-ci.org/foosel/OctoPrint/builds/403408885?utm_source=github_status&utm_medium=notification). The `Example`s in the doc strings are doc tests. Could you fix that?

Also: Thanks for this. I have to admit I didn't know one could do `float("inf")` and learnt something new today! 👍

eyal0:
I undid the last two commits with the infinity stuff.  Those didn't affect the speed of the code as much as the rest.  Lost just 5 out of the 30 percent speed-up.

Those unit tests are kind of broken anyway.  They test the case where MinMax3D is called with a `None` coordinate but that does actually work?  I don't think so:

```
	>>> none_after_found = MinMax3D()
	>>> none_after_found.record(Vector3D(1,1,1))
	>>> none_after_found.record(Vector3D(2,2,2))
	>>> none_after_found.record(Vector3D(2,None,2))
	>>> str(none_after_found.size)
	Vector3D(x=1, y=0.0, z=1, length=1.41421356237)
```
I would expect that the None means no data so size would be 1x1x1.  Instead, the y size was set to zero.  This is for the new and old code.

So really, this function doesn't work properly with `None` in any of x, y or z.  So all the test cases where `None` is in one of the inputs isn't valid anyway.  Luckily, the code never inputs a None so it doesn't break.

5% isn't that much but on RPi is can feel like a lot!  If you want it done, let me know and I'll make another PR.

foosel:
I agree, looking at those tests now I'm not entirely sure what's up there either. I'll happily accept another PR that gets rid of anything senseless there ;)

PR #2732, opened 2018-07-14
Map/filter wrapped into list()

razerraz:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

In python 3, map/filter return specific object types that :
* are not directly iterable
* doesn't work with len() function
* doesn't allow list concatenation

tests are ok

python3 compatibility brainstorm

### DIFF
--- a/docs/sphinxext/codeblockext.py
+++ b/docs/sphinxext/codeblockext.py
@@ -84,7 +84,7 @@ class CodeBlockExt(CodeBlock):
 				return node
 
 		# replace all created literal_blocks with literal_block_ext instances
-		return map(find_and_wrap_literal_block, code_block)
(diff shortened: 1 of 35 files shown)


razerraz:
Maybe we have to think more deeply about this PR.
In python3, almost everything returns iterables when python2 returns list.
Seems a good idea to follow this behavior in the code when it's possible, and I have to look deeply on that

razerraz:
To be more precise about my thinking, we have to look both at itertools for python2 compatibility and ways to handle iterables more deeply inside code, imho...
In case we achieve all unicode strings and bytes conversion smoothly, ie in my dreams....

PR #2804, opened 2018-09-02
Fixed up documentation for system api. Now correctly matches current behaviour

cameroncros:
Fixes the system API documentation.

`Sphinx-build -b html . html`
All the changes verified, but I didn't manage to get html output.


foosel:
👍

PR #2807, opened 2018-09-02
201809 py3 devel prep batch b

tedder:
Work towards python3 compatability, focusing on tests. The following tests and related code are now compatible with both py2 and py3.

- tests/util/test_caseinsensitive_set.py
- tests/util/test_pip.py
- tests/timelapse/test_timelapse_renderjob.py
- tests/server/util/test_flask.py
- tests/filemanager/test_filemanager.py

### DIFF
--- a/src/octoprint/util/__init__.py
+++ b/src/octoprint/util/__init__.py
@@ -11,6 +11,7 @@ __license__ = 'GNU Affero General Public License http://www.gnu.org/licenses/agp
 
 import os
 import traceback
+import six
 import sys
 import re
 import tempfile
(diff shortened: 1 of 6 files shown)


src/octoprint/util/__init__.py
```diff
@@ -11,6 +11,7 @@
 
 import os
 import traceback
+import six
```
foosel: Hmm... I've so far avoided using `six` (even though it is a sub dependency due to I think flask and werkzeug) and instead used [python future](https://python-future.org/) as compatibility layer. I think for consistency reasons it would probably be better to stick to that and instead of replacing `(str, unicode)` with six.string_types replace it with `basestring` imported from `past.builtins`.

PR #2828, opened 2018-09-23
tedder's python3 fixes, round 3

tedder:
Leaving individual commits because they each tackle a subtopic. Let me know if you disagree with any of the solutions.

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -1502,7 +1502,7 @@ class Settings(object):
 		return value is not None
 
 	def getBaseFolder(self, type, create=True, allow_fallback=True, log_error=False, check_writable=True, deep_check_writable=False):
-		if type not in default_settings["folder"].keys() + ["base"]:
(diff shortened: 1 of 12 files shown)


tedder:
Travis barfed but I don't think it's my fault- the past.builtins import is in 18 libs.

foosel:
I fixed the error (`past.builtins` isn't yet available when `octoprint_setuptools` is invoked as that's invoked during the build and before any dependencies such as `future` are pulled in and available). Now it's failing due to the chainmap issues you mentioned in #2829 if I'm not mistaken, correct?

foosel:
Heh, the error with the chainmap was actually caused by a bug in your `_load_yaml` helper, which was opening the file not for reading but for writing. Fixing that made the tests green which means I can now merge this. 

What I'm not clear on yet though is if there is now still something left with regards to the chainmap issue, as what you mentioned was caused by an empty config.yaml, which less a problem with Py3 and more with properly handling that case in general. Can you clarify if there's still something else?

PR #2840, opened 2018-10-09
Py3 unicode type removing

razerraz:
Prepare Python 3 migration by removing all references to unicode python2 string type


src/octoprint/util/__init__.py
```diff
@@ -502,6 +504,10 @@ def filter_non_ascii(line):
 
 def to_str(s_or_u, encoding="utf-8", errors="strict"):
 	"""Make sure ``s_or_u`` is a str."""
+	if PY3:
```
foosel: Isn't this backwards? `to_str` should basically be `to_bytes` under Py3, so we'd need to check whether stuff is a `unicode` in 2 or `str` in 3 and if so decode, otherwise just return, or not?

src/octoprint/util/__init__.py
```diff
@@ -510,6 +516,8 @@ def to_str(s_or_u, encoding="utf-8", errors="strict"):
 
 def to_unicode(s_or_u, encoding="utf-8", errors="strict"):
 	"""Make sure ``s_or_u`` is a unicode string."""
+	if PY3:
```
foosel: Backwards? See above

src/octoprint/util/comm.py
```diff
@@ -949,10 +952,14 @@ def _getGcodeScript(self, scriptName, replacements=None):
 					continue
 
 				def to_list(data):
-					if isinstance(data, str):
-						data = map(str.strip, data.split("\n"))
-					elif isinstance(data, unicode):
-						data = map(unicode.strip, data.split("\n"))
+					if PY3:
```
foosel: Come to think of it, wouldn't something like 

``` python
data = map(lambda x: x.strip(), data.split("\n"))
```

actually be the trivial solution here?

src/octoprint/util/__init__.py
razerraz: Giving a better look at this : 
Strings objects are always unicode in Py3. Then, unicode object become string object in Py3.
Bytes objects are just containers commonly used when playing with io
The pep describe that well
https://www.python.org/dev/peps/pep-0358/

With that in mind, to_unicode() calls will become useless in python3, and keeping the calls in the code could be a easy way to get rid of this mess when dropping Py2

src/octoprint/util/comm.py
razerraz: Sorry, I don't get how you will get rid of the tests, and avoiding to have unicode reference when running python3 with that.

foosel: What I mean is that this whole section of code with the `if`s is only there to decide what function to hand to `map` based on the type of `data`, because handing the wrong type to either function will result in an error: `str.strip` is it's a py2 str, `unicode.strip` if it's a py2 unicode.

But instead of figuring out what data type it is (which gets more complicated once py2/py3 need to coexist), we can just as well just say that we don't care as long as it has a `strip` function (which is an assumption here) and just go for that by calling it directly on the object in the `lambda` function. Probably wrapped with something like `if isinstance(data, basestring)` though to take care of the assumption.

I might be bad today at explaining myself for some reason...

src/octoprint/util/__init__.py
foosel: The thing is that `to_unicode` and `to_str` should probably have been called `to_unicode_string` and `to_byte_representation`. 

`to_unicode` should ensure that whatever it gets, py3 str or py3 bytes or py2 unicode or py2 str, a unicode string (py3 str & py2 unicode) is the result.

`to_str` should ensure that whatever it gets, py3 str or py3 bytes or py2 unicode or py2 str, a byte representation (py3 bytes & py2 str) is the result.

src/octoprint/util/comm.py
razerraz: Ok, now I see it. I've always sucked understanding lambda things, don't like it

razerraz:
Changes done in the 2 latest commits

foosel:
Closing this since it has been superseded by #3032 which has now been merged to `devel`. 

Still: Thank you! 👍

PR #2851, opened 2018-10-13
Py3 compatible serial calls

razerraz:
Prepare Python 3 migration, taking care of str <-> bytes conversion on every read/write serial calls
 * serial.read() returns an str in python2, a byte in python3
 * serial.write() takes an str as argument in python2, a byte in python3

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -12,6 +12,9 @@ import re
 import threading
 import contextlib
 import copy
+import sys
+
+PY3 = sys.version_info[0] == 3
 
 try:
 	import queue
@@ -2573,6 +2576,9 @@ class MachineCom(object):
 				self.close(is_error=True)
 			return None
 
+		if PY3:
+			ret = ret.decode()
+
 		if ret != "":
 			try:
 				self._log("Recv: " + sanitize_ascii(ret))
@@ -3169,7 +3175,10 @@ class MachineCom(object):
 			old_written = written
 
 			try:
-				result = self._serial.write(to_send)
+				if PY3:
+					result = self._serial.write(to_send.encode('utf-8'))
+				else:
+					result = self._serial.write(to_send)
 				if result is None or not isinstance(result, int):
 					# probably some plugin not returning the written bytes, assuming all of them
 					written += len(cmd)
@@ -3178,7 +3187,10 @@ class MachineCom(object):
 			except serial.SerialTimeoutException:
 				self._log("Serial timeout while writing to serial port, trying again.")
 				try:
-					result = self._serial.write(to_send)
+					if PY3:
+						result = self._serial.write(to_send.encode('utf-8'))
+					else:
+						result = self._serial.write(to_send)
 					if result is None or not isinstance(result, int):
 						# probably some plugin not returning the written bytes, assuming all of them
 						written += len(cmd)


src/octoprint/util/comm.py
```diff
@@ -3169,7 +3175,10 @@ def _do_send_without_checksum(self, cmd, log=True):
 			old_written = written
 
 			try:
-				result = self._serial.write(to_send)
+				if PY3:
+					result = self._serial.write(to_send.encode('utf-8'))
```
foosel: Careful. According to the [reprap wiki](https://reprap.org/wiki/G-code) the interface between hosts and printers is ASCII only:

> All communication is in printable ASCII characters.

That line admittedly is located in the section that talks about replies back from the printer, based on its phrasing and also my general experience with that stuff so far I'm wondering if we can risk encoding to utf-8 here.

razerraz: You're right. Ascii encoding seems a way better solution.
Besides, this utf8 encoding have passed all the tests I've made, virtual and marlin printer, since it doesn't change anything for ascii characters I suppose.
I will test with that

JustAnother1: utf-8 is identical to ASCII in the first 128 characters ( codes 0 to 127) Above that the code pages(cp850 or cp437,..) come into play and you can either have special characters or those ugly lines for that dos window stuff.
Therefore you will not find any issues regardless of ASCII or utf8.
But for better compatibility with different language specific character sets I would suggest to use utf8 as it more and more becomes the standard. And you would prevent issues with special characters like ß or ä in the future.

razerraz: > utf-8 is identical to ASCII in the first 128 characters
That's what I thought. In addition, after having worked a lot with weird serial to RF modules like xbee, and always use utf8 as encoding despite the ascii mode of the modules.
> And you would prevent issues with special characters like ß or ä in the future
In case we add in the future any gcode command with a famous octoprint maintainer as name :1st_place_medal:

JustAnother1: I respect your cautious approach. But https://en.wikipedia.org/wiki/ASCII
specifies the 8th bit to be a parity bit. Therefore even the 60s style pure ASCII would need to support 8 bit data. And the reason for parity is that it sometimes doesn't match. So all 256 codes need to be acceptable for a device that claims to support ASCII. 

Are you sure that Python "ASCII" is really 7bit ASCII from the 1960s? Or might it be extended ASCII https://en.wikipedia.org/wiki/Extended_ASCII

So basically does ASCII handle German umlauts? If it does then it is extended ASCII and therefore an 8bit character set. If it works then it probably only works on your PC as you have configured the correct code page. With extended ASCII the code page has to be configured correctly for those new characters to work. Do you really want the user to specify her code page?  Sending utf8 instead of extended ASCII results in non ASCII characters becoming two characters. But that is already someone misusing the system by sending non supported characters. All pure ASCII codes are identical. And if the firmware supports utf8 then it just works. So changing from ASCII to utf8 changes the meaning of the codes 128 to 255 from a now undefined (or differently defined) state into a well known working state.

Therefore if 8bits are OK the only sensible thing to do is to send utf8.

smurfix: The problem is, you really don't want to reject your input because of encoding errors. Might just be a stray 8th bit or whatever. So if we really want to be safe, incoming should be decoded with latin1 if UTF-8 doesn't work.

Outgoing, we don't have a non-ASCII usecase, so until we do we should limit the conversion to ASCII.

DanielJoyce:
I would think this would be handled by printer-profile data, with the default assumed to be ASCII and otherwise overrideable in the profile.

foosel:
I think the solution implemented by @smurfix in #3032 is a good compromise: We still only ever send 7bit-ASCII, but now do allow receiving UTF8.

Closing this since #3032 has been merged. Thanks everyone for the discussion! 👍

razerraz:
I'm currently testing devel branch with a py3 venv : seems to work fine, what a tremendous job you've done on a small amount of time it's amazing. The speed gain is noticeable on the small nanopi board I'm using.
My test case is trying to install third party plugins on a fresh install, and I've already found a bug that prevent installing them, and sadly one of my PR is the cause. Fix can be found here :
https://github.com/razerraz/OctoPrint/commit/d269c361a4e4f90fbb35afa4948f8e88fe5fecd0

Cheers

PR #2867, opened 2018-11-01
add proof of concept throttling command

tedder:
Videocore is MIT licensed and has its source listed; this is entirely untested but I'm hoping it is 90% complete so you can complete the remaining 90%.

### DIFF
--- a/src/octoprint/plugins/pi_support/__init__.py
+++ b/src/octoprint/plugins/pi_support/__init__.py
@@ -6,7 +6,7 @@ __copyright__ = "Copyright (C) 2017 The OctoPrint Project - Released under terms
 
 import flask
 import os
-import sarge
+import videocore
 
 from flask_babel import gettext
 from octoprint.util import RepeatedTimer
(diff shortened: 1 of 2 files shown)


foosel:
Can you take a look into why the tests failed before I review this? Thank you! ❤️

tedder:
Life's in the way, but I'll tag you after I do, yeah.

foosel:
I've thought about this some more... To be honest, I'd rather like to stick with `vcgencmd get_throttled`. Yes, it means calling a subprocess, but on the other hand that makes it easier to maintain considering that it is an "official API". 

Whatever `vcgencmd get_throttled` is reporting to OctoPrint can also be easily verified directly from command line via the exact same way. Makes debugging any issues down the road easier, and also puts the responsibility for proper reporting at the correct functioning of that tool, not our implementation of reading out more or less undocumented system devices. 

In this case I think a fully integrated solution might be too clever for our own good :)

PR #2878, opened 2018-11-06
add throttle_state to tracking

tedder:
- weave the throttle_state through pi_support and tracking. This should come in handy to see how many throttled Pis are out there, as well as if it is correllated with canceled prints.
- upgraded disclosure of tracking information so it shows as INFO-level in the logs and all the information is dumped. If tracking.octoprint.org gets it, I should have it too :)

### DIFF
--- a/src/octoprint/plugins/tracking/__init__.py
+++ b/src/octoprint/plugins/tracking/__init__.py
@@ -169,7 +169,7 @@ class TrackingPlugin(octoprint.plugin.SettingsPlugin,
 		               ram=self._environment[b"hardware"][b"ram"])
 
 		if self._helpers_get_throttle_state:
-			payload[b"throttle_state"] = self._helpers_get_throttle_state()
+			payload[b"throttle_state"] = self._helpers_get_throttle_state(raw_value_only=True)
 
 		if b"plugins" in self._environment and b"pi_support" in self._environment[b"plugins"]:
 			payload[b"pi_model"] = self._environment[b"plugins"][b"pi_support"][b"model"]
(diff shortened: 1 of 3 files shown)


tedder:
I thought staging/maintenance was right to get it into the RC, 'bot.

foosel:
It is, but the bot doesn't know that ;) Ignore it...

---

First of all: I like the idea of adding info on the throttling state, great thinking! 👍 

There's still an issue though here since the current setup doesn't support nested data. This is how stuff arrives on the server:

```
[07/Nov/2018:06:26:16 +0100] "GET /track/REDACTED/print_done/?origin=local&throttle_state=raw_value&throttle_state=current_overheat&throttle_state=past_issue&throttle_state=past_overheat&throttle_state=current_issue&throttle_state=past_undervoltage&throttle_state=current_undervoltage&file=2f8bf9808c12c571457b73f35eca8baaad36b07c&elapsed=2486 HTTP/1.1" 204 0 "-" "OctoPrint/1.3.10rc1"
```

So that structure needs to be flattened. 

I'm also thinking it might be better to have the pi_support generate a custom event on status change and subscribe to that in the tracking plugin, and when the event is received send over the status update. That way it's easier to filter for, and the noise on the line is also reduced (our GET requests are limited in length as well by the HTTP spec).

Now, about the logging, I get where you are coming from, and I'm not against this, but it was done the way it was done for a simple reason: I don't want to see the instance UUIDs from people when they share their logs with me since that would de-anonymize them, and by logging this at INFO level I would ;) 

How about modifying this to not include the URL after all, at least not on INFO level? I *really* don't want to get knowledge of any UUIDs if it can be avoided :)

tedder:
I know I can ignore the bot, but sometimes it's nice to talk back a little. :)

FWIW GET is probably limited to about 2k characters. Certainly the nested dict isn't helping. I reduced it to only send the `raw_value` for the throttle state. I thought about two values so we could do basic bitmath before sending (so history would be 0-5, current would be 0-5), which would keep the server from knowing how to calculate it.

I know what you mean about sending it as an event, but it's sitting on about 20 characters now, which to me is worthwhile to not have to correlate GET requests to figure out if a failed print previously triggered overvoltage. Thoughts?

> How about modifying this to not include the URL after all, at least not on INFO level? I really don't want to get knowledge of any UUIDs if it can be avoided :)

As you can see I removed the URL and added a note to explain it. I think that's a great way to go.

Pushing the code and testing locally over the next 24h, but wanted to put it up for discussion.

src/octoprint/plugins/tracking/__init__.py
```diff
@@ -219,6 +231,11 @@ def _track_printjob_event(self, event, payload):
 			track_event = "print_failed"
 		elif event == Events.PRINT_CANCELLED:
 			track_event = "print_cancelled"
+		else:
+			logger.info("unknown event, skipping: {}".format(event))
```
foosel: That should be `debug`, otherwise we'll spam the log.

foosel:
> I know what you mean about sending it as an event, but it's sitting on about 20 characters now, which to me is worthwhile to not have to correlate GET requests to figure out if a failed print previously triggered overvoltage. Thoughts?

Why not both :D I added the events, so it's now possible to figure out how many throttled instances are out there without them needing to restart or print - for that I also added a new toggle so no one can complain there. And I also left the throttle state on the print events. Tested it against the backend and things are looking great.

tedder:
Looks great. Thanks for the modifications.

PR #2882, opened 2018-11-11
Prevent installing on python 3.

fake-name:
Right now, it installs, and then just explodes at runtime.

Fixes https://github.com/foosel/OctoPrint/issues/2881

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -218,6 +218,9 @@ def params():
 			"octoprint = octoprint:main"
 		]
 	}
+	
+	# Don't install on python 3
+	python_requires = '<3'
 
 	return locals()
 
--- a/src/octoprint/__main__.py
+++ b/src/octoprint/__main__.py
@@ -2,6 +2,10 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function
 
+import sys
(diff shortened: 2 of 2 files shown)


foosel:
The first of your commits has already been covered in 909dd551d6246b60760c26bbb6aaf448e3a21ef0 (which is part of 1.3.10rc1 but hasn't yet been merged to `devel` since I encountered some merge issues which I still have to sort out). To work around the issue with ancient setuptools - could you redo this PR against the `maintenance` branch with only the runtime check? Thanks :)

ByReaL:
i'll try make it work with python3 as i have time, i would like to still be able to install it on python3,
maybe a warning should be good or a way to force the install on python3

foosel:
`devel` branch is going to get this restriction lifted again for this very reason.

foosel:
I've cherry picked the second commit on `maintenance` as 52c7321c78c8a81b6b557d67fe5af44b8b32df4c (and commented out the changes on `devel` in 5b4f4e3d77f946a695abdb04188613ad4f21d648 while we work towards Py3 support there).

Closing this manually since Github doesn't detect cherry picks.

Issue #1030, opened 2015-08-20
[Request] Warn user about disk space issues


foosel:
OctoPrint should display a big red warning notification when the disk space for file uploads starts becoming a rare good (compare also #655). It should also detect when the disk space on the partition the configuration file is stored on runs low and refuse to even save settings from a certain point onward to prevent corruption of files.

- [ ] add setting for warning and "refuses to operate" space thresholds
- [ ] add new API to provide data (free vs total) for all the configurable folders (uploads, timelapses, timelapse temp, logs, watched, plugin data folder, ...) plus the location of the configuration file
- [ ] utilze API from frontend to query data and display information next to folder paths in the settings
- [ ] also display warning if user is admin when data anywhere falls below a certain threshold, include information how to free up space
- [ ] optional: include new hook for plugins to provide additional space information

ntoff:
Also what about temp space for mjpg streamer? I know it's not a part of octoprint but how would you handle things if say, you had 1gb free space and mjpg streamer took all of it with its temp files? Would you also warn users that there's potentially low space issues if they happen to be using a webcam and the print will take over an estimated certain amount of time (potentially using up all that temp space)?

foosel:
> I know it's not a part of octoprint

Actually it is. The capturing for timelapse snapshots is done by OctoPrint, the only part the mjpg_streamer plays here is providing the snapshot URL from which the frames are captured.

However, the temporary folder is part of the "all configurable folders" part above.

ntoff:
Sorry, what I meant by the second question was does octoprint know how much space is theoretically required for a time lapse video + snapshots based on how long it thinks the print will take? And does it take that into account when starting a print (for the "refuse to operate" feature)?

So if I set a warning for 500mb because gcode files are small, but I start a long print and there's say, 1gb left on the disk then octoprint will be happy. But while printing what if the time lapse ends up requiring 2gb of space.

Is there any way for octoprint to look at the remaining space and the timelapse quality settings and say "ok, this print will go for A hours, each snapshot will theoretically be 300kb, I'm going to need X amount of space for it and Y amount for the end video" and possibly throw a warning that, while there's sufficient space now, you will probably run out half way through the print and end up with a truncated timelapse video?

foosel:
Hm... Only by adding a setting of the average snapshot size. That depends on too many factors (resolution, quality, compressability which depends on the image itself etc) to reliably guess.

ntoff:
[...]

PR #2891, opened 2018-11-14
[WIP] Add api folders usage endpoint

HarlemSquirrel:
This is a work in progress. 

Adds an API endpoint for querying folder usage. Needed for #1030

### DIFF
--- a/src/octoprint/server/api/folders.py
+++ b/src/octoprint/server/api/folders.py
@@ -77,7 +77,6 @@ def _create_etag(path, filter, recursive, lm=None):
                                                                               request.values.get("recursive", False)),
                             unless=lambda: request.values.get("force", False) or request.values.get("_refresh", False))
 def readUsageForFolder(folder):
-	# import pry; pry()
 	folder_path = settings().getBaseFolder(folder, check_writable=False)
 
 	if folder_path is None:
--- a/src/octoprint/server/api/folders.py
+++ b/src/octoprint/server/api/folders.py
@@ -6,8 +6,6 @@ __license__ = 'GNU Affero General Public License http://www.gnu.org/licenses/agp
 __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms of the AGPLv3 License"
 
 
-import copy
-
 from flask import jsonify, make_response, request, url_for
 
 from octoprint.server.api import api
--- a/src/octoprint/server/api/folders.py
+++ b/src/octoprint/server/api/folders.py
@@ -1,0 +1,88 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function
+
+__author__ = "Gina Häußge <osd@foosel.net>"
+__license__ = 'GNU Affero General Public License http://www.gnu.org/licenses/agpl.html'
+__copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms of the AGPLv3 License"
+
+
+import copy
+
+from flask import jsonify, make_response, request, url_for
+
+from octoprint.server.api import api
+from octoprint.server import printer, fileManager, slicingManager, eventManager, NO_CONTENT, current_user
+from octoprint.server.util.flask import restricted_access, get_json_command_from_request, with_revalidation_checking
+from octoprint.settings import settings, valid_boolean_trues
+
+import psutil
+import logging
+
+def _create_lastmodified(path, recursive):
+	if path.endswith("/files"):
+		# all storages involved
+		lms = [0]
+		for storage in fileManager.registered_storages:
+			try:
+				lms.append(fileManager.last_modified(storage, recursive=recursive))
+			except:
+				logging.getLogger(__name__).exception("There was an error retrieving the last modified data from storage {}".format(storage))
+				lms.append(None)
+
+		if filter(lambda x: x is None, lms):
+			# we return None if ANY of the involved storages returned None
+			return None
+
+		# if we reach this point, we return the maximum of all dates
+		return max(lms)
+
+	elif path.endswith("/files/local"):
+		# only local storage involved
+		try:
+			return fileManager.last_modified(FileDestinations.LOCAL, recursive=recursive)
+		except:
+			logging.getLogger(__name__).exception("There was an error retrieving the last modified data from storage {}".format(FileDestinations.LOCAL))
+			return None
+
+	else:
+		return None
+
+
+def _create_etag(path, filter, recursive, lm=None):
+	if lm is None:
+		lm = _create_lastmodified(path, recursive)
+
+	if lm is None:
+		return None
+
+	hash = hashlib.sha1()
+	hash.update(str(lm))
+	hash.update(str(filter))
+	hash.update(str(recursive))
+
+	if path.endswith("/files") or path.endswith("/files/sdcard"):
+		# include sd data in etag
+		hash.update(repr(sorted(printer.get_sd_files(), key=lambda x: x[0])))
+
+	hash.update(_DATA_FORMAT_VERSION) # increment version if we change the API format
+
+	return hash.hexdigest()
+
+@api.route("/folders/local/<string:folder>/usage", methods=["GET"])
+@with_revalidation_checking(etag_factory=lambda lm=None: _create_etag(request.path,
+                                                                      request.values.get("filter", False),
+                                                                      request.values.get("recursive", False),
+                                                                      lm=lm),
+                            lastmodified_factory=lambda: _create_lastmodified(request.path,
+                                                                              request.values.get("recursive", False)),
+                            unless=lambda: request.values.get("force", False) or request.values.get("_refresh", False))
+def readUsageForFolder(folder):
+	# import pry; pry()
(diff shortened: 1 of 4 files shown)


HarlemSquirrel:
As I'm new to developing OctoPrint, I'm not sure why `ValueError: Settings not initialized yet` is getting raised in the test.

foosel:
I'll take a look ASAP but it might be a couple of days because of the current 1.3.10 RCs.

HarlemSquirrel:
I figured out how to initialize settings. It couldn't be easier than `settings(init=True)` 😋

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
+# coding=utf-8
```
foosel: Hm... I'd rather not have a completely new sub API just for this to be honest, especially not with an arbitrary name like `/folders` that could just as well refer to uploaded files. IMHO fetching information about the system such as this could be located either underneath the `/settings` endpoint or maybe the `/system` endpoint (e.g. `/system/folders`).

I'd also like to see a "fetch all folder usage stats" endpoint, otherwise the frontend will have to do quite a number of requests just to be able to display consistent information. We can still have sub resources for the individual folders, but I'd first and foremost concentrate on a catch all.

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
...
+
+	if path.endswith("/files") or path.endswith("/files/sdcard"):
+		# include sd data in etag
+		hash.update(repr(sorted(printer.get_sd_files(), key=lambda x: x[0])))
+
+	hash.update(_DATA_FORMAT_VERSION) # increment version if we change the API format
+
+	return hash.hexdigest()
+
+@api.route("/folders/local/<string:folder>/usage", methods=["GET"])
```
foosel: Why `/local`? What other folders than local ones do you expect to query here? Do not mix this with upload locations, there already is an API for that.

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
...
+from octoprint.settings import settings, valid_boolean_trues
+
+import psutil
+import logging
+
+def _create_lastmodified(path, recursive):
+	if path.endswith("/files"):
+		# all storages involved
+		lms = [0]
+		for storage in fileManager.registered_storages:
```
foosel: Huh? Why is the stuff in file manager part of the lastmodified value for the internal folders?

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
...
+			return fileManager.last_modified(FileDestinations.LOCAL, recursive=recursive)
+		except:
+			logging.getLogger(__name__).exception("There was an error retrieving the last modified data from storage {}".format(FileDestinations.LOCAL))
+			return None
+
+	else:
+		return None
+
+
+def _create_etag(path, filter, recursive, lm=None):
```
foosel: Just like lastmodified, etag should also only look at the resources here, not filemanager stuff.

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
...
+		# include sd data in etag
+		hash.update(repr(sorted(printer.get_sd_files(), key=lambda x: x[0])))
+
+	hash.update(_DATA_FORMAT_VERSION) # increment version if we change the API format
+
+	return hash.hexdigest()
+
+@api.route("/folders/local/<string:folder>/usage", methods=["GET"])
+@with_revalidation_checking(etag_factory=lambda lm=None: _create_etag(request.path,
+                                                                      request.values.get("filter", False),
```
foosel: You use neither `filter` nor `recursive` so why is it part of this?

src/octoprint/server/api/folders.py
```diff
@@ -0,0 +1,85 @@
...
+                                                                      request.values.get("recursive", False),
+                                                                      lm=lm),
+                            lastmodified_factory=lambda: _create_lastmodified(request.path,
+                                                                              request.values.get("recursive", False)),
```
foosel: You don't use `recursive` (and it also doesn't make much sense here)

tests/server/api/test_folders.py
```diff
@@ -0,0 +1,52 @@
...
+
+class GetFolderUsageTest(unittest.TestCase):
+
+    def setUp(self):
+        # Use Flask's test client for our test.
+        self.test_app = app.test_client()
+
+    def test_readUsageForFolder(self):
+        usage_stub = UsageStub(free=randint(0, 100), total=randint(200, 1000))
+        psutil.disk_usage = MagicMock(return_value=usage_stub)
```
foosel: This actually replaces the original implementation instead of just mocking it out for the test. That might cause problems with other tests. Take a look at the [patch context manager](https://docs.python.org/3.7/library/unittest.mock.html#the-patchers).

tests/server/api/test_folders.py
```diff
@@ -0,0 +1,52 @@
...
+import psutil
+from random import randint
+import unittest
+
+from octoprint.settings import settings
+from octoprint.server import app
+from octoprint.server.api import api
+
+# Initialize settings
+settings(init=True)
```
foosel: This way you are using the actual settings (and if they don't exist yet, have new ones created, cluttering up whatever machine is running the test). You should instead `mock` the settings API and inject it into the resource under test. 

An example on how to that can be found [here](https://github.com/foosel/OctoPrint/blob/112a8b9c8bf7f5895f6a29dbe3d98ea4d9e0f400/tests/timelapse/test_timelapse_helpers.py#L23). In a nutshell:

  * in `setUp` create a patcher, getter and mock:

    ``` python
    def setUp(self):
        self.settings_patcher = mock.patch("octoprint.server.api.folders.settings") # needs to be adjusted to import
        self.settings_getter = self.settings_patcher.start()
        self.settings = mock.create_autospec(octoprint.settings.Settings)
        self.settings_getter.return_value = self.settings
    ```
  * in `cleanUp` clean up the patcher

    ``` python
    def cleanUp(self):
        self.settings_patcher.stop()
    ```
  * in your test method, mock out the behaviour you need from the settings in your expected scenario, e.g.

    ``` python
    def test_read_usage_for_folder(self):
        # setup
        self.settings.getBaseFolder.return_value = "/path/to/basefolder"

        # implement your test ...
    ```

tests/server/api/test_folders.py
```diff
@@ -0,0 +1,52 @@
...
+
+__author__ = "Gina Häußge <osd@foosel.net>"
+__license__ = 'GNU Affero General Public License http://www.gnu.org/licenses/agpl.html'
+__copyright__ = "Copyright (C) 2016 The OctoPrint Project - Released under terms of the AGPLv3 License"
+
+
+import json
+from mock import MagicMock
+import psutil
+from random import randint
```
foosel: I'm not a huge fan of random test data in tests - can make it impossible to reproduce a failure.

src/octoprint/server/api/folders.py
HarlemSquirrel: That makes sense. `/system/folders` sounds like a good place to me.

src/octoprint/server/api/folders.py
HarlemSquirrel: I was trying to mimic the structure used on the files endpoint but I totally agree that `/local` should not be needed here.

https://github.com/foosel/OctoPrint/blob/112a8b9c8bf7f5895f6a29dbe3d98ea4d9e0f400/src/octoprint/server/api/files.py#L110

src/octoprint/server/api/folders.py
HarlemSquirrel: Sorry, I am still getting familiar with Flask and the structure of OctoPrint so I copied a bunch of stuff from other places.

src/octoprint/server/api/folders.py
HarlemSquirrel: OK, makes sense

src/octoprint/server/api/folders.py
HarlemSquirrel: I copied from another endpoint so we can definitely lose those pieces.

tests/server/api/test_folders.py
HarlemSquirrel: OK, I will take a look

tests/server/api/test_folders.py
HarlemSquirrel: OK, no problem, we can use some static values.

tests/server/api/test_folders.py
HarlemSquirrel: This is a big help, thank you! I will get this in there.

foosel:
Heyas 👋 And happy new year!

I was just going through my todo list and wondering about the status of this? :)

HarlemSquirrel:
I haven't had a ton of time but I am working on it. I've had some trouble
testing the API with authentication so I'll probably push that up for a
review soon.

HarlemSquirrel:
Closing in favor of #3010

PR #2856, opened 2018-10-16
Dev/hook upload sd card

devildant:
Hello,

I created and tested this hook to configure the module I designed, to replace the SD copy system (SD printer) by a much more powerful method.
I reach speeds ranging from 25 to 50 MB / s.
so hook allows the creation of the octoprint plugin to replace the sd download.
I mentioned the topic in this issue: #2798
during my test I managed a copy of a gcode of 100mo in 14s

PR #2900, opened 2018-11-16
Improve/sdcard hook

devildant:
Hello @foosel ,
I applied / tested my changes for the hook sd on the branch "maintenance", I did this to avoid any extra work.
the original pull request is here :
#2856
I left it open for the moment, but we can close it if you want :)


src/octoprint/printer/standard.py
```diff
@@ -669,8 +679,19 @@ def add_sd_file(self, filename, absolutePath, on_success=None, on_failure=None,
 		else:
 			# probably something else added through a plugin, use it's basename as-is
 			remoteName = os.path.basename(filename)
-		self._create_estimator("stream")
-		self._comm.startFileTransfer(absolutePath, filename, "/" + remoteName,
+			
+		if len(self.sd_card_upload_hooks.items()) > 0:
```
foosel: Hooks that replace existing functionality so far always have a return value that indicates whether the hook took care of things or the default should be applied. See for example the `serial_factory` hook. This should be done here as well so that we really always run the default if no hook volunteered.

src/octoprint/printer/standard.py
```diff
@@ -93,6 +93,9 @@ def __init__(self, fileManager, analysisQueue, printerProfileManager):
 			except:
 				self._logger.exception("Error while processing analysis queues from {}".format(name))
 
+		#hook card upload
+		self.sd_card_upload_hooks = plugin_manager().get_hooks("octoprint.printer.cardupload")
```
foosel: Not 100% yet about the name. `octoprint.printer.sd_card_upload`?

docs/plugins/hooks.rst
```diff
@@ -1137,6 +1137,75 @@ octoprint.server.http.access_validator
...
+		
+      
+   ``
+
+   .. versionadded:: 1.3.11
+
+   :param str filename: filename inside octoprint
+   :param str absolutePath: absolutePath inside octoprint
+   :param str remoteName: name inside sd card printer
+   :param function is_sd_ready: check if sd card is ready on printer (no parameter)
```
foosel: Hm... instead of providing all these functions, maybe just inject `self` as `printer`? We do that on a couple of comm layer hooks already. Avoids having a ton of functions to inject.

If we can also somehow get rid of the success/error callbacks (see another comment, not sure if possible though, highly depends on the implementation I guess) we could reduce the call signature to

``` python
def sd_card_upload_hook(printer, filename, absolute_path, remote_name, *args, **kwargs)
```

What made you decide on this selection of injected methods btw?

docs/plugins/hooks.rst
```diff
@@ -1137,6 +1137,75 @@ octoprint.server.http.access_validator
 
       Implementing this hook will make your plugin require a restart of OctoPrint for enabling/disabling it fully.
 
+.. _sec-plugins-hook-octoprint-printer-cardupload:
+
+octoprint.printer.cardupload
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. py:function:: sd_card_upload_hook(filename, absolutePath, remoteName, on_success, on_error, *args, **kwargs)
```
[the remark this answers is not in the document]
devildant: yes there is a lot of parameter I agree, I can group
 callbacks in a json parameter, what do you think?

docs/plugins/hooks.rst
```diff
@@ -1137,6 +1137,75 @@ octoprint.server.http.access_validator
...
+   .. code-block:: python
+
+      payload = {
+			"local": filename,
+			"remote": remoteName,
+			"time": time.time() - timestart
+		}
+   This hook must also return the file name on the sd card when done.
+
+   **Example:**
```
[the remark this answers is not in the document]
devildant: hoo my bad i will fix that :)

src/octoprint/printer/standard.py
```diff
@@ -650,6 +653,13 @@ def get_sd_files(self, *args, **kwargs):
 			return []
 		return map(lambda x: (x[0][1:], x[1]), self._comm.getSdFiles())
 
+	def success_hook_sdcopy(self, payload):
```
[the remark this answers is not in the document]
devildant: my goal, with this change, is to be able to more discreetly replace the way to upload to the SD card, reminders "if successful" and "in case of error" allow me to follow the normal process if successful or of error (notification ..)
in fact I could possibly change the return of the hook by a Boolean.
Initially, I planned to return the name of the file, but it seems that changing the name of the file is problematic.
so I'm going to change the signature of the method to return a boolean

src/octoprint/printer/standard.py
devildant: mmm OK, as I understand it, I have to add a default method to the existing hook list, like that?
serial_factories = self._serial_factory_hooks.items() + [("default", default)]

src/octoprint/printer/standard.py
devildant: ha yes, I did not pay attention.
I will correct that :)

docs/plugins/hooks.rst
devildant: to be honest, I started with the python language with octoprint :p
I have not managed to use the "self" since the plugin, so I sent his reminders.
at the beginning i will replace the function return with a boolean (this will already eliminate two parameters)
I will try to use the printer parameter :)

devildant:
thanks for your return @foosel , i will try to fix all this asap.
I replied to your comments.
I will start the dev, but I have to wait for the end of my print to test this :)

docs/plugins/hooks.rst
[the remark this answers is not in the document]
devildant: normally it's fix

docs/plugins/hooks.rst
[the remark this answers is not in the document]
devildant: it's ok i reduce the number of parameter

src/octoprint/printer/standard.py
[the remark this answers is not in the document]
devildant: i remove the callbacks

src/octoprint/printer/standard.py
devildant: it's change :)

src/octoprint/printer/standard.py
devildant: fixed :p

docs/plugins/hooks.rst
devildant:
```
def sd_card_upload_hook(printer, filename, absolute_path, remote_name, *args, **kwargs)

```

it's done
sorry for the flood :p

devildant:
> Thanks for the updates! There are still some issues, see the inline comments.
> 
> I've also given the callbacks vs no callbacks situation another thought and now that we have the default implementation there have realized that it probably won't work without callbacks at all, simply because an upload can take a long while and we can't block the processing in the meantime.
> 
> I'm still a bit worried about third party plugins failing to call the callbacks, but that's a risk we'll just have to take.
> 
> I just tried to jot down an alternative implementation for this and then had to realize that the whole asynchronous nature of the file transfer makes it really tricky to implement. The problem here is the success signaling of the default implementation, which happens outside of the whole add_sd_file context and gets triggered by the comm layer. Having the callbacks just be methods bound on the printer object (like you had it initially) might indeed make the most sense in that case after all - sorry for not spotting that earlier 😕
> 
> With the implementation as-is now, the return value of the hook doesn't make a ton of sense in the default implementation, and it's impossible to have it return something meaningful since it might take several minutes until the upload completes. So - callbacks after all. Also: Always feel free to challenge me on my reviews - I can also miss use cases like now with the callback situation, and if I do please make me aware! :)
> 
> Finally, a quick heads-up: I'll leave for some holiday vacation tomorrow afternoon and will keep my hands off of anything OctoPrint related until my return, to give my batteries some chance to recharge after this year. I'll be back around the second week of January though. Just so you don't wonder.

I'm not sure I understand, do you want me to restore the on_success and on_error callbacks that I originally did?

devildant:
Actually in test

devildant:
I applied the changes.
I will test for a few days, even if everything seems to work.
Have good holidays

foosel:
> I'm not sure I understand, do you want me to restore the on_success and on_error callbacks that I originally did?

Yes, that was my point.

The asynchronous nature of doing a file transfer makes direct return codes not suitable. I missed that the first time around.

devildant:
no problem, at first I made the "callback" for that, and when we talk about it, I forgot: p
So, the idea of ​​the return function did not shock me :)

I made the changes and it seems OK.
but I think you'll live it after your vacation.
happy Holidays

foosel:
I refactored some things after experimenting with it a bit, but it's now merged! 👍 [Docs are live as well](http://docs.octoprint.org/en/maintenance/plugins/hooks.html#octoprint-printer-sdcardupload).

devildant:
Hoo thx :) 
I hope you enjoyed your holidays.
I will modify my plugin following your changes.

foosel:
> I hope you enjoyed your holidays.

I did, just as usual they felt too short 😂

devildant:
hahaa i understand.
I modified my plugin accordingly, parcontre I encounter 2 problems that I do not understand, the elements websocket are no longer received, and especially after upload the print does not launch automatically unlike before

foosel:
I don't understand the problem description. What websocket elements? And why did it launch right after upload before because that certainly wasn't the default behaviour? Or what do you even mean by "launch"?

devildant:
sorry it's my bad english :p
on my plugin I put, a menu in the sidebar that displays different status.
for that I use the websocket.
with this :
```
self._plugin_manager.send_plugin_message(__name__, dict(status=pstatus,success=psuccess,error=perror,speed=pspeed,time=ptime, size=psize, filename=pfilename))
```

image

name had the value of plugin_identifier, but now for me it's plugin_package.

i fix that it's ok 

for the SD upload, I use a cura plugin directly, which has an "automatically start print job after uploading" option
image
it does not work anymore, I do not understand where is the problem

devildant:
I have a track, I test it right now (I think a problem with the remote_name)

devildant:
the remote_name looks good, I do not understand

devildant:
I created the issue :) : https://github.com/foosel/OctoPrint/issues/3005

devildant:
sorry the test plugin took me a little while :)

PR #2924, opened 2018-12-01
py23 compat: replaced open with io.open

ByReaL:
  * [na] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

for python 2 and 3 compatibility is good to use io.open
also it is good top specify clearly if file is open in text mode or binary mode

all unit tests ran and passing
i had to modify the unitests to use io.open in Mock instead of builtins.open
also i had to modify the unitest calls to specify also in clear if file is used in binary or text mode to match the call from the actual source code
also i started octoprint and did a bit of system test especially around functionality that i believe deal with files and might have been impacted by my change

smaller change from a bigger commit for py23 compat

na

na

#### Further notes
risk: this was tested with py2 only and gets us closer to py3 compatibility but may require further changes to get python3 working properley

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -69,7 +69,7 @@ date of first contribution):
   * [Philipp Baum](https://github.com/philphilphil)
   * [Kyle Evans](https://github.com/kevans91)
   * [Javier Martínez Arrieta](https://github.com/Javierma)
-  * ["MirceaDan"](https://github.com/ByReaL)
+  * [Mircea Dan Gheorghe](https://github.com/ByReaL)
   * [Ovidiu Hossu](https://github.com/MoonshineSG)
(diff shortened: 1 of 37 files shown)


ByReaL:
risky but needed, all this unicode, bytes, string... business py2 and py3 is the most risky change
i tested it:
by running octoprint with python2 after the change, and doing some manual testing
by running all unit tests in python2 and made them pass 100%
by running all unit tests in python37 and made most of them pass  95% (few are failing but the idea is to keep python2 working as it is now and once all those 10+ pull request i have opened get in to finalize the unicode business by adding unicode literals at future import and making it work fully in python 3 as well)

PR #2934, opened 2018-12-02
py23 compat: setting encoding and  future import for all the files

ByReaL:
# coding=utf-8
from __future__ import absolute_import, division, print_function

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

some files had encoding set some did not some had a different format
set all files with same encoding in same format

imported in all files (majority had but there were few that did not)
from __future__ import absolute_import, division, print_function

this will prepare to add unicode_literals later to the list and ensure py23 compatibility in all files

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,6 @@
-# -*- coding: utf-8 -*-
+# coding=utf-8
+from __future__ import absolute_import, division, print_function
+
 #
 # OctoPrint documentation build configuration file, created by
 # sphinx-quickstart on Mon Dec 02 17:08:50 2013.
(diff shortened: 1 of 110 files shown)


ByReaL:
the first 2 lines in all the files will be the same, 
1) they will define the file encoding as utf-8
what is sad is the fact there is no standardization , emacs have a format, other editor other, and other other ... there are at least 3 ways we can tell python the source code is utf-8 , i just used the one that was the most present in majority of files already to reduce the number of changes.
2) from future import 
2a)  absolute import risk since it change the import behavior
2b) division risk since there may be division that relay on old behavior and that should be replaced with //, i could not find any such place, but still a risk
2c) print function, this should be safe as i could not find any print that is not already a function
2d) NOT present in this change list but this will open get us closer of a future change where we add unicode literals to same line , but that change will require a lot of work and testing, and must be done on it's own

PR #2969, opened 2018-12-23
report plugin exceptions to tracker

tedder:
one of the things that we're seeing is that outdated plugins are breaking Octoprint instances. Given an RC is installed by ~1000 people and about ~20-30 speak up about running it, it's a pretty safe bet that there are fairly large-scale plugin failures that are barely surfacing on github issues or discourse.

this doesn't catch ALL exceptions, just ones that come from plugins.

tested  by adding a `raise` to the cura plugin at load time. kinda fun:

    2018-12-23 19:59:46,312 - octoprint.plugins.tracking - INFO - Sent tracking event server_exception, payload: {'path': u'plugin/cura/on_startup', 'exception_trace': 'Traceback (most recent call last):\n  File "/home/user/devel/OctoPrint/src/octoprint/plugin/__init__.py", line 233, in call_plugin\n    result = getattr(plugin, method)(*args, **kwargs)\n  File "/home/user/devel/OctoPrint/src/octoprint/plugins/cura/__init__.py", line 86, in on_startup\n    raise Exception("ted quick fail test")\nException: ted quick fail test\n', 'plugin_name': u''}

- carefully look for event object and raise an event if a problem is caught
- add documentation for why 'events' is injected

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

### DIFF
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -31,6 +31,7 @@ class Events(object):
 	STARTUP = "Startup"
 	SHUTDOWN = "Shutdown"
 	CONNECTIVITY_CHANGED = "ConnectivityChanged"
+	EXCEPTION = "Exception"
 
 	# connect/disconnect to printer
 	CONNECTING = "Connecting"
@@ -165,6 +166,10 @@ class EventManager(object):
 					except:
 						self._logger.exception("Got an exception while sending event %s (Payload: %r) to %s" % (event, payload, listener))
 
+				# inject our event objects to the plugin manager. We import the plugin class, which
+				# means it can't import this events class without issues.
+				octoprint.plugin._eventmanager = eventManager()
+				octoprint.plugin._events = Events
 				octoprint.plugin.call_plugin(octoprint.plugin.types.EventHandlerPlugin,
 				                             "on_event",
 				                             args=(event, payload),
--- a/src/octoprint/plugin/__init__.py
+++ b/src/octoprint/plugin/__init__.py
@@ -22,6 +22,7 @@ __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms
 import os
 import logging
 import threading
+import traceback
 
 from octoprint.settings import settings as s
 from octoprint.plugin.core import (PluginInfo, PluginManager, Plugin)
@@ -31,6 +32,10 @@ from octoprint.util import deprecated
 
 # singleton
 _instance = None
+
+# injected in from 'events'. We can't import that class without circular issues.
+_eventmanager = None
+_events = None
 
 def _validate_plugin(phase, plugin_info):
 	if phase == "after_load":
@@ -234,6 +239,13 @@ def call_plugin(types, method, args=None, kwargs=None, callback=None, error_call
 				logger.exception("Error while calling plugin %s" % plugin._identifier)
 				if error_callback:
 					error_callback(plugin._identifier, plugin, exc)
+
+				if _eventmanager:
+					_eventmanager.fire(_events.EXCEPTION, payload={
+						'plugin_identifier': plugin._identifier,
+						'plugin_method': method,
+						'exception_trace': traceback.format_exc(exc)
+					})
 
 
 class PluginSettings(object):
--- a/src/octoprint/plugins/tracking/__init__.py
+++ b/src/octoprint/plugins/tracking/__init__.py
@@ -132,6 +132,8 @@ class TrackingPlugin(octoprint.plugin.SettingsPlugin,
 		elif event in (Events.FIRMWARE_DATA,) and self._record_next_firmware_info:
 			self._record_next_firmware_info = False
 			self._track_printer_event(event, payload)
+		elif event in (Events.EXCEPTION,):
+			self._track_server_exception_event(event, payload)
 
 	##~~ TemplatePlugin
 
@@ -357,6 +359,19 @@ class TrackingPlugin(octoprint.plugin.SettingsPlugin,
 		self._track("printer_safety_warning",
 		            printer_safety_warning_type=payload.get(b"warning_type", "unknown"),
 		            printer_safety_check_name=payload.get(b"check_name", "unknown"))
+
+	def _track_server_exception_event(self, event, payload):
+		if not self._settings.get_boolean([b"enabled"]):
+			return
+
+		path = '' # hopefully unique path to the code that failed
+		if payload.get(b'plugin_identifier'):
+			path = "/".join(["plugin", payload.get(b'plugin_identifier', ''), payload.get(b'plugin_method', '')])
+
+		self._track("server_exception",
+		            path=path,
+		            plugin_name=payload.get(b"plugin_name", ""),
+		            exception_trace=payload.get(b"exception_trace", ""))
 
 	def _track(self, event, **kwargs):
 		if not self._settings.get_boolean([b"enabled"]):


foosel:
I like the idea in principle because there's indeed huge potential during the RC phase with this. 

However, I have big privacy concerns with freely pushing stack traces to the tracking server. There might be personal information in the exception messages, e.g. "Cannot connect to X with API key Y" (already a problem with logs of course). And there are file paths involved too (which can contain user names). 

I actually have an account at sentry.io that I thought about using long term for some extra error tracking, since they also feature [data scrubbing](https://docs.sentry.io/data-management/sensitive-data/). But then again that's not self hosted so I feel a bit meh about it with GDPR now being active. 

Another problem is the payload length limit we currently have due to tracking happening via GET (I might have a solution for that though, have to test it).

tedder:
Maybe the thing to do is to fix the info leaks as they come up so it isn't toxic anymore. Once they are detected they shouldn't be horrible- sort of like when I masked the uuid for the pi support stuff.

As far as the paths are concerned, that could be masked too- I mean, basically we filter out `/home/.../` and `/Users/.../` and add more masks as they come up. That lets us still see the majority of the path, which should be good for detecting manual installs vs octopi style installs.

I'm coming out of 5 years of healthcare tech, which has pretty strong concerns around leaky logs, but it's still important to be able to troubleshoot, even if you are in a sensitive environment.

foosel:
Let me still put this on the backburner a bit and see how involved it would be to rather use a tool tailored to stacktrace and error analysis for that (e.g. a sentry.io instance).

foosel:
Huh... I tried to push the below to the branch, but either you've disabled that or I'm too stupid to do it right now 😂

In any case, the eventmanager injection isn't actually needed, see the change below. I also added limited scrubbing of the user's homedir at least. Still not 100% convinced though that this is the best approach to tracking these kinds of errors. I'll try to set aside a couple of hours in the next week or so and see if a local sentry install isn't better. We have a couple of challenges here. One is scrubbing the data. Another is the current GET based tracking. Then the `call_plugin` method isn't even remotely the only place where plugins are called - both hooks and implementations have some calls spread throughout the source, `call_plugin` is only a convenient helper for the most common calling pattern. Finally getting the data into the tracking server is one thing, being able to get it out again in a meaningful way is another and I don't think my current setup is well suited for that - it's more focused on running queries that produce numerical output then tracking singular events (even if they occur in bulk). So... I agree that it makes sense to look into error reporting, maybe not only for plugins but other kinds as well. But I'm not sure the platform setup to suite the anonymous usage tracking requirements is the perfect candidate for that and I'd like to look into alternatives first before committing to do this.

---

``` patch
Index: src/octoprint/events.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/octoprint/events.py	(date 1545586260000)
+++ src/octoprint/events.py	(date 1547226438000)
@@ -31,7 +31,7 @@
 	STARTUP = "Startup"
 	SHUTDOWN = "Shutdown"
 	CONNECTIVITY_CHANGED = "ConnectivityChanged"
-	EXCEPTION = "Exception"
+	PLUGIN_EXCEPTION = "PluginException"
 
 	# connect/disconnect to printer
 	CONNECTING = "Connecting"
@@ -166,10 +166,6 @@
 					except:
 						self._logger.exception("Got an exception while sending event %s (Payload: %r) to %s" % (event, payload, listener))
 
-				# inject our event objects to the plugin manager. We import the plugin class, which
-				# means it can't import this events class without issues.
-				octoprint.plugin._eventmanager = eventManager()
-				octoprint.plugin._events = Events
 				octoprint.plugin.call_plugin(octoprint.plugin.types.EventHandlerPlugin,
 				                             "on_event",
 				                             args=(event, payload),
Index: src/octoprint/plugin/__init__.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/octoprint/plugin/__init__.py	(date 1545586260000)
+++ src/octoprint/plugin/__init__.py	(date 1547226438000)
@@ -33,10 +33,6 @@
 # singleton
 _instance = None
 
-# injected in from 'events'. We can't import that class without circular issues.
-_eventmanager = None
-_events = None
-
 def _validate_plugin(phase, plugin_info):
 	if phase == "after_load":
 		if plugin_info.implementation is not None and isinstance(plugin_info.implementation, AppPlugin):
@@ -215,6 +211,8 @@
 	        is determined be presence of injected ``_identifier`` property.
 	"""
 
+	from octoprint.events import eventManager, Events
+
 	if not isinstance(types, (list, tuple)):
 		types = [types]
 	if args is None:
@@ -240,12 +238,9 @@
 				if error_callback:
 					error_callback(plugin._identifier, plugin, exc)
 
-				if _eventmanager:
-					_eventmanager.fire(_events.EXCEPTION, payload={
-						'plugin_identifier': plugin._identifier,
-						'plugin_method': method,
-						'exception_trace': traceback.format_exc(exc)
-					})
+				eventManager().fire(Events.PLUGIN_EXCEPTION, payload=dict(plugin=plugin._identifier,
+				                                                          method=method,
+				                                                          stacktrace=traceback.format_exc(exc)))
 
 
 class PluginSettings(object):
Index: src/octoprint/plugins/tracking/__init__.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/octoprint/plugins/tracking/__init__.py	(date 1545586260000)
+++ src/octoprint/plugins/tracking/__init__.py	(date 1547226438000)
@@ -11,6 +11,7 @@
 import requests
 import hashlib
 import logging
+import re
 
 try:
 	# noinspection PyCompatibility
@@ -27,6 +28,12 @@
 
 TRACKING_URL = "https://tracking.octoprint.org/track/{id}/{event}/"
 
+
+def _scrub_trace(stacktrace):
+	from os.path import expanduser
+	return stacktrace.replace(expanduser("~"), "REDACTED")
+
+
 # noinspection PyMissingConstructor
 class TrackingPlugin(octoprint.plugin.SettingsPlugin,
                      octoprint.plugin.EnvironmentDetectionPlugin,
@@ -67,7 +74,8 @@
 		                        update=True,
 		                        printer=True,
 		                        printer_safety_check=True,
-		                        throttled=True))
+		                        throttled=True,
+		                        plugin_exception=True))
 
 	def get_settings_restricted_paths(self):
 		return dict(admin=[["enabled"], ["unique_id"], ["events"]],
@@ -132,8 +140,8 @@
 		elif event in (Events.FIRMWARE_DATA,) and self._record_next_firmware_info:
 			self._record_next_firmware_info = False
 			self._track_printer_event(event, payload)
-		elif event in (Events.EXCEPTION,):
-			self._track_server_exception_event(event, payload)
+		elif event in (Events.PLUGIN_EXCEPTION,):
+			self._track_plugin_exception_event(event, payload)
 
 	##~~ TemplatePlugin
 
@@ -360,18 +368,17 @@
 		            printer_safety_warning_type=payload.get(b"warning_type", "unknown"),
 		            printer_safety_check_name=payload.get(b"check_name", "unknown"))
 
-	def _track_server_exception_event(self, event, payload):
+	def _track_plugin_exception_event(self, event, payload):
 		if not self._settings.get_boolean([b"enabled"]):
 			return
 
-		path = '' # hopefully unique path to the code that failed
-		if payload.get(b'plugin_identifier'):
-			path = "/".join(["plugin", payload.get(b'plugin_identifier', ''), payload.get(b'plugin_method', '')])
+		if not self._settings.get_boolean(["events", "plugin_exception"]):
+			return
 
-		self._track("server_exception",
-		            path=path,
-		            plugin_name=payload.get(b"plugin_name", ""),
-		            exception_trace=payload.get(b"exception_trace", ""))
+		self._track("plugin_exception",
+		            plugin=payload.get(b"plugin", ""),
+		            method=payload.get(b"method", ""),
+		            stacktrace=_scrub_trace(payload.get(b"stacktrace", "")))
 
 	def _track(self, event, **kwargs):
 		if not self._settings.get_boolean([b"enabled"]):
Index: tests/plugins/tracking/__init__.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/plugins/tracking/__init__.py	(date 1547226438000)
+++ tests/plugins/tracking/__init__.py	(date 1547226438000)
@@ -0,0 +1,0 @@
Index: tests/plugins/tracking/test_tracking.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/plugins/tracking/test_tracking.py	(date 1547226438000)
+++ tests/plugins/tracking/test_tracking.py	(date 1547226438000)
@@ -0,0 +1,27 @@
+from __future__ import unicode_literals
+
+import unittest
+import ddt
+import mock
+
+@ddt.ddt
+class TrackingTestCase(unittest.TestCase):
+
+	@ddt.data(
+		("/home/user",
+		 "A stacktrace with /home/user/some/path/in/it, just to test",
+		 "A stacktrace with REDACTED/some/path/in/it, just to test"),
+
+		("C:\\Users\\User",
+		 "A stacktrace with C:\\Users\\User\\Some\\Path\\in\\it, just to test",
+		 "A stacktrace with REDACTED\\Some\\Path\\in\\it, just to test")
+	)
+	@ddt.unpack
+	def test_scrub_stacktrace_homedir(self, homedir, stacktrace, expected):
+		from octoprint.plugins.tracking import _scrub_trace
+
+		with mock.patch("os.path.expanduser") as mocked:
+			mocked.return_value = homedir
+			actual = _scrub_trace(stacktrace)
+
+		self.assertEqual(expected, actual)
```

foosel:
I've decided to rather look into making the sentry route ready for prime time. It has pretty much everything out of the box working already that would cause me a TON of effort to implement with the ELK variant, I don't want errors from everyone out there (I fear drowning in plugin errors caused by broken plugins instead of changes in OctoPrint on RC release) but rather ideally only people a) running RCs (which can be asked via a wizard to enable error tracking) or b) running into current issues (which can be asked to enable error tracking in bug reporting/support request templates).

PR #2998, opened 2019-01-11
Fixup temperature parsing for CR-10 S Pro

wez:
The temperature updates from this printer look like this in the terminal output:

```
Recv:  ==T:20.00 /0.00 ==B:21.02 /0.00 @:0 B@:0
```

`regex_temp` knows how to extract the data from this line, but there is some different logic that gates running the regex that needs to know what else to look for.  The result of this is that the temperature chart doesn't show any useful information for me.

This diff expands the gating logic to allow recognizing `==T:` as well as ` T:`, and that allows the temperature graph to function.


wez:
blurg, can't re-base the pr, will re-submit

PR #2999, opened 2019-01-11
Fixup actual temperature parsing for CR-10 S Pro

wez:
The temperature updates from this printer look like this in the terminal output:

```
Recv:  ==T:20.00 /0.00 ==B:21.02 /0.00 @:0 B@:0
```

`regex_temp` knows how to extract the data from this line, but there is some different logic that gates running the regex that needs to know what else to look for. The result of this is that the temperature chart doesn't show any useful information for me.

This diff expands the gating logic to allow recognizing `==T:` as well as `T:`, and that allows the temperature graph to function.

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -1721,6 +1721,7 @@ class MachineCom(object):
 
 				##~~ temperature processing
 				elif ' T:' in line or line.startswith('T:') or ' T0:' in line or line.startswith('T0:') \
+						or '==T:' in line or '==B:' in line \
 						or ((' B:' in line or line.startswith('B:')) and not 'A:' in line):
 
 					if not disable_external_heatup_detection and not self._temperature_autoreporting \


foosel:
I'm sorry, but I am not going to merge this.

As pointed out by me elsewhere already (can't find the link right now, will add if that changes) this is on Creality. They decided to break their firmware in a way that makes it incompatible to existing host software. If I start patching OctoPrint to accomodate single vendor decisions like this, I'll soon drown in if/else clauses in the code for all the other vendors out there expecting the same treatment. That becomes an unmaintainable mess. I have to draw the line somewhere to keep things manageable in the long run.

A plugin infrastructure is in place that allows to [rewrite the responses from the firmware](http://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-comm-protocol-gcode-received) in order to work around issues like these. It is pretty easy to write a plugin that fixes the CR-10s response without adding code to OctoPrint that will then have to be maintained for all eternity.

So please - create a plugin instead, register it on the repository and turn it into a win for everybody.

Issue #2996, opened 2019-01-10
Communication timeout when uploading to SD, m105 not being sent


gloomyandy:
#### What were you doing?
Attempting to upload a gcode file to the printers SD card. I have also seen a similar situation during a print, but the upload is easier to reproduce. Simply upload a very large (in this case 20MByte) gcode file using the "Upload to Sd" button.

#### What did you expect to happen?
Upload to complete with no error.
#### What happened instead?
Upload stopped with repeated messages "Communication timeout while printing, trying to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves."

Looking at serial.log it seems that the command to get a fresh response from the print (M105) is not being sent. I have also (once) seen a hang during printing and believe that this was a similar problem, but the upload to SD card is easier to reproduce.
#### Did the same happen when running OctoPrint in safe mode?
Yes

#### Version of OctoPrint
1.3.10

#### Operating System running OctoPrint
OctoPi 0.15.1

#### Printer model & used firmware incl. version
I3 clone running Marlin Bugfix 2.0.x on MKS Sbase 1.3

#### Browser and version of browser, operating system running browser
Chrome v 71.0.3578.98 on Windows 10

#### Link to octoprint.log
https://drive.google.com/open?id=1M0R8BszFdQutkDcE2RMU8Lcdb3Rxcdha

#### Link to contents of terminal tab or serial.log
https://drive.google.com/open?id=1-xSNTeoLwulpfu8wWkooH2FWJ4TifOZm

#### Link to contents of Javascript console in the browser
N/A

#### Screenshot(s)/video(s) showing the problem:
N/A

I have read the FAQ.

Please also see the following thread: https://discourse.octoprint.org/t/communication-timeout-when-uploading-to-sd-m105-not-being-sent/6136

foosel:
I was just about to comment on your proposed change related to that, then noticed that I wasn't in a PR, deleted the comment again and then realized that you haven't yet *created* a PR for the above commit.

So it became [a comment on the commit](https://github.com/gloomyandy/OctoPrint/commit/ba99c29650a7c6dc130d70fb443dd93147f6087b#r31943296) after all. Still happy to get a PR for that though :)

[...]

PR #3004, opened 2019-01-15
Use non blocking I/O for writes to implement write timeout (take 2)

gloomyandy:
Currently write timeouts are not really implemented (as they are set to 10000 seconds). The current code also has the unwanted side effect of causing the use of a select call (in the Linux PySerial library) even when a write does not need to block. This call to select can in some odd cases cause the write to block for the duration of the set timeout which in turn triggers a transmission error.

This change addresses both issues by allowing the use of the existing write retry code to handle timeouts and by using non-blocking write mode to avoid the use of select.

See the following discussion for details of various tests: https://discourse.octoprint.org/t/communication-timeout-when-uploading-to-sd-m105-not-being-sent/6136/13

https://discourse.octoprint.org/t/communication-timeout-when-uploading-to-sd-m105-not-being-sent/6136/13

#2996

#### Further notes
The current PR contains code to make each re-transmission attempt wait an increasingly long time before retrying. This may not strictly be required. But it seems better than simply battering against a possibly full queue again and again, as this will allow a (small) amount of time for the queue to be emptied by the printer. But feel free to amend this code as you see fit. The actual delays used (0.1s per retry) is pretty arbitrary.

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -2475,13 +2475,13 @@ class MachineCom(object):
 				serial_obj = serial.Serial(str(port),
 				                           baudrates[0],
 				                           timeout=read_timeout,
-				                           write_timeout=10000,
+				                           write_timeout=0,
 				                           parity=serial.PARITY_ODD)
 			else:
 				serial_obj = serial.Serial(str(port),
 				                           baudrate,
 				                           timeout=read_timeout,
-				                           write_timeout=10000,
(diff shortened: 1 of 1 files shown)


foosel:
This pull request has been mentioned on **OctoPrint Community Forum**. There might be relevant details there:

https://discourse.octoprint.org/t/communication-timeout-when-uploading-to-sd-m105-not-being-sent/6136/14

PR #3010, opened 2019-01-20
[WIP] Add system usage API endpoint

HarlemSquirrel:
This is a work in progress.

Adds an API endpoint for querying folder usage. Needed for #1030

Replaces #2891

### DIFF
--- a/src/octoprint/server/api/system.py
+++ b/src/octoprint/server/api/system.py
@@ -12,12 +12,28 @@ import threading
 from flask import request, make_response, jsonify, url_for
 from flask_babel import gettext
 
+import psutil
+
 from octoprint.settings import settings as s
 
 from octoprint.server import admin_permission, NO_CONTENT
 from octoprint.server.api import api
 from octoprint.server.util.flask import restricted_access, get_remote_address
 from octoprint.logging import prefix_multilines
+
+
+@api.route("/system/usage", methods=["GET"])
+@restricted_access
+@admin_permission.require(403)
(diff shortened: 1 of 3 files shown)


HarlemSquirrel:
I'm still having trouble setting up the test session properly. This endpoint requires an API key and valid session cookie.

foosel:
I'd honestly not test the endpoint through flask but rather test the underlying function directly. That way you can get rid of a ton of mocking and also don't run into the apikey issues you are having.

Your goal isn't to test that the flask routing works properly for the endpoint but rather that the endpoint returns the expected data for a mocked environment.

HarlemSquirrel:
OK, I've simplified the tests a bit but am running into `AttributeError: 'Flask' object has no attribute 'login_manager'`. Maybe I need to mock `flask_login`?

foosel:
My bad. I forgot that decorators replace the decorated function completely, so you still can't test it without the need of a flask context. Which is ... certainly going down a rabbit's hole that is way beyond what this test should be about. I pushed a commit with a small refactoring that makes stuff easier testable, and also fixes the test while at it. Please take a look and tell me if that looks ok for you.

HarlemSquirrel:
Ah, this is slick! I like it. :slightly_smiling_face:

foosel:
Great! So, do you want to still add something to this specific PR or is it mergeable?

HarlemSquirrel:
I think this works and is mergeable if you are happy.

PR #3045, opened 2019-02-11
I fixed a few typos in the German translation

akraus53:
This PR fixes a few typing and grammar mistakes in the German translation. I'm a native speaker and know German better than English, which might be noticeable. Of course it is not necessary to merge this but why wouldn't you?

This wasn't tested, because it's just a translation file and most of the time I only changed a few letters. 

I used a RegEx to extract all German translations and pasted them into MS Word, then I ran the spellchecker.


foosel:
Thank you! 👍

PR #3053, opened 2019-02-17
Fixed error in GCode viewer where multiple filaments available (e.g. …

AndyQ:
When multiple filaments are available (e.g. using a Prusa MMU2), but you are printing with say filament 4 only, the GCode Viewer throws errors as it loops over all filaments set on the layer - the unused ones are undefined and an exception gets generated).  This causes the layer height slider to not get updated correctly.

Tested on with Prusa MK3 with MMU2.

N/A
#### Further notes
Fix is just to add check if filament is not undefined before using it


foosel:
Well spotted, thanks for the fix! 👍

PR #3055, opened 2019-02-18
Added ability to preview the current layer to be drawn

AndyQ:
This PR adds an additional option to then GCodeViewer to allow the current layer to be previewed as well as the next and previous layers.  If selected, it draws the full preview of the current layer at a 30% opacity and will then draw the progress over the top.  This allows the user to see how much of the current layer is remaining.

It was tested on a Raspberry PI with a Prusa MK3 printer and has been used on over 15 prints with no issues  so far.

I made this change as I like seeing how much left of a layer is remaining - especially if I'm doing an additional manual filament change and want to pause at a specific point.


foosel:
Thanks! 👍

PR #3066, opened 2019-02-28
Add missing documentation about state in `jobs` endpoint

fieldOfView:
* checklist checked

This PR fixes the documentation of the `job` api endpoint. The current documentation omits the `state` member of the returned output. 

It is a documentation fix, no functional changes were made.

#### Further notes
The `state` member of the returned output is a string, not a structure. I am not sure if my additions to the documentation reflect that properly.

### DIFF
--- a/docs/api/datamodel.rst
+++ b/docs/api/datamodel.rst
@@ -219,6 +219,25 @@ Progress information
      - Integer
      - Estimate of time left to print, in seconds
 
+.. _sec-api-datamodel-jobs-state:
+
+State information
+-----------------
+
+.. list-table::
+   :widths: 15 5 10 30
+   :header-rows: 1
+
+   * - Name
+     - Multiplicity
+     - Type
+     - Description
+   * - body
+     - 1
+     - String
+     - A textual representation of the current state of the job, i.e. "Operational", "Printing", "Pausing", "Paused",
+       "Cancelling", "Error" or "Offline".
+
 .. _sec-api-datamodel-files:
 
 File related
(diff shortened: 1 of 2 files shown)


foosel:
Small issue: the way it is currently documented suggests a structure like

```
{ 
    "job": { ... },
    "progress": { ... },
    "state": {
        "body": "Operational"
    }
}
```

instead of the correct:

```
{ 
    "job": { ... },
    "progress": { ... },
    "state": "Operational"
}
```

Just putting what's now under the state data model right into the job data model would fix that. 

Larger issue: I'm unsure right now why the state is even in there and if it needs to be in there, considering it's the state of the *printer*  and not the job that's being reported here. Then again, it's been that way since at least 2014, so I guess it's time to grandfather it in...

Ah well... happy to merge this then if the data model is less confusing (see "small issue" above) - will you take care of that or shall I? :)

fieldOfView:
I could add a note that this `state` is deprecated?

foosel:
I meant as reflected in 4dbb24e - your latest change just turned it into

``` 
{ 
    "job": { ... },
    "progress": { ... },
    "state": {
        "state": "Operational"
    }
}
```

which would still be wrong. 

Documenting it and immediately marking it deprecated strikes me as odd. Let's got with it as is. It's not the full state object at least but only the textual representation, so it's not as misplaced as it could be, and I fear there are probably a ton of clients already depending on it, even though it was undocumented.

fieldOfView:
Doh, now I get it. Thanks.

probonopd:
Thank you very much for this correction @fieldOfView @foosel. Was driving me insane before, this commit made my day!

PR #3106, opened 2019-04-04
README: Use explicit python version for virtualenv

hlovdal:

This is actually a pull request explicitly targeting `master`.

The current `README.md` on the `master` branch (which is the one people see when landing on the github page) suggest running `virtualenv venv` which will use python 3 when run on Fedora 29 and not create a usable environment.

The document do mention that only version 2.7 is supported, but it is not obvious that that requires an extra argument to virtualenv so it is best to show this explicitly.

I used that command when trying to run octoprint from the git repository.


foosel:
In that case let's ignore the bot and just merge this. Thanks!

PR #3134, opened 2019-04-28
Remote user support

tduehr:

Adds basic support for users supplied by a reverse proxy (e.g., Apache, nginx) and automatically creating them if not found.

I'm not sure how to test it thoroughly, as I don't do much Python development. I'll be testing it against our SSO system. It should be testable the same way the Basic Authentication functionality is tested.

Users want an increasing variety of authentication methods. Adding support for remote users opens up many of these (e.g., SAML, CAS, etc.) by pushing authentication to the reverse proxy (e.g., to Apache via `mod_auth_cas`). 

#### Further notes

This is only basic support, many remote user plugins (e.g., `mod_auth_cas`) support roles or groups as well.

### DIFF
--- a/docs/configuration/config_yaml.rst
+++ b/docs/configuration/config_yaml.rst
@@ -65,8 +65,8 @@ Use the following settings to enable access control:
      # Whether to trust remote user headers. If you have setup authentication in front of
      # OctoPrint and the user names you use there match OctoPrint accounts, by setting this to true users will
      # be logged into OctoPrint as the user provided in the header. Your should ONLY ENABLE THIS if your
-     # OctoPrint instance is only accessible through a connection locked down through an authenitating reverse proxy!
-     trustBasicAuthentication: false
+     # OctoPrint instance is only accessible through a connection locked down through an authenticating reverse proxy!
+     trustRemoteUser: false
 
      # Header used by the reverse proxy to convey the authenticated user.
      remoteUserHeader: REMOTE_USER
--- a/docs/configuration/config_yaml.rst
+++ b/docs/configuration/config_yaml.rst
@@ -62,6 +62,18 @@ Use the following settings to enable access control:
      # header and login the user without further checks. Use with caution.
      checkBasicAuthenticationPassword: true
 
+     # Whether to trust remote user headers. If you have setup authentication in front of
+     # OctoPrint and the user names you use there match OctoPrint accounts, by setting this to true users will
+     # be logged into OctoPrint as the user provided in the header. Your should ONLY ENABLE THIS if your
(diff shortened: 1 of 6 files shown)


foosel:
Hi @tduehr, can you please take a look at the failing tests before I do a review on this? Thanks!

tduehr:
Done.

docs/configuration/config_yaml.rst
```diff
@@ -62,6 +62,18 @@ Use the following settings to enable access control:
      # header and login the user without further checks. Use with caution.
      checkBasicAuthenticationPassword: true
 
+     # Whether to trust remote user headers. If you have setup authentication in front of
+     # OctoPrint and the user names you use there match OctoPrint accounts, by setting this to true users will
+     # be logged into OctoPrint as the user provided in the header. Your should ONLY ENABLE THIS if your
+     # OctoPrint instance is only accessible through a connection locked down through an authenitating reverse proxy!
+     trustBasicAuthentication: false
```
foosel: There seems to be a documentation error here, this should say `trustRemoteUser`.

docs/configuration/config_yaml.rst
```diff
@@ -62,6 +62,18 @@ Use the following settings to enable access control:
      # header and login the user without further checks. Use with caution.
      checkBasicAuthenticationPassword: true
 
+     # Whether to trust remote user headers. If you have setup authentication in front of
+     # OctoPrint and the user names you use there match OctoPrint accounts, by setting this to true users will
+     # be logged into OctoPrint as the user provided in the header. Your should ONLY ENABLE THIS if your
+     # OctoPrint instance is only accessible through a connection locked down through an authenitating reverse proxy!
```
foosel: `authenitating` => `authenticating`

src/octoprint/server/__init__.py
```diff
@@ -1653,6 +1659,10 @@ def _setup_login_manager(self):
 		if self._settings.get(["accessControl", "trustBasicAuthentication"]):
 			loginManager.request_callback = load_user_from_request
 
+		# Login via REMOTE_USER header
+		if self._settings.get(["accessControl", "useRemoteUser"]):
```
foosel: This should probably also say `trustRemoteUser`.

tduehr: There's probably a better way to do this than I did but I'm not the best with Python.

tduehr:
That should work now. Still probably a better way to do it...

foosel:
I've rebased this on the `maintenance` branch since it's more of an improvement of existing functionality than full blown new functionality and this way it'll make it into 1.3.12.

PR #3146, opened 2019-05-11
Add autocofus to the username in the login plugin

esver:

When I want to login, I need to focus to the username entry manually.
With autofocus, I can login more faster using only keyboard.

Tested on Firefox on my computer and on Firefox on Android.
Connect to Octoprint and the focus is automaticaly to the username entry.


esver:
Moved this to a separate branch, and re-created the pull request.

PR #3167, opened 2019-05-18
comm: Don't open twice on FreeBSD

kevans91:
This PR stops opening the serial port twice on FreeBSD. Further details in the background context. 

Tested by opening/closing my serial port periodically. An interested reviewer could test with a FreeBSD snapshot, installing OctoPrint with this patch applied, and doing the same. =)

The primary motivation for this change is that the FreeBSD CH340 driver explicitly rejects parity setting due to a notion once upon a time that it simply didn't work with every chip. While it would be trivial to fix the driver to support parity setting, OctoPrint doesn't actually need it and we don't need to be doing this workaround for a Linux/Debian bug.

N/A

N/A

#### Further notes
N/A


foosel:
Merged and slightly modified and turned into a setting:

image

The "detect" option will only make it active if OctoPrint is running on a Debian GNU/Linux (`get_os() == "linux" and os.path.exists("/etc/debian_version")`)

kevans91:
Awesome, thanks!

Issue #2896, opened 2018-11-15
[Request] Confirmation dialog when removing a plugin


yet-another-average-joe:
Hello, and thanks for this great piece of software !

A confirmation dialog box would be usefull when deleting or even deactivating a plugin.

ntoff:
Removing one I can see a confirmation being handy, but do you really need one when de-activating? I'm pretty sure there are warnings when you try to disable one of the system critical plugins, but as far as disabling 3rd party plugins, if you accidentally disable one you don't want to disable, just re-enable it.

Also if you're removing multiple plugins, a dialog for each would be super annoying so some way to suppress it, or even select multiple plugins for mass removal, would need to be thought up first.

ntoff:
Then perhaps you should post an issue to that plugin's github page, to see if the issue can be fixed.

yet-another-average-joe:
Yes, I did it already. It is TouchUI. I found a solution that works, at least for me.

A question : could a plugin display its own warning dialog box when the user flips the activation switch, or hits the delete button ? Can a plugin do something, or is it completely passive ? I searched the OctoPrint Wiki for some sort of deactivate/uninstall event that a plugin could handle, but I found nothing...

dovecode:
I think we can close this issue now that the changes are in `maintenance`, right (cf. 79558fd)?

foosel:
Actually, I'll mark it as solved and close it once 1.3.12 (the release that includes that change) is released.

dovecode:
Ah, ok. That makes sense. Thanks.

foosel:
1.3.12 has been released.

yet-another-average-joe:
Thanks for this enhancement !

PR #3179, opened 2019-05-25
Added confirmation dialog before uninstalling plugin

dovecode:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Adding (optional) confirmation dialogs before uninstalling and disabling plugins. The former is to avoid accidentally uninstalling plugins. The latter because some (granted, broken) plugins behave badly when disabled.

The behavior is controlled with settings in the plugin manager and the default setting for both is "don't ask", i.e. the current behavior, in order to minimize disruption to existing users.

Tested by manual tests of the following use cases:

1. Start new server without changing settings and verify that neither disabling nor uninstalling results in a prompt (test defaults)

2. Change to enable prompting for uninstall and verify that the user is prompted when attempting to uninstall a plugin. Testing both branches (answering negatively leaves plugin, answering positively uninstalls)

3. Restart server, restart browser and try previous test case (tests correct persistence and reload of setting)

4. Change to enable prompting for disabling and verify that the user is prompted when attempting to disable a plugin. Testing both branches (answering negatively leaves plugin enabled, answering positively disables)

5. Re-enable plugin to verify that this path does not result in a prompt.

6. Restart server, restart browser and try previous test cases (tests correct persistence and reload of setting)

7. Disable both settings and test immediate results (no prompt for any action).

8. Restart server and browser and repeat previous test.

Note that several of the test cases run with `nosetests --with-doctest` do fail, but these are existing issues (same failures with a brand new `devel` branch checked out. This PR does nor fail more than `devel` ;)

Output from the test suite: `E.E....EEE..E..EF`

Most of the changes (apart from persisting and loading settings) are UI changes. I don't believe we have a testing framework for this, so a lot of manual testing was necessary. I recommend commenting out the uninstall code and add a log statement when testing this to save time reinstalling plugins when testing the uninstall prompting logic.

I don't personally need the disable prompt, but it was mentioned in the linked ticket below so I included it, as it was a very simple addition.

As mentioned, I left defaults to not prompt so as to avoid disrupting existing users, but strictly speaking, the more robust solution would be to set the uninstall prompt to be enabled by default.

When I added the checkboxes for the settings, I added them at the top of the setting dialog for the plugin manager. This is because I don't believe the current caveat at the top ("You should normally not have to change any of the following settings, they are purely provided for convenience here.") apply to these settings -- they are purely user preferences. I am open for better ways to accomplish this.

#2896 

##### Uninstallation prompt

Uninstallation prompt

##### Disabling prompt

Disabling prompt

##### iNew configuration dialog for the plugin manager

iNew configuration dialog for the plugin manager

#### Further notes

This PR adds screen prompts. I am unclear from the contribution guide as to whether I need to add them to a translation file anywhere, so I left the existing files unchanged. Do let me know if I need to take action on that.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -102,6 +102,7 @@ date of first contribution):
   * [Jack Wilsdon](https://github.com/jackwilsdon)
   * [Ryan Finnie](https://github.com/rfinnie)
   * [Timur Duehr](https://github.com/tduehr)
+  * [Nicolai Nielsen](https://github.com/dovecode)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/plugins/pluginmanager/__init__.py
+++ b/src/octoprint/plugins/pluginmanager/__init__.py
@@ -120,6 +120,8 @@ class PluginManagerPlugin(octoprint.plugin.SimpleApiPlugin,
 		self._repository_cache_ttl = self._settings.get_int(["repository_ttl"]) * 60
 		self._notices_cache_path = os.path.join(self.get_plugin_data_folder(), "notices.json")
 		self._notices_cache_ttl = self._settings.get_int(["notices_ttl"]) * 60
+		self._confirm_uninstall = self._settings.global_get_boolean(["confirm_uninstall"])
+		self._confirm_disable = self._settings.global_get_boolean(["confirm_disable"])
 
 		self._pip_caller = LocalPipCaller(force_user=self._settings.get_boolean(["pip_force_user"]))
 		self._pip_caller.on_log_call = self._log_call
@@ -157,6 +159,8 @@ class PluginManagerPlugin(octoprint.plugin.SimpleApiPlugin,
 			notices_ttl=6*60,
 			pip_args=None,
 			pip_force_user=False,
+			confirm_uninstall=True,
+			confirm_disable=False,
 			dependency_links=False,
 			hidden=[]
 		)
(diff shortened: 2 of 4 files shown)


foosel:
> Note that several of the test cases run with `nosetests --with-doctest` do fail, but these are existing issues (same failures with a brand new `devel` branch checked out. This PR does nor fail more than `devel` ;)

That's weird, they do not seem to fail in the actual CI build (see the green checkmark by Travis) nor does `devel` fail here...

> As mentioned, I left defaults to not prompt so as to avoid disrupting existing users, but strictly speaking, the more robust solution would be to set the uninstall prompt to be enabled by default.

I'd actually suggest to enable confirmation for uninstall by default - if a user has an issue with the confirmation they'll survive having to do three clicks to get back the old behaviour and it will indeed be the better solution for the majority of users out there.

> I am unclear from the contribution guide as to whether I need to add them to a translation file anywhere

You don't have to do that manually, since you marked up the translatable strings already, they will be found by `python setup.py babel_refresh` and then be included in the translation files. But thanks for thinking about that!

Finally, since this is an improvement for existing functionality instead of a fully new feature, I'd also suggest to file this PR against the `maintenance` branch so that it can be included in the next maintenance release 1.3.12. 

(I also noticed that the PR check list still lacked that distinction of `maintenance` vs `devel` and have adjusted it accordingly)

dovecode:
I suppose there's an issue with my setup that causes tests to fail. Is there a part of the setup guide about this that I've missed (went through the setup in http://docs.octoprint.org/en/master/development/index.html without seeing any mention specifically on the test suite). But I suppose as long as Travis is happy, we're good!

I'll switch the default. As I mentioned, I think that's a more sensible default -- I just weren't sure the convention in this project of weighing sensible defaults against user impact.

Wrt. switching to `maintenance`, as a Github newbie (I'm more used to Gitlab), is it enough for me to switch the PR to `maintenance` or should I rebase, force push and then switch to `maintenance`?

foosel:
Just changing the PR to another base branch will probably add all commits between `maintenance` and `devel` by me to the PR as well. Rebasing just your commit on `maintenance`, force pushing and then switching will yield better results.

dovecode:
OK, took a few pulls, pushes, and rebases to get this PR changed to `maintenance`, but I managed it without messing up the git history. The extra push was me realizing my `maintenance` branch fork had become out of date with yours, but it should be all good now.

foosel:
Looks like it, thank you!

PR #3180, opened 2019-05-26
Added button to terminal window to all log window to be cleared out

AndyQ:
This PR adds an addition button "Clear log window" to the terminal screen which allows the contents of the terminal window to be cleared out.
Whilst debugging other plugins, the logs can become quite tricky to find a specific place so its often useful to clear the existing logs before doing something which makes finding things a little simpler.

Populate Terminal log window either with output from Virtual printer or real connected printer.  Click the "Clear log window" button.  Existing log output should be cleared.

N/A
N/A
N/A

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/terminal.js
+++ b/src/octoprint/static/js/app/viewmodels/terminal.js
@@ -338,6 +338,14 @@ $(function() {
             copyToClipboard(lines.join("\n"));
         };
 
+        self.clearLogWidow = function() {
+            if (self.fancyFunctionality()) {
+                self.log([]);
+            } else {
+                self.plainLogLines([]);
+            }
+            
+        };
         // command matching regex
         // (Example output for inputs G0, G1, G28.1, M117 test)
         // - 1: code including optional subcode. Example: G0, G1, G28.1, M117
--- a/src/octoprint/static/js/app/viewmodels/terminal.js
+++ b/src/octoprint/static/js/app/viewmodels/terminal.js
@@ -338,13 +338,9 @@ $(function() {
             copyToClipboard(lines.join("\n"));
         };
 
-        self.clearLogWidow = function() {
-            if (self.fancyFunctionality()) {
-                self.log([]);
-            } else {
-                self.plainLogLines([]);
-            }
-            
+        self.clearAllLogs = function() {
+            self.log([]);
+            self.plainLogLines([]);            
         };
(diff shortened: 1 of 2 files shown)


src/octoprint/templates/tabs/terminal.jinja2
```diff
@@ -13,7 +13,8 @@
         <span data-bind="visible: !autoscrollEnabled() && enableFancyFunctionality()" style="display: none">(<a href="javascript:void(0)" data-bind="click: scrollToEnd">{{ _("Scroll to end") }}</a>)</span>
     </small>
     <small class="pull-right">
-        <a href="javascript:void(0)" data-bind="click: copyAll"><i class="fa fa-copy" title="{{ _('Copy all') }}"></i> {{ _('Copy all') }}</a>
+        <a href="javascript:void(0)" data-bind="click: copyAll"><i class="fa fa-copy" title="{{ _('Copy all') }}"></i> {{ _('Copy all') }}</a><br>
+        <a href="javascript:void(0)" data-bind="click: clearLogWidow"><i class="fa fa-copy" title="{{ _('Clear log widow') }}"></i> {{ _('Clear log widow') }}</a>
```
foosel: Typo ;) "widow" -> "window". I'd actually turn it into just "Clear all" though, at least for the link text itself and leave the title as is.

src/octoprint/static/js/app/viewmodels/terminal.js
```diff
@@ -338,6 +338,14 @@ $(function() {
             copyToClipboard(lines.join("\n"));
         };
 
+        self.clearLogWidow = function() {
```
foosel: Wouldn't it make more sense to always clear both the plain and fancy version of the log? Otherwise things could get weird if you swap between the two...

src/octoprint/templates/tabs/terminal.jinja2
AndyQ: Totally - no idea how on earth I didn't spot that - been clicking that button many many times and just skipepd over it! Fixed and changed link to ClearAllLogs

src/octoprint/static/js/app/viewmodels/terminal.js
AndyQ: Indeed it would - Fixed!

foosel:
👍 Did a tiny icon & wording change and merged.

Issue #3075, opened 2019-03-05
[Request] Filament usage should take into account M605 duplication/mirroring commad


JanneMantyharju:
**Is your feature request related to a problem? Please describe.**
At least Marlin firmware supports M605 command on printers with two printing heads to duplicate or mirror on second tool what the first tool is printing. OctoPrint does not currently recognize this and thus the filament counter plugins can't take into account the filament usage on second tool

**Describe the solution you'd like**
When M605 S2, S4, S5 or S6 command is seen, filament usage of tool 0 should be added to tool 1 usage

foosel:
Where can one find a full documentation of this command and what the available parameters do exactly on what kind of firmware? [The RepRap wiki](https://reprap.org/wiki/G-code#M605:_Set_dual_x-carriage_movement_mode) doesn't go into a lot of detail.

JanneMantyharju:
Good question. I found this information from Marlin source code: 

> 
> // There are a few selectable movement modes for dual x-carriages using M605 S<mode>
> //    M605 S0: Full control mode. The slicer has full control over x-carriage movement
> //    M605 S1: Auto-park mode. The inactive head will auto park/unpark without slicer involvement
> //    M605 S2 [Xnnn] [Rmmm]: Duplication mode. The second extruder will duplicate the first with nnn
> //                         millimeters x-offset and an optional differential hotend temperature of
> //                         mmm degrees. E.g., with "M605 S2 X100 R2" the second extruder will duplicate
> //                         the first with a spacing of 100mm in the x direction and 2 degrees hotter.
> //
> //    Note: the X axis should be homed after changing dual x-carriage mode.
> //    M605 S3: Default mode
> //	  M605 S4: Duplication Mirror mode
> //    M605 S5 [Xnnn] [Rmmm]: Duplication Raft mode
> //	  M605 S6: Duplication Raft Mirror mode
> //
> //    Note: the X axis should be homed after changing dual x-carriage mode.

foosel:
That's sadly not very specific - as someone without access to a mirror setup these different modes tell me pretty much nothing and I'm not sure how to handle them with regards to filament usage calculation.

JanneMantyharju:
After M605 S2, S4, S5 or S6 are issued, from that command onwards second extruder does exactly what the first extruder does. Any other M605 SX command stops second extruder duplicating first extruder extrusion commands.

[...]

PR #3181, opened 2019-05-26
Calculate filament usage for all extruders is M605 duplicate/mirror c…

JanneMantyharju:
…ommand is used (Issue #3075)

Some printers with more than one extruder support M605 command to duplicate or mirror what first extruder does. This change calculates the filament usage correctly for all extruders

Write M605 S2 to the beginning of gcode file (enable duplication mode). Do analysis and check that the filament usage is the same for all extruders.

#3075


foosel:
Hey, thank you for this PR! Since this is more an improvement on existing functionality instead of a full blown new big feature, I think it would be better suited to the `maintenance` branch so it can go into the next maintenance release 1.3.12. Could you rebase your PR against `maintenance`? :)

PR #3209, opened 2019-06-26
[WIP] Asynchronous serial

### LINKED ISSUE #2834: [Brainstorming] Making use of ADVANCED_OK in MarlinFirmware to speed up serial
I've been experiencing slowdowns in printing small complex models due to octoprint not streaming commands to the printer fast enough to stop the buffer for emptying. see: #2799 

I've been following #2346 in hopes that it will provide a solution, but it's obvious that the upstream issues with software flow control in the linux usb-serial stack are not something that are likely to be fixed soon, and it's not something i feel competent to take a look at myself.

I have however come across a feature in MarlinFirmware that looks like it could provide a solution: ADVANCED_OK
This feature changes the format of 'ok' messages to include two new fields:

- P: the remaining space in the move planner buffer
- B: the remaining space in the unprocessed command buffer

With this information I believe we should be able to send multiple commands to the printer between ok responses as long as we don't send more than the available unprocessed buffer.

I've had a play and done a hacky implementation in comm.py and the initial results seem great, I've been testing purposfully over detailed gcode that always stalled with the normal communications code and I've been unable to drain the planned buffer regardless of complexity or print speed. It also increased sd transfer speeds although not enough to make it quicker than just unplugging and copying.

So far my implementation is hack and unsuitable for a pull request, and i've only tested with a pi3 on my printer at 250000 baud rate and a fairly large buffer, however if this is something you'd consider adding I'd be happy to spend the time to turn it into an actual feature and PR it.

Things to think about:
- Feature detection: Marlin does't currently report ADVANCED_OK in it's M115 response, however the presence of the P and B parameters in the ok responses is something we could check for along with a setting in the connection settings to manually turn on and off (we can't assume the presence of the P and B parameters anyway as not every ok message includes them even with the feature turned on). Alternatively, maybe MarlinFirmware would accept a PR to add it.
- Resends seem to work from my limited testing, however with a large buffer on the printer it's possible to have sent more commands than the history contains which can kill a print if we recieve a resend request for one of those commands.
- Implementation: The code I've been plaing with so far just skips calling _clear_to_send.clear() if we've sent less than B (space in the unprocessed buffer) commands since the last ok. it's current form is hacky and maybe there is a better way to do it using the functinality of CountedEvent (_clear_to_send), I'm really open to suggestions, python isn't my native programming language so I'm a bit unfamiliar with the tools at my disposal.

Sorry for the wall of text, I'd be glad to hear if this is something useful.

linvinus:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Resolve #2834 https://github.com/foosel/OctoPrint/issues/2834

compile marlin with following options
```
#define BAUDRATE 500000
#define BLOCK_BUFFER_SIZE 64
#define BUFSIZE 32
#define TX_BUFFER_SIZE 32
#define RX_BUFFER_SIZE 2048
#define ADVANCED_OK
```
print some cylinder at maximum printing speed, compare print quality 

no
#2834


linvinus:
looks like travis still using my previous branch. error on line 3170 but in my repository this line is different https://github.com/linvinus/OctoPrint/blob/5b6bd5a7f53bf815a8eef0cc1b73ee0dbd7d189b/src/octoprint/util/comm.py#L3170

linvinus:
@foosel now working good with maintenance branch

src/octoprint/util/comm.py
```diff
@@ -2825,11 +2858,13 @@ def _get_next_from_job(self):
 			if isinstance(self._currentFile, StreamingGcodeFileInformation):
 				self._finishFileTransfer()
 			else:
-				self._changeState(self.STATE_FINISHING)
-				self.sendCommand("M400", part_of_job=True)
-				self._callback.on_comm_print_job_done()
 				def finalize():
-					self._changeState(self.STATE_OPERATIONAL)
+					self._changeState(self.STATE_FINISHING)
```
foosel: Why the double decoupling here? And the change in the first place? Feels unrelated to the PR.

src/octoprint/util/comm.py
```diff
@@ -2213,10 +2222,31 @@ def _handle_ok(self):
...
+
+			parsed = parse_advanced_ok_line(line)
+			if parsed:
+				self._AdvancedOkSupported = True
+				self._LastProcessedLine = parsed.get("N")
+				free_planner_buff   = parsed.get("P")
+				free_command_buff   = parsed.get("B")
+
+				if (self._LastProcessedLine >= self._SynchronousCommand ):
+					self._AdvancedOkSendNextLines = self._LastProcessedLine + max(free_planner_buff,free_command_buff)
```
foosel: We should always leave 1 slot free in the command buffer for emergency codes (e.g. M112).

So this should probably rather be 

``` python
min(max(free_planner_buff, free_command_buff - 1), 0)
```

src/octoprint/util/comm.py
```diff
@@ -3122,9 +3157,8 @@ def _send_loop(self):
 		sending (through received ``ok`` responses from the printer's firmware.
 		"""
 
-		self._clear_to_send.wait()
-
-		while self._send_queue_active:
+		def _send_one_from_queue():
```
foosel: I think this merits an extraction into a proper member function of the class.

It also lacks documentation of what the return values `True` and `False` signify. This needs to be changed, otherwise maintenance further down will become a nightmare. E.g.

``` python
def _send_one_from_queue(self):
    """
    Attempts to send one item from the send queue.

    Returns:
        (bool) True if no line was sent, False otherwise.
    """

    # ...
```

Side note: If the above logic is correct, it feels a bit backwards. I'd expect a method called `send_one_from_queue` to return `True` if it actually *does* send something, and `False` if not, so if I understand this correctly, this needs to be reversed.

src/octoprint/util/comm.py
```diff
@@ -3240,10 +3274,55 @@ def _send_loop(self):
 					# are done processing the last fetched queue entry
 					self._send_queue.task_done()
 
-				# now we just wait for the next clear and then start again
-				self._clear_to_send.wait()
+					if ( self._AdvancedOkSupported ):
```
foosel: This whole section needs comments because otherwise in 6 months whoever ends up maintaining this (probably me) will have *no* idea anymore what is going on here and why the conditions for the `if`s are defined as they are.

src/octoprint/util/comm.py
```diff
@@ -3240,10 +3274,55 @@ def _send_loop(self):
...
+						if ( self._state == self.STATE_PRINTING
```
foosel: What about `STATE_PAUSING`, `STATE_RESUMING`, `STATE_CANCELLING`? Printing state is more than just `STATE_PRINTING`.

src/octoprint/util/comm.py
```diff
@@ -3240,10 +3274,55 @@ def _send_loop(self):
...
+			if not self._send_queue_active:
+				break
+
+			if ( self._AdvancedOkSupported ):
+				if (self._send_queue.resend_active
+					or self._SynchronousCommand >0
+					or self._AdvancedOkSendNextLines < 0
+					):
+					# now we just wait for the next clear and then start again
+					while True:
```
foosel: This feels *very* dangerous and is a huge deadlock waiting to happen. It *shouldn't* happen as long as sensible data is being exchanged, but that's not something we can count on with finicky cabling and controllers. This needs *some* kind of fallback to ensure we don't get stuck in this loop forever.

linvinus:
@foosel hi, i have made a little improvement - clear send queue on canceling.
Should I hand this over to my branch?

src/octoprint/util/comm.py
```diff
@@ -3240,10 +3274,55 @@ def _send_loop(self):
...
+						self._clear_to_send.wait()
+						if (self._LastProcessedLine >= self._SynchronousCommand):
+							break
+				else:
+					cruLineTimeout=200 * self._serial.timeout #wait untill busy message will be recieved
```
foosel: What is `cruLineTimeout` supposed to mean? Typo of `cur`? Write out words fully please, for readability, and in snake_case => `current_line_timeout`.

src/octoprint/util/comm.py
```diff
@@ -3240,10 +3274,55 @@ def _send_loop(self):
...
+					while( self._currentLine > self._AdvancedOkSendNextLines and cruLineTimeout >0):
```
foosel: I'm completely unclear on what this whole block does. Needs more comments ;)

And busy polls are always a red flag - do we really need that here? Comments might clear up why, if so.

src/octoprint/util/comm.py
```diff
@@ -4641,6 +4720,26 @@ def parse_resend_line(line):
 
 	return None
 
+def parse_advanced_ok_line(line):
+	"""
+	Parses the provided marlin ADVANCED_OK line.
+	Args:
+		line (str): the line to parse
+	Returns:
+    last_processed_line
```
foosel: From the looks of it this should rather be

``` python
def parse_advanced_ok_line(line):
    """
    Parses the provided ADVANCED_OK line.

    Format:
        ok N<last processed line> P<free planner slots> B<free command slots>

    Args:
        line (str): the line to parse

    Returns:
        (dict or None) A dict containing the last processed line number (``N``), the amount of free 
        slots in the motion planner buffer (``P``) and the amount of free slots in the command buffer (``B``), 
        or None if no match could be made.
    """ 
    # ...
```

And something about the whitespace seems to be fishy in the function as well.

foosel:
> 
> 
> @foosel hi, i have made a little improvement - clear send queue on canceling.
> Should I hand this over to my branch?

That sounds like it should be a separate PR. Let's concentrate on what we have here first.

src/octoprint/settings.py
```diff
@@ -114,6 +114,7 @@ def settings(init=False, basedir=None, configfile=None):
 		"additionalPorts": [],
 		"additionalBaudrates": [],
 		"longRunningCommands": ["G4", "G28", "G29", "G30", "G32", "M400", "M226", "M600"],
+		"synchronousCommands": ["M109","M190","G4", "G28", "G29", "G30", "G32", "M400", "M226", "M600"],
```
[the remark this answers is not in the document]
linvinus: i'm working only with marlin 1.1.9-bugfix
it have 3 buffer layers
1) RX_BUFFER_SIZE - 2048bytes. in my configuration, it is incoming serial buffer filled  from serial interface interrupts

2) BUFSIZE - 32 commands. it contain unparsed lines (simply Gcode line), but checksum already checked 

3) BLOCK_BUFFER_SIZE - 64 commands. it is planner buffered commands (actual buffered commands in terms of reprap )

both BUFSIZE and BLOCK_BUFFER_SIZE processing in main loop as follow:

```

void loop() {
  void get_available_commands() {
    inline void get_serial_commands() {
      while (commands_in_queue < BUFSIZE && (c = MYSERIAL0.read()) >= 0) {
      }
  }
  process_next_command(){
    char * const current_command = command_queue[cmd_queue_index_r];
    parser.parse(current_command);
    process_parsed_command(){
               buffer_line() 
               planar.buffer_segment() {
                      block_buffer[BLOCK_BUFFER_SIZE]
                  }
      ok_to_send();
    }
  }
  --commands_in_queue;

  endstops.event_handler();
  idle(){
    //lcd_update();
    host_keepalive();
    thermalManager.manage_heater();
    thermalManager.auto_report_temperatures();
  }
}

```

"synchronous commands" is designed to prevent overflow in RX_BUFFER_SIZE, just clear this list and try to print, very quickly you will get resend request

as alternative we could send asynchronously only reprap "buffered commands"

src/octoprint/util/comm.py
linvinus: because of asynchronous transmission, at the moment of read end-of-file there will be a lot of commands in queue, actual printing state must be synchronized with printing queue

src/octoprint/util/comm.py
linvinus: i see the problem but your solution will cause overflow, i will think about that

src/octoprint/util/comm.py
linvinus: this code filling the send queue with respect to marlin buffers, in non printing state older synchronous mode should be working fine

src/octoprint/util/comm.py
linvinus: you are right, self._advanced_ok_last_line must be retested to -1 at communication timeout, if we loose synchronization

src/octoprint/util/comm.py
linvinus: sorry haven't enough time now, simply in asynchronous transaction mode, synchronization through _clear_to_send queue is not working at all.

src/octoprint/settings.py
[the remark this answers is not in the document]
linvinus: @foosel hi!
synchronousCommands - is a commands on which marlin may not send "ok" immediately, it will send busy or wait but after some timeout

imagine we have following gcode file
```
M80 ; power on
M190 S110
M109 S270
G1 ....
>100 lines of G0-G1 commands, each line is about 86 bytes long
```
after M80 we will get `ok P63 B31`
so marlin say that it ready to receive up to 94 commands, but
if we will send 94 lines from gcode file marlin will not accept them, instead it will stuck at M190 and will waiting for heating complete. at the same time in that moment it will not send to us nor ok nor busy nor wait, it will send busy later after timeout.

also, because of restricted RX_BUFFER_SIZE, when we send 94 * 86 = 8084 bytes RX_BUFFER will be overflowed, some command will be corrupted and marlin will send resend request, which will worsen print quality. 

so after send `M190 S110` we will not have any feedback to continue send commands asynchronously.

I think this is bug in marlin protocol but right now we can't resolve it, that is why we need special list with such commands no which marlin will not answer immediately.

this list is pretty close to `longRunningCommands`, i don't know could we merge this two lists or probably better create another one - `bufferedCommans` (instead synchronousCommands) which will contain buffered  commands according to reprap documentation, and send asynchronously only commands from it but as soon as send one command not from bufferedCommands stop sending asynchronously and continue sending synchronously.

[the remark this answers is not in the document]
linvinus: >Buffered commands should usually only be G0-G3 and G28-G32 (see also here).
i disagree, as i describe  earlier G28-G32 may not answer immediately.

also   following commands working well asynchronously
```
M420 S1 Z1.0 ; Load auto bed leveling data from epprom
M900 K0.025; titan aero 150
M900 S0 ; activate
M203 X500.00 Y500.00 Z5.00 E25.00 ;Max Feedrate
M204 P2000 R1000 T3000 ;Default Acceleration
M205  J0.1 E15
M201 X5000 Y5000 E5000 ;Max Acceleration

G92 E0 ; Reset Extruder
```

so list of bufferedCommands may be much longer than my current `synchronousCommands`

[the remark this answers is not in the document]
foosel: > i disagree, as i describe earlier G28-G32 may not answer immediately.

No buffered command will answer immediately if the planner buffer is full. And I was going off of what I found across firmware.



> `bufferedCommands` in my understanding is synonym of asynchronous commands, correct me if i'm wrong

As far as I understand your interpretation of "asynchronous commands", yes.

I'd got with a list of buffered/asynchronous commands instead of the opposite since it's usually the shorter one (every command by default is unbuffered/synchronous unless it explicitly touches the planner buffer) and more likely to match across firmware variants. Remember that Marlin 1.1.9-bugfix is but a droplet in the ocean of firmware variants out there.



> as alternative we could send asynchronously only reprap "buffered commands"

I'd prefer that due to the aforementioned reason since it limits the risk of considering a command capable of this that in fact isn't on some firmware variant out there. And it also makes more sense from a "matching up with firmware implementations" point of view, since we'd stick to the planner buffer related commands that way instead of a seemingly arbitrary list that no one understands how it came to be.

src/octoprint/util/comm.py
foosel: Please document that then.

src/octoprint/util/comm.py
foosel: How will it cause overflow to send one *less* command than we are allowed to according to the firmware's buffer indicators? 🤔

src/octoprint/util/comm.py
foosel: The aforementioned states are also printing states however.

src/octoprint/util/comm.py
foosel: Keep in mind that I will have to maintain this code going forward - this needs proper documentation so that that will be possible.

src/octoprint/settings.py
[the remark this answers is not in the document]
linvinus: > I'd prefer that due to the aforementioned reason since it limits the risk of considering a command capable of this that in fact isn't on some firmware variant out there.

@foosel  done in last commit, working but needs for more testing

Evg33:
Hello !
What about resolve conflicts ?

linvinus:
Hi guys. Until the 5th of August I'm on vacation.
@Evg33 which conflicts do you mean?

Evg33:
@linvinus 
```
Conflicting files
AUTHORS.md
src/octoprint/util/comm.py
```

linvinus:
With which version? 

travis don't see any conflicts

Evg33:
image

linvinus:
@Evg33 Ohh, in mobile version i don't see that.

linvinus:
@Evg33 resolved

linvinus:
Currently I have noticed the following:
1) ATMEGA 16MHz can not printing faster than 35mm / s when the LA,JD and S-curve is turned on

For speed test i use the following model https://www.thingiverse.com/thing:1185928
this model don't have straight lines, and almost don't have transfers, which is why the whole gcode consists of many small lines

Model was sliced in Cura with following settings
Layer width 0.4
Layer height 0.2
Shell count 5

2) From octoprint side, the information about buffers in ADVANCED_OK is useless, from it only the line number is needed. If planner buffer goes down octoprint will not be able to react on that fast enough.
The best what  we can do is better fill marlin RX_BUFFER, so this buffer will be as a piece of real gcode  file, so marlin will have all it needs for successful printing.

3) In marlin BUFSIZE is used to protect against transmission errors, but at high speeds with asynchronous transmission, it is useless. if a transmission error was occurs, while the octoprint will resends a broken command the printer will stop the nozzle

4) From octoprint side more importantly is to have information about filling of RX_BUFFER_SIZE , but there is no information about it in ADVANCED_OK

Evg33:
@linvinus Thank you !!!

linvinus:
@Evg33, no problems. let me know if you have any questions. It seems that you are the first man who will testing this patches, i will appreciate for any feedback!

foosel:
@linvinus I have to admit that I lost track of this somewhat after my initial review... I see that you pushed some changes in the meantime, but some points appear to remain unaddressed.

Additionally from [your comment here](https://github.com/foosel/OctoPrint/pull/3209#issuecomment-514375345) it sounds like this isn't ready for prime time yet due to firmware issues, or did I misunderstand that?

linvinus:
@foosel hi!
patch it self is working, but i found that on 8bit processors marlin performance is not enough to print circles faster than 35mm/s.
This is doesn't mean that patch is useless, it steel helps a lot on more simple models, like 3D benchy.

Currently i'm planning to replace 8-bit processor with STM32F1 and then compare print quality.

DiegoGiovany:
+1 for this... printing over wifi at 300mm/s on a tinyfab cpu (LPC1769) in a UP BOX with the linvinus version, with the oficial version, I have multiple slowdowns and blobs, and  even comunications errors...

Evg33:
+1 for this.

foosel:
Please don't leave "+1" comments but instead vote using the built-in Github reactions:

image

src/octoprint/settings.py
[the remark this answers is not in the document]
foosel: The list of buffered commands still seems way too long. `M104` is definitely not buffered. It might be that some of the other ones are buffered on your specific fork, but not in general. We should really stick to universally buffered commands here which - again - are `G0` through `G3` and `G28` through `G32`. Anything more than that and we risk issues with firmware variants.

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+				continue
+
+			if not self._send_queue_active:
+				break
 
-				# now we just wait for the next clear and then start again
+			self._log("####1 wfl={} maxl={} lastl={}".format(self._advanced_ok_wait_for_line, self._advanced_ok_max_line,self._advanced_ok_last_line ))
+			if ( self._advanced_ok_detected
+			and self._advanced_ok_max_line > 0
+			and self._state == self.STATE_PRINTING):
```
foosel: starting, pausing, cancelling, resuming, finishing? Probably better: `self._state in self.PRINTING_STATES`

src/octoprint/util/comm.py
```diff
@@ -4651,6 +4788,30 @@ def parse_resend_line(line):
 
 	return None
 
+def parse_advanced_ok_line(line):
```
foosel: Can you please add some unit tests for this as well?

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+	def _wait_for_line(self):
+		"""
+		Waiting for delivery confirmation of specific line, and while _advanced_ok_max_line >0
+		if for some reason transaction with printer will be broken, the _advanced_ok_max_line will be setted to -1 from outside
+		return value not used
 		"""
+		self._log("_wait_for_line1")
 
-		self._clear_to_send.wait()
+		while True:
```
foosel: busy loop, will max CPU, needs sleep

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+									else:
+										self._advanced_ok_wait_for_line = -1 #don't wait, pool instead
+										return False #stop sending and wait for free buffer
+
+
+				else:
+					# Otherwise we need to tickle the read queue - there might not be a reply
+					# to this command, so our _monitor loop will stay waiting until timeout. We
+					# definitely do not want that, so we tickle the queue manually here
+					self._continue_sending()
```
foosel: Needs a `return True` now or we won't actually continue to try send a line.

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+			if ( self._advanced_ok_detected
+			and self._advanced_ok_max_line > 0
+			and self._state == self.STATE_PRINTING):
+				if(self._advanced_ok_wait_for_line > 0 ):
+					self._wait_for_line()
+				else:
+					self._pool_for_freebuffer()
+			else:
 				self._clear_to_send.wait()
-			except:
```
foosel: This exception handler **must stay**

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
 			self._logger.debug("Type already in send queue: " + e.type)
 			return False
 
-	def _send_loop(self):
+	def _pool_for_freebuffer(self):
```
foosel: Please rename to `_wait_for_free_buffer` or something. `pool_for_freebuffer` really just makes me scratch my head and doesn't tell me anything.

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+		self._log("pool")
+		#pool for free buffer
+		while True:
+			if( self._advanced_ok_max_line > 0
+				and self._current_line > self._advanced_ok_max_line):
+				time.sleep(0.005) #pool 5ms
+			else:
+				return True
+
+	def _wait_for_line(self):
```
foosel: rename to `_wait_for_line_confirmation` - we aren't actually waiting for a line.

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+				if gcode is not None:
+					use_up_clear = True
+
+				if use_up_clear:
+					# if we need to use up a clear, do that now
+					self._clear_to_send.clear()
+
+					if ( self._advanced_ok_detected and self._state == self.STATE_PRINTING):
+						with self._line_mutex:
+							linenumber = self._current_line - 1
```
foosel: You'll have to take into account that a line might have been sent without a number (possible for in-band commands that aren't GCODE, or just for certain configuration options).

src/octoprint/util/comm.py
```diff
@@ -3126,134 +3184,213 @@ def _enqueue_for_sending(self, command, linenumber=None, command_type=None, on_s
...
+							self._advanced_ok_wait_for_line = linenumber #wait for line by default
+							self._log(">>> LN={} G={} State={}".format(linenumber,gcode,self.getStateString()))
+
+							if( self._advanced_ok_max_line > 0 and not self._send_queue.resend_active):
+								if( gcode in self._advanced_ok_buffered_cmds ):
+									if(linenumber < self._advanced_ok_max_line and self._continue_sending()):
```
foosel: needs documentation

src/octoprint/util/comm.py
```diff
@@ -455,6 +460,11 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
 		self._blockWhileDwelling = settings().getBoolean(["serial", "blockWhileDwelling"])
 		self._send_m112_on_error = settings().getBoolean(["serial", "sendM112OnError"])
 		self._current_line = 1
+		self._advanced_ok_detected = False
+		self._advanced_ok_max_line = -1          #the last line which we could send in asynchronous mode without cause overflow, negative value used to interrupt any waitings
```
foosel: rename: `self._advanced_ok_last_sent_line`

src/octoprint/util/comm.py
```diff
@@ -455,6 +460,11 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
...
+		self._advanced_ok_last_line = -1         #the last line that printer has successfully processed
```
foosel: rename: `self._advanced_ok_last_ackd_line`

src/octoprint/util/comm.py
```diff
@@ -455,6 +460,11 @@ def __init__(self, port = None, baudrate=None, callbackObject=None, printerProfi
...
+		self._advanced_ok_wait_for_line = -1     #the next line which we will waiting before continue sending
+		self._advanced_ok_buffered_cmds = settings().get(["serial", "bufferedCommands"])
```
foosel: rename: `self._advanced_ok_buffered_commands`

linvinus:
octoprint
FIG1

@foosel hi!
i saw your comments, but before continue discussing code let me describe where the problem is, what this patch trying to resolve.

Please have a look at figure 1.
On left side we have octoprint on the right firmware (marlin).
Because ram in firmware is too small we can't send whole file, so we send it by chunks - line by line.
On firmware side some commands executing synchronously, that is mean that marlin send OK report only after movements or other physical actions was performed.
(T1 -> T2 -> T3 -> T4 -> T6 -> T7)

Others, instead, goes to special buffer with size of BLOCK_BUFFER_SIZE, and OK report sends immediately after checking crc summ, but  before actual physical actions was performed.
(T1 -> T2 -> T3 -> T4 -> T5)

The problem is  "poor printing quality  like bulbs or unstable line width".

Where there come from?

Because of STL file is triangle based, it unable to describe circles. 
The circles actually represented as small lines. Usually slicers allow configure resolution of such lines.
So any non straight lines is printing as such small lines.
This is meas that some times to print one short figure printer must parse a lot of simple short lines.

So to resolve this printing problems we should
1) increase buffers on printer side as much as possible (or place whole file on sd-card)
2) increase speed of data transfer from octoprint to firmware
3) firmware performance must be enough to parse GCODE fast enough

You may notice that all of them is actually hardware related, so simple use powerful processor with big enough ram and you will not have this problems.

If we can't change hardware, what can be done from software perspective?

From software we may optimize protocol between octoprint and firmware.

Even the shortest loop (T1 -> T2 -> T3 -> T4 -> T5) actually too long, this is meas that
(T1 -> T2 -> T3 -> T4 -> T5) >> T6 - nozzle moving faster than new commands arrive

So, instead of waiting for report from T5 step, we may send new commands in advance.

THIS IS WHAT I MEAN BY "ASYNCHRONOUS COMMANDS"

How much? As much as enough space in RX_BUFFER.
Typical GCODE command length is less than 96 bytes (96 is max command length in Marlin), so with buffer of 2000 bytes we may send around 20 commands in advance.

Because RX_BUFFER is flat and commands usually smaller than 96bytes, if we will be smart enough, on octoprint side we may calculate free space in RX_BUFFER and send commands with respect of that.
So, with commands of size 88 bytes we may send in advance around 22 commands.

So, doesn't matter is the command is "buffered" or not in terms of firmware, we send it in advance.
Yes that is mean that user commands like stop or pause will be interpreted by 22 commands later, this is disadvantage of that optimization. 

But not all of commands may be send in advance, that is why we have bufferedCommands list, it contain commands that may be send asynchronously without causing error or unpredictable behavior in firmware.
It working for marlin, probably, with some others firmware some of this commands may cause problems. Needs testing and user feed backs. 
Any way, the user may adjust this list at its discretion at any time.
Also this new octoprint behavior may be disabled by default.

foosel:
I'm well aware of the buffering situation in Marlin and similar firmwares (and in fact it's three buffers: rx -> command -> planner) and even had to explain it to current Marlin maintainers ;)

I'm also aware of what this patch is trying to achieve and how it's trying to achieve it. I've been playing with stuff like this long before the `ADVANCED_OK` format was introduced and stumbled over some crippling bugs in the firmware in the process which I patched (rx -> command not keeping command order).

Remember that I have to wrap my head around code I didn't write myself here, and if you call something "buffered" commands, then I assume you mean those that are officially considered buffered, as in, they go into the planner buffer. Not arbitrary commands. After taking another look at the code I now understood that you are actually concentrating on keeping the *command* buffer filled, and only that, not the *planner* buffer.

> But not all of commands may be send in advance, that is why we have bufferedCommands list, it contain commands that may be send asynchronously without causing error or unpredictable behavior in firmware.

This is where you've lost me. Things should go straight from RX to Command buffer, in the same order as received, without any changes other than stripping line number and checksum, if set, and potentially requesting a resend first. Nothing should be executed here and hence no command should be blacklisted from getting sent to fill the command buffer. If we are tracking the command buffer here and not the planner buffer, then we shouldn't need a distinction between command types because all commands are the same until it comes to the question, "planner buffer yes or no". And the command buffer will always be executed in order.

```
                            +-------------------------------------+
        line                |                                     |
+------------------------>  |          RX Buffer                  |
                            |                                     |
                            +------------+------+-----------------+
             "Resend:n"                  |      |
<----------------------------------------+      | checksum & linenr ok, strip
                "ok"                            v
                            +-------------------+-----------------+
                            |                                     |
                            |          Command Buffer             |
                            |                                     |
                            +--------+--------------------+-------+
         "ok"                        |                    |
<------------------------------------+--------------------+ buffered
                                                          |
                            +-----------------------------v--------+
                            |                                      |
                            |             Planner Buffer           |
                            |                                      |
                            +--------------------------------------+
```

RX to command buffer in Marlin [is being done here](https://github.com/MarlinFirmware/Marlin/blob/0110faa717e642574ade9eb7aed250ee2840e942/Marlin/Marlin_main.cpp#L1021-L1148). No `ok` generated there other than the one belonging to a resend request triggered by one of the various `gcode_line_error` lines. After that stuff is in the command buffer (which - in Marlin - [gets processed here](https://github.com/MarlinFirmware/Marlin/blob/1.1.x/Marlin/Marlin_main.cpp#L12821)). RepetierFirmware and Smoothie look similar.

So either I'm missing something crucial here, or this whole buffered list that seems to be main concern here isn't even needed (and misnamed since the term "buffered command" is already reserved for commands that go into the *planner* buffer). And if I'm missing something crucial here, please tell me what.

Coffee0297:
if you wanna test with something that will be octoprint bottleneck, take a look at klipper. we have to use virtual sd to get speed on circles (70mm/s+).
https://github.com/KevinOConnor/klipper

fiveangle:
Connecting to original #450 issue...

foosel:
Closing this as it doesn't look like this will work reliably with the current comm layer and/or firmware implementations, and the review comment were never addressed either. Additionally there are also now projects like https://github.com/chendo/BufferBuddy

PR #3225, opened 2019-07-09
Quote arguments in pip install

eyal0:
If the argument is not quoted, the shell will interpret the braces.


foosel:
Just out of interest - what shell? I routinely run this on a bash and never ran into issues O_O

eyal0:
zsh

I recall that it used to work for me, too. I used to use bash.

PR #3246, opened 2019-08-07
added possibility to fire custom events from SockJS

aurror:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR adds the possibility to fire custom events from a SockJS connection.
This allows developers to send data from a SockJS client to a plugin they wrote, essentially allowing bidirectional communication since the plugin > SockJS direction was already possible, especially since the intoduction of custom events. 
This PR builds up on the custom events, making it possible to send data via SockJS to plugins, which can then react on it.

Manual testing by initializing a connection using the Javascript Client Library. I'm not sure how to include a test for the test suite in this case since it requires a socket connection.

**Note**: I tested it on the main branch, where it works and all unit tests pass. However, I could not connect or open the octoprint web interface on this devel branch (even without the code changes, looks like devel is broken atm).

My initial requirement was to display a message box in OctoPrint when I press a button on a printer display. This could've been done by extending the API via the plugin, but since the introduction of custom events ( see [this original issue](https://github.com/foosel/OctoPrint/issues/2965), I thought it would be good to extend this functionality and allow SockJS Clients to communicate with plugins over these custom events. This makes it easy for developers to exchange or synchronize data between their octoprint plugin and e.g. a display connected to the printer or other custom functionality.

See [my community post](https://community.octoprint.org/t/sending-data-via-socket-to-plugin/11070), my original idea was to send the data directly to the javascript side of the plugin, but afterwards I thought it would be better to send it to the python side and channel it through to the js side if necessary.

#### Further notes
Here's an example on how it can operate with a custom plugin and a SocketJS connection using the Javascript Client Library:

First register your custom events in your plugin
```
def register_custom_events(self, *args, **kwargs):
        return ["from_sockjs_to_plugin", "from_plugin_to_sockjs"]
```
Then you can start communication.

---

To send data from SockJS to your plugin, you can use
```
OctoPrint.socket.sendMessage("plugin_myplugin_from_sockjs_to_plugin", myJSONData })
```
and react on the plugin side on it by registering it for events using [the EventHandler Mixin](https://docs.octoprint.org/en/master/plugins/mixins.html#eventhandlerplugin) like this:
```
def on_event(self, event, payload):
        if (event == octoprint.events.Events.PLUGIN_MYPLUGIN_FROM_SOCKJS_TO_PYTHON):
                # ...
```

---

To send data from your plugin to SockJS:
```
self._event_bus.fire(octoprint.events.Events.PLUGIN_MYPLUGIN_FROM_PYTHON_TO_SOCKJS, myData)
```
and react on the SockJS side by registering 
```
OctoPrint.socket.onMessage("event", function(message) {
    if (message.data.type == "plugin_myplugin_from_python_to_sockjs")
    {
         // do something...
         // message.data.payload == myData
    }
});
```
**Note:** This is my first pull request, if something is wrong please tell me. :)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -105,6 +105,7 @@ date of first contribution):
   * [Jack Wilsdon](https://github.com/jackwilsdon)
   * [Ryan Finnie](https://github.com/rfinnie)
   * [Timur Duehr](https://github.com/tduehr)
+  * [Florian Horn](https://github.com/aurror)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/docs/jsclientlib/socket.rst
+++ b/docs/jsclientlib/socket.rst
@@ -64,8 +64,14 @@
 
    Sends a message of type ``type`` with the provided ``payload`` to the server.
 
-   Note that at the time of writing, OctoPrint only supports the ``throttle`` message. See
-   also the :ref:`Push API documentation <sec-api-push>`.
+   Note that at the time of writing, OctoPrint supports the ``throttle`` and ``auth`` message as well as messages for registered custom events. See
+   also the :ref:`Push API documentation <sec-api-push>` and the :ref:`custom events hook <sec-plugins-hook-events-register_custom_events>`.
+
+   An example on how to use it with custom events is shown below:
+
+   .. code-block:: javascript
+
+      OctoPrint.socket.sendMessage("plugin_myplugin_my_custom_event", {"data": myData})
 
    :param string type: Type of message to send
    :param object payload: Payload to send
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -26,6 +26,9 @@ _instance = None
 def all_events():
 	return [getattr(Events, name) for name in Events.__dict__ if not name.startswith("_") and not name in ("register_event",)]
 
+# return all custom events that were created using register_custom_events
+def custom_events():
+	return [getattr(Events, name) for name in Events.__dict__ if not name.startswith("_") and not name in ("register_event",) and name.startswith("PLUGIN_")]
 
 class Events(object):
 	# server
--- a/src/octoprint/server/util/sockjs.py
+++ b/src/octoprint/server/util/sockjs.py
@@ -209,6 +209,14 @@ class PrinterStateConnection(octoprint.vendor.sockjs.tornado.SockJSConnection,
 			else:
 				self._throttleFactor = throttle
 				self._logger.debug("Set throttle factor for client {} to {}".format(self._remoteAddress, self._throttleFactor))
+		else:
+			# handle custom events
+			for eventIdentfier in message:
+				registeredEvents = octoprint.events.custom_events()
+				if eventIdentfier in registeredEvents:
+					self._eventManager.fire(eventIdentfier, message[eventIdentfier])
+				else:
+					self._logger.warn("received SockJS message without an associated registered event: {} ".format(eventIdentfier))
 
 	def on_printer_send_current_data(self, data):
 		if not self._user.has_permission(Permissions.STATUS):


foosel:
Hmmm.... I'm not on board with this. From how I understand you'd use this it feels more like you are abusing the event subsystem for communication flows that should rather go on the API.

I introduced the ability to register custom events not to add an additional message transport for plugins - it was already possible to [send arbitrary plugin push messages via the socket](http://docs.octoprint.org/en/master/modules/plugin.html#octoprint.plugin.core.PluginManager.send_plugin_message) (though the documentation could definitely be better here I admit) - but rather to allow plugins to introduce custom events for general consumption that could then be used by other event consumers (e.g. for triggering shell scripts, or messages on a MQTT broker). That they also get pushed to the frontend is simply part of the event subsystem implementation. If you ever find yourself introducing a custom event that is only relevant for your specific plugin frontend code and there's no other use for it, a custom event sounds like the wrong thing to use.

The "right" way to do a bidirectional communication is primarily the REST API through [custom endpoints](http://docs.octoprint.org/en/master/plugins/mixins.html#blueprintplugin) or the [SimpleApiPlugin mixin](http://docs.octoprint.org/en/master/plugins/mixins.html#simpleapiplugin). If things need to happen asynchronously these API calls can be triggered by pushing messages over the socket from the server to the client. That way the amount of data on the socket (that currently gets mirrored to all clients) stays minimal and we can also depend on existing caching mechanism for large responses (think file listings).

I'd rather prefer to make it more obvious how to use these existing mechanisms by improving the existing documentation and maybe add a new chapter to the plugin tutorial than introducing a concurring mechanism that utilizes the event subsystem for things that shouldn't be events in the first place.

I've thought about opening up the socket for bidirectional communication in the past, but so far there never was an actual need other than moving *all* communication between client and server to it, and I'm not sold yet on that being the most sensible approach here due to that making a ton of stuff more difficult (caching, authentication, debugging, ...).

Coming back to your original use case, I'd say a custom event for a firmware generated event would definitely make sense (e.g. waiting for user), but I'd react to that with a defined API (if client interaction is needed) or just a server side event subscriber.

aurror:
You're right, I didn't think about the overhead this would create for other event receivers.
If I understood it correctly, I can avoid this overhead when sending data from the client to the plugin via an API, but in the other direction I still have to use push messages (if I can't trigger it from the client side when I don't know data is available now) which get send to every receiver. 

Is there a way to send data from a plugin to a JS Client directly without sending it to every message receiver?
All I can think of right now would be creating own data channels which you could subscribe to, on which messages get only send to subscribed clients and not to e.g. the on_event function to avoid the overhead.

foosel:
> If I understood it correctly, I can avoid this overhead when sending data from the client to the plugin via an API, but in the other direction I still have to use push messages (if I can't trigger it from the client side when I don't know data is available now) which get send to every receiver.

Yes. Think of messages on the socket more as someone putting a note in your mailbox that says "please call me if you are interested in X".

> Is there a way to send data from a plugin to a JS Client directly without sending it to every message receiver?

No, currently clients don't have an identifier or anything like that, and there's no way to even determine the connected clients from the deeper parts of the backend from which you can trigger messages or events. Simply wasn't needed so far, because the "there was an update in this part of the system" approach with triggered refresh and caching suffices.

> All I can think of right now would be creating own data channels which you could subscribe to, on which messages get only send to subscribed clients and not to e.g. the on_event function to avoid the overhead.

The question is what kind of problem you want to solve here. If there's an asynchronous state change in the backend, it usually makes sense to notify everyone about it with a quick small "heads-up" message (whether via a dedicated plugin message or a system wide event depends on the exact use case) because connected clients might need to react to it. But since we don't know *how* they might need to react to it, it's a small heads-up, not a full blown data dump. Only the information "hey, this and this changed", not the full catalogue of "and here's all the new resulting internal state". The receiver can then decide what to do about it, if it wants to do anything about it at all.

If there's no asynchronous state change that needs to be immediately handled, it suffices if the client fetches new data when needed (e.g. when the user opens a related dialog or clicks a button).

I currently can't think of any case where I ever felt the need to address one specific connected client for something after initial handshake, but I remember several cases where I hooked several different triggered requests to on simple event pushed by the system for which I used the REST API.

aurror:
Okay, 
I'll follow the route of notifying about the update approach and reacting on it with API.
Thanks for the clarification and explanation. :)

PR #3269, opened 2019-09-11
Fix typo

Guilouz:
Just a minor typo fix.


foosel:
That's actually intentional ;) The ping is every 15min, the pong is every 24h.

PR #3335, opened 2019-11-09
Flask clean all data list iterator fix

cbxbiker61:
Fixes a dictionary changed during iteration exception.  Potentially triggered only on python3.

BTW, I'm running Octoprint on Python 3.8 now without any apparent issues.


foosel:
Ah, great, thanks for spotting this and fixing it!

> BTW, I'm running Octoprint on Python 3.8 now without any apparent issues.

That's great to know! Thank you!

Issue #3006, opened 2019-01-18
[Request] Allow disabling the GCODE analysis


nforceroh:
#### What were you doing?
gcode visualizer runs even if disabled in settings which causes Raspberry PI Zero to come to a crawl after octoprint is restarted.
```
pi@delta01:~ $ ps -ef|grep octoprint
pi         342     1 57 12:23 ?        00:05:11 /home/pi/oprint/bin/python2 /home/pi/oprint/bin/octoprint serve --host=127.0.0.1 --port=5000
pi         862   342 33 12:24 ?        00:02:28 /home/pi/oprint/bin/python2 -m octoprint analysis gcode --speed-x=6000 --speed-y=6000 --max-t=10 --throttle=0.01 --throttle-lines=100 /home/pi/.octoprint/uploads/Body_Cover.gcode
 a few minutes later
pi        3400   342 42 12:38 ?        00:01:18 /home/pi/oprint/bin/python2 -m octoprint analysis gcode --speed-x=6000 --speed-y=6000 --max-t=10 --throttle=0.01 --throttle-lines=100 /home/pi/.octoprint/uploads/Wheel_extruder.gcode
````
#### What did you expect to happen?
gcode visualizer should not run if disabled

#### What happened instead?
gcode visualizer  is running

#### Did the same happen when running OctoPrint in safe mode?
no

#### Version of OctoPrint
OctoPrint 1.3.10 

#### Operating System running OctoPrint
OctoPi 0.15.1

#### Printer model & used firmware incl. version
Marlin 2.x

#### Browser and version of browser, operating system running browser
chrome

#### Link to octoprint.log
```2019-01-18 12:23:57,672 - octoprint.startup - INFO - ******************************************************************************
2019-01-18 12:23:57,682 - octoprint.startup - INFO - Starting OctoPrint 1.3.10
2019-01-18 12:23:57,685 - octoprint.startup - INFO - ******************************************************************************
2019-01-18 12:23:57,957 - octoprint.util.connectivity_checker - INFO - Connectivity changed from offline to online
2019-01-18 12:24:01,501 - octoprint.startup - INFO - Blacklist processing done, adding 1 blacklisted plugin versions: roomtemp (any)
2019-01-18 12:24:01,753 - octoprint.plugin.core - INFO - Loading plugins from /home/pi/oprint/lib/python2.7/site-packages/octoprint/plugins, /home/pi/.octoprint/plugins and installed plugin packages...
[...]

PR #3337, opened 2019-11-12
Setting to allow gcode analysis to continue during printing

CapnBry:
GCODE analysis is great for providing a print time estimation, but currently you have to wait for analysis to finish before hitting the print button, because analysis can't run during printing by design. Since most users are on multicore systems (Pi 2 and above, or a PC) it is nice to use those extra cores to continue analysis during the print. This adds an option to `config.yaml -> gcodeAnalysis -> runAt: default=idle` to allow analysis to continue during printing to populate the time estimate or to disable analysis completely.

I've been running with this change for several years, hundreds of prints, no issues. The time estimate just pops in once the analysis completes. I've never seen any strange behavior of the two processes interacting poorly, or becoming wedged in a deadlock. 

I think you're nice?

None

None

#### Further notes
Ideally, the code should allow the background analysis to switch throttle_lines or the priority mode of the already-running gcode process, but that would require some sort of IPC pipe to the analysis process and I felt that was too big a change for what would be very little benefit. There is plenty of CPU on a Pi3 to run both processes at full priority with no degradation to the gcode streaming/print functionality.

~~I also considered making the analysis process an enum `{ Never, OnlyWhenIdle, Always }` to allow the user to also completely block the analysis process entirely (i.e. never analyze files) but I don't think there's much of a use case for that. OnlyWhenIdle would be the current functionality, and Always would be the same as this patch with pauseDuringPrint=false.~~ This has been added

### DIFF
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -19,7 +19,6 @@ from .analysis import QueueEntry, AnalysisQueue
 from .storage import LocalFileStorage
 from .util import AbstractFileWrapper, StreamWrapper, DiskFileWrapper
 from octoprint.util import get_fully_qualified_classname as fqcn
-from octoprint.settings import settings
 
 from collections import namedtuple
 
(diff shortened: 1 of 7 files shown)


CapnBry:
Updated PR with option to fully disable gcode background analysis under Settings -> Features -> Analyze gcode for time and model size estimates. Should cover #3006

foosel:
:+1:

PR #3346, opened 2019-11-25
Etag and refresh optimizations

justfalter:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Two performance improvements that came as a result of profiling OctoPrint. My primary goal has been to reduce the CPU impact when it comes to serving /.  All performance numbers I'm reporting here are from running using python 3.7.5 on my 2016 Macbook Pro. I expect the improvements to be even more significant on a typical raspberry pi.

**1. Improve performance of etag calculation for / (root route):**
- Only compute etag once (rather than twice) per request.
- Only stat files once per request.
  - Avoid calling `isdir` or `isfile` in `collect_files`.
  - Leave it to `compute_lastmodified` to stat files, which tells us if the path exists, if it is a file, and it's last modified.
- Ultimately, this eliminates ~70% of `stat` system calls. This has a bigger impact on raspberry PI systems. 

Timing (as reported by Tornado) on my 2016 Macbook Pro when fetching / (`curl -o /dev/null http://octoprint/`):
- Before: ~50ms
- After: ~38ms

**2. Memoize settings hashes to speed up cache refresh:**
Memoize `settings.config_hash` and `settings.effective_hash`. The
hashes will only be calculated on the first call to the respective
functions. These memoized values are invalidated when the settings
object becomes dirty, or the settings are loaded from the config file.

Reason for change: During a refresh of the root route (/), the hash of
`settings.effective.yaml` is calculated 5 times. Each call lead to a
large configuration object to be serialized to a YAML string, which would
then be hashed. As it turns out, serializing to YAML is expensive.

Note: while I do not see `settings.config_hash` hit during a refresh,
but making a change to it was easy and appropriate. It's possible that
other areas of the application may see some benefit.

Performance (on my 2016 Macbook Pro:
- building of initial cache on startup:
  - before: ~3.4s
  - after:  ~2.9s
- loading `http://octoprint/?_refresh=1`
  - before: ~1100ms
  - after:  ~200ms

Manually.
I ran pytest, and everything passes.

### DIFF
--- a/src/octoprint/server/util/flask.py
+++ b/src/octoprint/server/util/flask.py
@@ -1393,7 +1393,7 @@ class SettingsCheckUpdater(webassets.updater.BaseUpdater):
 			return False
 
 		cache_key = ('octo', 'settings')
-		current_hash = webassets.utils.hash_func(json.dumps(settings().effective_yaml))
+		current_hash = settings().effective_hash
 		cached_hash = ctx.cache.get(cache_key)
 		# This may seem counter-intuitive, but if no cache entry is found
 		# then we actually return "no update needed". This is because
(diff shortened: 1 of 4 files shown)


foosel:
Thanks. I've scheduled that for 1.4.1 to give it a more thorough look than I can currently do due to 1.4.0 stress.

foosel:
Manually merged into `maintenance` (realized too late that this was targeting `devel` and should have been rebased first)

Issue #3394, opened 2019-12-17
comm port autodetection unreliable/broken


bzed:
Hi,

printer auto connection is still broken on staging/devel. Connection times out instead of working as expected.

```
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/comm.py", line 2693, in _openSerial
    serial_obj = factory(self, self._port, self._baudrate, settings().getFloat(["serial", "timeout", "connection"]))
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/comm.py", line 2646, in default
    port = self._detect_port()
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/comm.py", line 2627, in _detect_port
    programmer.connect(p)
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/avr_isp/stk500v2.py", line 36, in connect
    self.sendMessage([1])
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/avr_isp/stk500v2.py", line 110, in sendMessage
    return self.recvMessage()
  File "/home/bzed/tools/OctoPrint/src/octoprint/util/avr_isp/stk500v2.py", line 121, in recvMessage
    raise ispBase.IspError("Timeout")
octoprint.util.avr_isp.ispBase.IspError: 'Timeout'
```

I'll keep poking into it tonight, more to come (hopefully...).

foosel:
So I've tried reproducing this and while at least against a creality board (which should be similar or even the same to what I know you have) it indeed doesn't work under py2 or 3 under 1.4.0rc3, neither does it work under 1.3.12. 

So right now this doesn't look like a regression to me, and it actually autodetects another board (arduino mega) just fine.

Would you happen to remember when you last saw this working for you?

I'll fix this in any case by finally throwing out the stk500/programmer mode stuff because that was error prone AF to begin with, but since that will require some rewiring of quite sensitive parts of the comm layer that is not something I want to risk on an RC.

foosel:
Considering this not a regression for now since I still haven't been able to make stuff that doesn't autodetect under 1.4.0 to autodetect under 1.3.x either. Going to move this to the 1.4.1 milestone and to be solved by finally getting rid of this horribly messy programmer mode bs altogether (as I said, that's been a thorn in my side for way too long anyhow)

MatejSpindler:
I have a similar issue on 1.4.0rc5.
Versions 1.3 worked OK.

[...]

PR #3397, opened 2019-12-17
comm port autodetection rewrite

bzed:
Hi :)

So this is my take on fixing #3394.
I *know* that comm port autodetection worked before, but with the current staging/devel checkout it failed, with Python 2 and Python 3.

So I've started to look into the avr code and tried to debug it, got annoyed, and removed the whole code and replaced the whole thing by a stk500v2 sync and sign_on. I don't think there is a need to check if there is actually a programmable board behind that port, also supporting all possible boards sounds impossible.

What I'm absolutely not sure about is the timeout setting, I think it should be done to 1s or something like that.
I've tested whats happening with my multimeter (which does not even accept writes, especially not with !=9k6 baud), Octoprint ignores it just fine.

There is (imho) one bug left: after a disconnect, it is not able to detect the port on the first try. As far as I can see the fix for that is to 'sync' a few times until its in sync. Tiny loop around [0x30].

Otherwise the coew works just fine here.


src/octoprint/util/comm.py
```diff
@@ -2616,23 +2603,29 @@ def _detect_port(self):
...
+				try:
+				        serial_obj = serial.Serial(str(p), baudrate, timeout=1, writeTimeout=10000)
+				except Exception:
+					self._logger.info("Could not connect to {} ({}).".format(p, baudrate))
+					continue
 				try:
-					self._log("Trying {}".format(p))
-					programmer.connect(p)
-					serial_obj = programmer.leaveISP()
+					if _send_message(serial_obj, [0x30]) != b'\x00\n':
```
bzed: We should try this up to 5 times. This should fix a reconnect after a  disconnection.

bzed:
Oh well, yes dear bot.
Applies fine to devel, though :)

foosel:
Ignore the bot here, I forgot to whitelist the branch.

I'll take a closer look at this ASAP, but it might be only after my upcoming vacation.

foosel:
Ah, but don't ignore the test failure, though I can also take care of that. Wrapping things up for 2019 now though, so it'll have to wait until I'm back to work in January 🎄

bzed:
> Ah, but don't ignore the test failure, though I can also take care of that. Wrapping things up for 2019 now though, so it'll have to wait until I'm back to work in January

yeah, but I have similar issues with free time, so I thought I'd trow the branch at you in case you want to fix it before I find the time to do so ;)

bzed:
@foosel oh well, that was easy. Travis is happy.

bzed:
Reconnecting after disconnect should work now, too. Not tested yet, though.

src/octoprint/util/comm.py
bzed: Fixed in 08dd797

foosel:
Planning to get rid of the "try to enter programming mode" approach in 1.4.1 altogether, since this is all just horribly messy anyhow. Hence not going to merge this (especially not since I couldn't get it to work on my end with printers available to me as discussed on IRC, so there's something wonky about all of this anyhow).

PR #3411, opened 2020-01-01
Show retracts, moves, and head position only for current layer

drifkind:

In the GCode viewer,  turning on "also show current" or "also show previous" makes it easier to follow job progress, but showing retracts etc. for "also" layers makes too much visual noise.

I did not see a need to make this behavior optional. You could argue that at least the current layer should show moves, but it works for me.

Far too much time spent staring at the screen while the printer grinds away.

### DIFF
--- a/src/octoprint/static/gcodeviewer/js/renderer.js
+++ b/src/octoprint/static/gcodeviewer/js/renderer.js
@@ -680,7 +680,7 @@ GCODE.renderer = (function(){
                 // neither extrusion nor move
                 if (cmd.retract == -1) {
                     // retract => draw dot if configured to do so
-                    if (renderOptions["showRetracts"]) {
+                    if (renderOptions["showRetracts"] && !isNotCurrentLayer) {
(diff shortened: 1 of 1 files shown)


foosel:
Thanks! I'll look into this for 1.4.1, I'm currently focused on 1.4.0 though so it might be a while until I get around to take a proper look.

fieldOfView:
I don't think your rebase worked as intended.

foosel:
As already pointed out by @fieldOfView the PR is completely messed up and contains commits from myself now.

Could you please make sure your PR is a) filed against the `maintenance` branch (it's not a big new feature, it's an improvement of an existing one if I understand correctly) and b) only contains the changes *you* made (ideally squashed into one commit)?

Thank you.

drifkind:
The actual change is tiny, and the only reason I don't start over is misguided stubbornness. I'll figure it out. You're welcome to laugh.

foosel:
Somehow I never got a notice that the PR was fixed, taking a look now.

PR #3497, opened 2020-03-29
Improve performance

flaviut:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR improves performance, especially on low-end hardware like my Bananna Pi Zero.

It does this in several ways:

- It compresses CSS & JS with gzip on startup, and serves up the gziped files. This reduces network transfer size, filesystem cache requirements, and disk IO requirements.
- Some of the javascript already has minified versions checked in (jq & sockjs), so in those cases, the minified version is added to the bundle.
- The CSS & JS & fonts are preloaded, so that all those requests are put in flight all at once, right when the html page is first loaded.
- less.js is included in the javascript bundle because it's not clear why that wasn't done in the first place. And that's one less request.
- JS has the extra whitespace stripped out of it, which reduces size of the transferred files. Identifiers are not mangled, and the scripts can be reformatted in the browser developer tools.
- Modernizr was rebuilt with only the things that are necessary. The big win here is reducing size, and avoiding a 50-ish ms css re-layout to test if the browser supports touch events.

Using the "Audit" functionality in the chrome developer tools in order to measure page load time.

Focus on the very bottom pane, at the "Timestamp" of the "Largest Contentful Paint". Also note how more requests are performed in parallel, and how the amount of data transferred is decreased.

##### Before
op-before

##### After
op-after


foosel:
Thanks! I did some minor changes in fde070665d9cb48a1c59af787fa3082bb6dd4941 and also merged it into `maintenance` so this will make it into 1.4.1 already.

flaviut:
Thanks for merging!

I left a couple comments on fde0706 that I'd appreciate being answered just for my curiosity.

Sorry about the flash of incorrect styling on loading. I should have caught that.

flaviut:
from youtube:

> *repcsi*: It takes way more time and computing power to process, but for text files like js files does it give you that more compression?

sure, but it's a one-time startup thing and I figured it's really not a big enough deal to bother changing it--but at the same time I wanted to make the default explicit rather than implicit.

Issue #2911, opened 2018-11-25
[Request] End the Feedrate / Flowrate confusion


ragflyer:
related closed issue: #2093
I too suffer confusion of terminology and have to look it up every couple of weeks when the default setup does not give a great result rightaway.
It was no surprise to find a closed issue, given the amount of hits you get when searching for the combination of the terms flowrate and feedrate.
What people actually want is change "Speed" and "Extrusion factor".
I agree a standard terminology should not be changed.
May I suggest adding a tooltip or better an additional description to spare new and occasional users from googling so often? Would make a big difference in usability.

ntoff:
Give this a try https://github.com/ntoff/OctoPrint-Flowspeedrename You'll have to install it manually since it's not registered. It adds a page in octoprint's settings to let you change the buttons to whatever you want (not entirely sure how, or even if, it'll handle weird characters from languages other than english).

![image](https://user-images.githubusercontent.com/9030085/48988720-b6c91f80-f172-11e8-8477-928f0c822578.png)

ragflyer:
Thanks, that works for me. Not sure about new users though.

ntoff:
and an unintended consequence of not sanitising the input, you can put fontawesome icons in there too

![image](https://user-images.githubusercontent.com/9030085/49332673-42243400-f5fd-11e8-9813-edfb2f8e1d7f.png)


eg ``<i class="fa fa-arrows"></i> Speed``

JohnEdwa:
A setup item which would match the terms with your slicer and/or firmware could make sense (kinda like the camera controls on Fusion 360), but that would still leave the most common "Marlin with Cura" with Feed Rate and Flow. 

The iconography @ntoff added does look pretty spiffy though, maybe something like that could be implemented by default? Feed Rate could use a [speedo/tachometer gauge](https://icons8.com/icon/1412/speed), and Feed Rate is obviously [Line Width](https://icons8.com/icon/46246/line-width), it even has a graphic for underextrusion :P !

Also, I quote foosel, as she does it exactly as I do:
>I just remind myself that the only thing that flows is molten plastic, and that the stepper movements themselves are therefore the other thing ¯\(ツ)/¯

RichyT:
Are we stating that Feed rate does not affect extrusion speed? Because I have seen elsewhere stated that it does. What fun when it seems that the people seeking to clarify for the confused themselves get things wrong (one way or the other).

[...]

PR #3503, opened 2020-04-01
Extend feedrate to 300%

tobiasschuerg:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Extends the range of the feedrate as 150% is too low if a print was started with a "fail save configuration".

Should work... at least there is no sanity check which could interfere...
https://github.com/OctoPrint/OctoPrint/blob/81735d07c9d96720b4dfdc17a832f9be2affa950/src/octoprint/static/js/app/client/printer.js#L143

Mentioned in #2911 by @hedva92 and I was wondering exactly the same.
https://github.com/OctoPrint/OctoPrint/issues/2911#issuecomment-596212724

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -115,6 +115,7 @@ date of first contribution):
   * [Ludovico de Nittis](https://github.com/RyuzakiKK)
   * [Dominik Paľo](https://github.com/DominikPalo)
   * [Kelly Anderson](https://github.com/cbxbiker61)
+  * [Tobias Schürg](https://github.com/tobiasschuerg)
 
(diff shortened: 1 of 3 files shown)


foosel:
You forgot a check in the backend, but I fixed that ;) Actually modified things to get rid of the upper limit altogether and switch to a number entry field because this would otherwise just have led to either a needed config setting or another PR down the road when someone didn't like 300% either. We'll leave it to the firmware now to stop anything that isn't within operational parameters.

Also added a note that this stuff is write-only because that also kept coming up all the time.

PR #3647, opened 2020-07-20
🎥 Added HLS webcam support.

chudsaviet:
HLS webcam stream support - https://en.wikipedia.org/wiki/HTTP_Live_Streaming .

Potential issues I see:
1. Now the streamer type (MJPG or HLS) is detected by URL's ending. If it ends with 'm3u8' - it's HLS. I think it's easier to do it like this than add a separate drop down menu to settings page. But it limits the URL flexibility, so in rare cases it won't work.
2. I'm not sure if I am able to support flip/mirror capabilities on UI level like OctoPrint have for MJPG stream. My opinion is that the mirror/flip/other filters can and shall be done on HLS streamer level.

Note: **[CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)** is a frequent reason of stream failures. CORS is a mechanism that restricts usage of resource from other websites. It's impossible to identify CORS error from JS *by design*, so open your **browser console** to determine if CORS is the issue.
Workarounds:
1. On the HLS stream **source**, allow OctoPrint website to use it's resources - set [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) HTTP header.
2. Proxy HLS stream with the same proxy that is used for OctoPrint. I will add this solution to OctoPi.

As a test stream with `Access-Control-Allow-Origin:*` you can use https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8 from [HLS.js demo](https://hls-js.netlify.app/demo/).

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -121,6 +121,7 @@ date of first contribution):
   * [Josh Friend](https://github.com/joshfriend)
   * [Lachlan Cresswell](https://github.com/lachyc)
   * [Khoi Pham](https://github.com/osubuu)
+  * [Timofei Korostelev](https://github.com/chudsaviet)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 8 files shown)


foosel:
Just a quick heads-up that I have not forgotten about this, I just have some RC related stuff going on at the moment.

chudsaviet:
No probs, take your time.

foosel:
I've cherry-picked this into `maintenance` as 105078f4b1450a462771a595669bf5a98c6384e7 so it will be part of the 1.5.x release and not have to wait until 2.x. 

Closing this PR manually since Github doesn't detect cherry-picking, but rest assured, it has been merged.

chudsaviet:
Danke schön!

PR #3710, opened 2020-08-30
option/settingsmenuname

OllisGit:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

In the settings menu on the left side the Link-Names are taken from the plugin_name variable. Some plugins (like my) use names like this: "SpoolManager Plugin". So, the post fix "Plugin" is redundant in the settings menu.
But if I change the plugin name to e.g. "Spool Manager", then the update manager and maybe also the stats doesn't work properly.

This PR provides the ability to adjust the settings link name via the plugin. It works like adjusting the Tab-Names.
The plugin author just overwrite the TemplatePlugin mixing like this:
```
	##~~ TemplatePlugin mixin
	def get_template_configs(self):
		return [
			dict(type="tab", name="Spools"),
			dict(type="settings", custom_bindings=True, settings_menu_label="LABEL: Spool Manager")
		]
```
If the key ```settings_menu_label```is not set, the the default way is used (plugin_name) 

In a sample plugin overwrite the TemplatePlugin mixing like above. Add
```
settings_menu_label="LABEL: Spool Manager"
```

no

https://github.com/OllisGit/OctoPrint-SpoolManager/issues/47

image

#### Further notes
If excepted, then this feature must be added to the documentation ;-)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -122,6 +122,7 @@ date of first contribution):
   * [Lachlan Cresswell](https://github.com/lachyc)
   * [Khoi Pham](https://github.com/osubuu)
   * [Federico Nembrini](https://github.com/FedericoNembrini)
+  * [Ollis Git](https://github.com/OllisGit)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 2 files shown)


jneilliii:
> But if I change the plugin name to e.g. "Spool Manager", then the update manager and maybe also the stats doesn't work properly.

Why would you think this to be the case? I'm pretty sure all that is based on plugin_identifier, and you can control the Display Name in the Software Update plugin as well in it's own callback.

cp2004:
Also, is this not already a thing through the `name` attribute? That's what I used in the 'Virtual Printer Settings' plugin, just set the name to 'Virtual Printer' so it is indistinguishable from the existing virtual printer page. You've used the name in the tab, why not also the settings? Unless I'm missing something.

OllisGit:
Damn, how could I missed that.

In my mind I still had the memory that I cannot rename my DeleteAfterPrint plugin to DeleteMoveAfterPrint. I think I also had a short chat with Gina.

Anyway, I can use the "name" key and everything works as expected.

Thanks for your support!!!
Olli

PR #3702, opened 2020-08-24
Fix Application keys manual registration not showing with no keys registered

cp2004:
* [x] Your changes are not possible to do through a plugin and relevant
    to a large audience (ideally all users of OctoPrint)
  * [x] If your changes are large or otherwise disruptive: You have
    made sure your changes don't interfere with current development by
    talking it through with the maintainers, e.g. through a
    Brainstorming ticket
  * [x] Your PR targets OctoPrint's devel branch if it's a completely 
    new feature, or maintenance if it's a bug fix or improvement of
    existing functionality for the current stable version (no PRs 
    against master or anything else please)
  * [ ] Your PR was opened from a custom branch on your repository
    (no PRs from your version of master, maintenance or devel please),
    e.g. dev/my_new_feature or fix/my_bugfix
    *oops, oh well it's only a small fix, on a fresh fork. I'll get it next time*
  * [x] Your PR only contains relevant changes: no unrelated files,
    no dead code, ideally only one commit - rebase and squash your PR 
    if necessary!
  * [x] Your changes follow the existing coding style
  * [ ] If your changes include style sheets: You have modified the
    .less source files, not the .css files (those are generated with
    lessc)
  * [x] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [x] You have run the existing unit tests against your changes and
    nothing broke
  * [x] You have added yourself to the AUTHORS.md file :)

#### What does this PR do and why is it necessary?
Fixes appkeys manual registration when not 

#### How was it tested? How can it be tested by the reviewer?
Run OctoPrint, look at the appkeys page - see update. 
Docs change has NOT been tested, I don't have it set up. Just swapped the image, so I think it will be ok.

#### Any background context you want to provide?
https://community.octoprint.org/t/no-way-to-register-an-application-key/23351/10
Thanks to @jneilliii for doing the research

#### What are the relevant tickets if any?
None (that I know of)
#### Screenshots (if appropriate)
bundledplugins-appkeys-settings

#### Further notes
None

PR #3712, opened 2020-08-31
fix: fixed cant add the firstApplication Key

lhelman:

In the OctoPrint Settings page, on the Application Keys feature,
you couldn't add any new key when there are no keys present.

After this change, it will show `Manually generate an application key`
text and it will allow you to create a new application key.

The way to replicate is, on a local OctoPrint

```bash
rm -rf ~/.octoprint
octoprint serve
```

Go in a browser to http://127.0.0.1:5000 and after configuration,
navigate to the `Application Keys` page.

Not sure why this one is not allowing, while the `template_usersettings`
has the fix.

* I've run octoprint serve on a clean install before and after the change
* I've run the tests and nothing broke.

#### Further notes

Best if the changes are seen without considering spaces.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -122,6 +122,7 @@ date of first contribution):
   * [Lachlan Cresswell](https://github.com/lachyc)
   * [Khoi Pham](https://github.com/osubuu)
   * [Federico Nembrini](https://github.com/FedericoNembrini)
+  * [Leonardo Helman](https://github.com/lhelman)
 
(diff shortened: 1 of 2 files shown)


jneilliii:
I think this is a duplicate of #3702

cp2004:
It is a duplicate of #3702, thanks for the contribution. In future can you make PRs against somewhere other than master please, as it forms the latest release.

PR #3723, opened 2020-09-11
Fixes for couple bugs

4e4o:
Hi, sorry for my english. I have 3d printer with fw Marlin 2.0.6.1, board - skr 1.3. Octoprint runs on orange pi and when octoprint starts - printer is always off (it is turned on later by SSR relay). At first i was connect board and orange pi with usb and connect/disconnect button in octoprint works fine and without bugs. But then i moved connection from usb to gpio/uart logic and that is where bugs comes. I think this bug appears because direct uart connection is faster than usb. Sometime octoprint hangs when connecting to board, everytime in the same place, see first attached screenshot. After a day of remote debugging i was able to find cause of this deadlock. First two commands that octoprint sends is sayHello commands, inside sayHello you are incrementing clear_to_send counter that checked by sender_thread. By default clear_to_send has max value = 1, so after second sayHello call clear_to_send will have value = 1. Inside sender_thread you have clear_to_send.wait() before send cycle, so if both sayHello call will be executed before this clear_to_send.wait() call, then later in send cycle it will hang forever in second clear_to_send.wait(). So it is race condition. Forgot to say that monitor thread is alive, only sender_thread is deadlocked, i can't send anything, but i receive temperature messages from board. I was able to workaround this bug by setting ok buffer to 2, but i think it is ugly solution, so i fixed it in another way (see commit). Also i found that after disconnecting from printer sender_thread not closed properly and every connect/disconnect cycle causes sender_threads amount to grow up. I fixed this too, see commits. Sorry for my bad english again.

### DIFF
--- a/.github/workflows/trigger_docker.yml
+++ b/.github/workflows/trigger_docker.yml
@@ -13,6 +13,6 @@ jobs:
         uses: peter-evans/repository-dispatch@v1
         with:
           token: ${{ secrets.DOCKER_ACCESS_TOKEN }}
-          repository: OctoPrint/docker
-          event-type: octoprint-release
-          client-payload: '{"ref": "${{ github.ref }}", "sha": "${{ github.sha }}", "event-name": "${{ github.event_name }}", "event-type": "${{ github.event.action }}"}'
+          repository: ${{ github.repository_owner }}/octoprint-docker
+          event-type: 'release'
+          client-payload: '{"tag_name": "${{ github.event.release.tag_name }}"}'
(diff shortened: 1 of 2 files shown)


cp2004:
Hi @4e4o - Thanks for the contribution! Unfortunately as the bot has pointed out, the PR should be against `maintenance`. You might need to close this one, and make the changes on a custom branch of your repository targeting `maintenance`. 

`master` is cloned by people, so ideally it should not change between releases.

4e4o:
> Hi @4e4o - Thanks for the contribution! Unfortunately as the bot has pointed out, the PR should be against `maintenance`. You might need to close this one, and make the changes on a custom branch of your repository targeting `maintenance`.
> 
> `master` is cloned by people, so ideally it should not change between releases.

ok, will fix it

PR #3725, opened 2020-09-11
Fixes for couple bugs

4e4o:
Hi, sorry for my english. I have 3d printer with fw Marlin 2.0.6.1, board - skr 1.3. Octoprint runs on orange pi and when octoprint starts - printer is always off (it is turned on later by SSR relay). At first i was connect board and orange pi with usb and connect/disconnect button in octoprint works fine and without bugs. But then i moved connection from usb to gpio/uart logic and that is where bugs comes. I think this bug appears because direct uart connection is faster than usb. Sometime octoprint hangs when connecting to board, everytime in the same place, see first attached screenshot. After a day of remote debugging i was able to find cause of this deadlock. First two commands that octoprint sends is sayHello commands, inside sayHello you are incrementing clear_to_send counter that checked by sender_thread. By default clear_to_send has max value = 1, so after second sayHello call clear_to_send will have value = 1. Inside sender_thread you have clear_to_send.wait() before send cycle, so if both sayHello call will be executed before this clear_to_send.wait() call, then later in send cycle it will hang forever in second clear_to_send.wait(). So it is race condition. Forgot to say that monitor thread is alive, only sender_thread is deadlocked, i can't send anything, but i receive temperature messages from board. I was able to workaround this bug by setting ok buffer to 2, but i think it is ugly solution, so i fixed it in another way (see commit). Also i found that after disconnecting from printer sender_thread not closed properly and every connect/disconnect cycle causes sender_threads amount to grow up. I fixed this too, see commits. Sorry for my bad english again.

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -1579,12 +1579,15 @@ class MachineCom(object):
 		self._callback.on_comm_sd_state_change(self._sdAvailable)
 		self._callback.on_comm_sd_files(self.getSdFiles())
 
-	def sayHello(self, tags=None):
+	def sayHello(self, tags=None, clear_to_send=True, on_sent_callback=None):
 		if tags is None:
 			tags = set()
 
-		self.sendCommand(self._hello_command, force=True, tags=tags | {"trigger:comm.say_hello",})
-		self._clear_to_send.set()
+		self.sendCommand(self._hello_command, on_sent=on_sent_callback, force=True, tags=tags | {"trigger:comm.say_hello",})
+
+		if clear_to_send == True:
+			self._clear_to_send.set()
+
 		self._hello_sent += 1
 
 	def resetLineNumbers(self, number=0, part_of_job=False, tags=None):
@@ -1708,11 +1711,13 @@ class MachineCom(object):
 
 		# enqueue the "hello command" first thing
 		if try_hello:
-			self.sayHello()
+			self.sayHello(None, True, lambda: self._clear_to_send.set())
 
 			# we send a second one right away because sometimes there's garbage on the line on first connect
 			# that can kill oks
-			self.sayHello()
+			# clear_to_send.set() must be executed right after first hello command,
+			# otherwise race condition can happen and send_thread will block forever
+			self.sayHello(None, False)
 
 		while self._monitoring_active:
 			try:
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -1003,6 +1003,12 @@ class MachineCom(object):
 
 		self._serial = None
 
+		# fix for send_thread leak
+		# this two commands is just for unblocking send_thread
+		# nothing will be sended because send queue is deactivated at this point
+		self._enqueue_for_sending(self._hello_command)
+		self._clear_to_send.set()
+
 		# if we are printing, this will also make sure of firing PRINT_FAILED
 		if is_error:
 			self._changeState(self.STATE_CLOSED_WITH_ERROR)


foosel:
First of all, your English is fine, don't worry about it!

I have a question however. From your screenshot, the second `M110` never gets acknowledged with an `ok` here. The firmware thus doesn't signal that it's free to receive again and communication stalls. If you click Fake Acknowledgement:

image

This should clear up the lock in theory. The question however is why this second `M110` never gets an `ok`. Is this always the case or just a case of this specific instance of the issue you outlined here?

I'm trying to fully understand the race condition here first and want to make sure it's not actually another issue. Changing *anything* about this particular initial handshake procedure can have far reaching consequences for printers out there and I'm not yet convinced that your changes are the correct solution for the problem.

4e4o:
I have just checked Fake Acknowledgement button and it works after click. 

Here is log before click:

Connected to: Serial<id=0xffff5a4b9820, open=True>(port='/dev/ttyS2', baudrate=250000, bytesize=8, parity='N', stopbits=1, timeout=10.0, xonxoff=False, rtscts=False, dsrdtr=False), starting monitor
Send: N0 M110 N0*125
Recv: ok
Send: N0 M110 N0*125
Changing monitoring state from "Connecting" to "Operational"
[...]
Send: N0 M110 N0*125

And log after click:

Recv: Error:No Checksum with line number, Last Line: 0
Recv: Resend: 1
Recv: ok
Send: N1 M115*39
Recv: FIRMWARE_NAME:Marlin 2.0.6.1 (Sep 13 2020 21:09:56) SOURCE_CODE_URL:https://github.com/MarlinFirmware/Marlin PROTOCOL_VERSION:1.0 MACHINE_TYPE:3D Printer EXTRUDER_COUNT:2 UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
Recv: Cap:SERIAL_XON_XOFF:0
Recv: Cap:BINARY_FILE_TRANSFER:0
.....

I think you are right. It seems that it is a uart/gpio issue. How can i ensure this?

4e4o:
I have a cheap china logic analyzer 24 mhz 8ch. I will try to debug uart with it today.

4e4o:
Ok, here is the results. Problem is definitely on the side of orange pi zero plus h5. But i can't be sure that it is hardware or octorpint problem. Steps that i do:

1. Stop octoprint service on orange pi, systemctl stop octoprint.serice
2. Power off skr 1.3.

Logic analyzer shows nothing on tx/rx - ok. (tx/rx relatively to orange pi)

Then:

1. Power on board skr 1.3.

Logic analyzer shows temperature messages on rx. Nothing on tx - ok. Seems good.

Then:

1. Start octoprint service with web page opened. 

Octoprint connects without hangs and here is tx log:

N0 M110 N0*125
N0 M110 N0*125
N0 M110 N0*125
N1 M115*39
M21
M155 S2

**And every first connect after octoprint service restart is always successful**.

Then disconnect/connect in octoprint.

tx log:

N0 M110 N0*125
N0 M1��

Second n0 M110 is broken. And any other disconnect/connect has same issue.

So it is definitely problem on orange pi side. But octoprint service restart helps for first connect to be successful.

foosel:
This only happens with direct UART connection though, right? Not via USB?

What I can tell you is that from OctoPrint's side, once stuff enters the send loop at gets handed over to pyserial, OctoPrint has no influence on the data anymore that I'm aware of. Stuff *should* get sent out correctly. For some reason here it isn't, and weirdly reproducible at that. Would be interesting to know what bytes those two unprintable characters are. It also looks like we didn't just get a bit flip or something like that, but parts of the message got cut off?

4e4o:
Yes, over usb all works fine. And it starts hangs after octoprint restart also. Unprintable characters are random. Sorry for this pull request, this fixes strangely helps for my case. Also i have strange print issues, i think it is all related to uart.

foosel:
What you could do is write a simple single file plugin that fakes an `ok` on connect. Considering that anything but the first initial connect always seems to fail due to the second `M110` being transmitted with garbage bytes in between, you could probably just work around this easily by having a plugin that counts the connect events since start up and in case of anything after the first one starts a 10 or so second timer on connect and calls `self._printer.fake_ack` one timer completion. That will at least get you to connect reliably in your specific setup. But if you also have other issues with transmitted data during prints you should probably rather investigate and fix those.

If you don't mind I'm going to close this PR for now - if your investigation unearthes an actual issue on OctoPrint's side after all we can always reopen it. But merging it could break a LOT of stuff for other people out there.

PR #3727, opened 2020-09-13
Add terminal filter for busy protocol

Sophist-UK:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR adds an additional matching string to the Suppress Wait Response terminal message filter to filter out messages of the form: "Recv:echo: processing".

This was tested on my 3D printer. To test it you simply need to create a move which takes a long time - say a move from one extreme to the other on the z-axis. Test this both without the filter and using the existing filter and you should see several "Recv:echo: processing" messages appear in the terminal window. Change the filter regex, reload the OctoPrint browser window, and retry and these messages should be filtered.

No

N/a

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -127,6 +127,7 @@ date of first contribution):
   * [Brian Vanderbusch](https://github.com/LongLiveCHIEF)
   * [Charlie Powell](https://github.com/cp2004)
   * [Ollis Git](https://github.com/OllisGit)
+  * [Sophist](https://github.com/Sophist-UK)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/docs/configuration/config_yaml.rst
+++ b/docs/configuration/config_yaml.rst
@@ -1122,6 +1122,8 @@ Use `Javascript regular expressions <https://developer.mozilla.org/en/docs/Web/J
      regex: '(Send: (N\d+\s+)?M27)|(Recv: SD printing byte)'
    - name: Suppress wait responses
      regex: 'Recv: wait'
+   - name: Suppress processing responses
+     regex: 'Recv: (echo:\s*)?busy:\s*processing'
 
 .. _sec-configuration-config_yaml-webcam:
 
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -382,7 +382,8 @@ default_settings = {
 	"terminalFilters": [
 		{ "name": "Suppress temperature messages", "regex": r"(Send: (N\d+\s+)?M105)|(Recv:\s+(ok\s+((P|B|N)\d+\s+)*)?(B|T\d*):\d+)" },
 		{ "name": "Suppress SD status messages", "regex": r"(Send: (N\d+\s+)?M27)|(Recv: SD printing byte)|(Recv: Not SD printing)" },
-		{ "name": "Suppress wait responses", "regex": "Recv: wait"}
+		{ "name": "Suppress wait responses", "regex": r"Recv: wait" },
+		{ "name": "Suppress processing responses", "regex": r"Recv: (echo:\s*)?busy:\s*processing" }
 	],
 	"plugins": {
 		"_disabled": [],


jneilliii:
Screenshot_20200913-100222_Chrome

Why not just add this to your terminal filters, that's what they are there for...?

Sophist-UK:
@jneilliii I have - the purpose of this PR is to make the fix available to everyone else.

foosel:
I'm happy to merge this, however it needs to be a separate entry than "wait" responses (those are NOT the same!) and it should capture all busy responses, not just the processing ones IMHO.

Sophist-UK:
@foosel 

1. What is the difference from a user perspective, rather than a technical perspective, between a wait response and a busy response? If there is no difference, then shouldn't there be a single filter for them both?

2. What other busy responses are there? Can you point me to a list?

cp2004:
Wait means the buffer is full, busy means that there is a long command eg. Homing, or probing, heating, etc.

You could just filter based on `busy` and ignore the processing bit?

foosel:
> Wait means the buffer is full

Wrong, but an easy mistake to make :)

`wait` means: Hi, I no longer have any planned moves in my buffer, if you are currently printing something is wrong and you might want to send more data, thanks.

`busy` means: Hi, I'm currently busy AF, please don't send any more commands until I tell you it's `ok` to do so, thanks.

So completely the opposite situation. Lumping them together is just wrong, especially since you should usually never see one (`wait`) while printing unless you have some severe data corruption going on.

I'd suggest to add "Suppress busy responses" with pattern `Recv:\s*echo:\s*busy:`

Sophist-UK:
Ok. I can see that these are different enough to warrant a separate filter - I will make this change.

(Ideally we might want to apply the wait filter only when OctoPrint believes that the printer is idle - but that sort of logic is outside the scope of this PR.)

foosel:
None of these filters will ever be enabled unless the user explicitly enabled them (and frankly, if I wasn't sure it would make a lot of people very angry I'd get rid of this whole filtering since it makes a lot of stuff way more complicated than it needs to be and has masked up firmware issues in the past when all users shared was a terminal screenshot) and the filtering will stay strictly frontend side, so I'd rather not start to build in code that tries to intelligently toggle them in this or another PR ;)

Sophist-UK:
Exactly - automatic toggling is way too complicated.

Sophist-UK:
Hopefully the regex will pick up `Recv:echo: busy now` but not `Recv:echo: unbusy` or `Recv:echo: busyest`.

I have tested this in my live OctoPrint but not run unit tests (which don't exist for wait or busy filters) or compiled OP to test it in a new install,

Sophist-UK:
On reflection perhaps I should have used `\w`.

foosel:
Just use `Recv: (echo:\s*)?busy:` (making the `echo:` optional contrary to my initial proposal up there, and making the whitespace after `Recv:` static as that gets injected by OctoPrint anyhow).

There's no reason for checking for prefixes other than `echo:\s*`, see the [reprap wiki](https://reprap.org/wiki/G-code#Replies_from_the_RepRap_machine_to_the_host_computer) (that's the closest that we have to a standard) and it would be wrong to filter anything that doesn't match (for that users can add their own filters if they want, we only want to support the actual busy protocol here, nothing more, nothing less). This `echo:` thing there isn't actually even supposed to be there, but most firmware still have it in (without whitespace mind you) and thus OctoPrint supports it too. 

It also suffices to match only parts of the string - we don't do a full match here but only "is this pattern in this line".

Your regex would currently match `echo: here's some additional text and maybe a busy: processing` but not `busy: processing`, and that's incorrect.

image

image

As a reference, this is the busy protocol implementation in OctoPrint: https://github.com/OctoPrint/OctoPrint/blob/bdb68ae28b98beac18ba9544da553ae325a0f1db/src/octoprint/util/comm.py#L1728-L1752

Sophist-UK:
Hmm - according to the reprap wiki:
> Examples: busy: processing, busy: paused for user

[Marlin v2](https://github.com/MarlinFirmware/Marlin/blob/9913333c33d234264a9072630e8b5ffbf07bad7e/Marlin/src/core/language.h#L135) defines the following strings: 
```
"busy: processing"
"busy: paused for user"
"busy: paused for input"
```
It seems to me that a paused string is of more interest to users than processing, so IMO we should revert back to filtering only `busy: processing`.

If there is agreement on this, I will make this further change and we should be good to go.

Sophist-UK:
I have pushed a further commit to show my proposed change. Once there is agreement I will squash them to a single commit again.

foosel:
Turn `Recv:\s*` back into `Recv: ` on both wait and processing filter and we have a deal. As I said, the whitespace there is under the control of OctoPrint, it will always be one space, and leading whitespace on the received line itself is not protocol conform and thus if it DOES happen (and consequently breaks communication in general) it should not be covered up by filters.

foosel:
Perfect, thank you! :)

PR #3817, opened 2020-11-14
update gcode_scripts.rst with correct UI path for print & resume

dh1tw:
Updating one line in the documentation to be coherent with the User Interface

  * [X] Your changes are not possible to do through a plugin and relevant
    to a large audience (ideally all users of OctoPrint)
  * [X] If your changes are large or otherwise disruptive: You have
    made sure your changes don't interfere with current development by
    talking it through with the maintainers, e.g. through a
    Brainstorming ticket
  * [X] Your PR targets OctoPrint's devel branch if it's a completely 
    new feature, or maintenance if it's a bug fix or improvement of
    existing functionality for the current stable version (no PRs 
    against master or anything else please)
  * [X] Your PR was opened from a custom branch on your repository
    (no PRs from your version of master, maintenance or devel please),
    e.g. dev/my_new_feature or fix/my_bugfix
  * [X] Your PR only contains relevant changes: no unrelated files,
    no dead code, ideally only one commit - rebase and squash your PR 
    if necessary!
  * [no code] Your changes follow the existing coding style
  * [n/a] If your changes include style sheets: You have modified the
    .less source files, not the .css files (those are generated with
    lessc)
  * [n/a] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [n/a] You have run the existing unit tests against your changes and
    nothing broke
  * [n/a] You have added yourself to the AUTHORS.md file :)

#### What does this PR do and why is it necessary?
Small update to the docs to be coherent with the octoprint Settings UI for the `Log position on pause` option.

#### How was it tested? How can it be tested by the reviewer?
n/a

#### Any background context you want to provide?
n/a

#### What are the relevant tickets if any?
n/a

#### Screenshots (if appropriate)
n/a

#### Further notes
n/a

Issue #3799, opened 2020-11-07
[Request] Include full jQuery UI module to avoid conflicts by plugins


jneilliii:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when another plugin causes a conflict with one of my plugins because of jQuery UI modules that have been added in order to have the additional functionality of those modules that are included by OctoPrint by default.

**Describe the solution you'd like**
Include the full jQuery UI module with all options so that plugin authors don't have to include their own individual bits and pieces.

**Describe alternatives you've considered**
Not sure if there is another way to get around this outside of possibly scoping. Not really familiar with how to do this and if it's even possible.

**Additional context**
https://github.com/jneilliii/OctoPrint-ConsolidatedTabs/issues/32

foosel:
Hm... that's a bit of a tricky one. If I could I frankly would throw it out altogether, the only reason it's (partially) included is due to the jquery file upload dependency needing it. And if I were to bundle it now that would then only cause issues for plugins that include their own versions, wouldn't it?

I fear this is a problem not easily solved...

jneilliii:
I don't know if it would cause issues, because if the plugins are already bundling it, their implementation would still overwrite the existing one, depending on load order of when it's bundled, or added to the page if not bundled? The idea is that if the full one was included plugin's wouldn't have to include their own. Just thinking in regard to the number of plugins that I have bundled a full UI module the more plugins you install the more the bloat increases. If I didn't have to bundle, then the combined size of multiple plugin installs would get reduced drastically. 

I'm more than happy to try and figure out how to alias jquery and the jquerui within my own plugins (any guidance/links appreciated), but then the bloat would still potentially be there.

PR #3818, opened 2020-11-15
[WIP] changes the backup filename to be based on appearance > name setting,…

jneilliii:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

When managing multiple printers, having the ability to identify which machine it came from is very useful. Especially if they are getting uploaded somewhere utilizing the new backup created hook.

Installed locally in development environment and ran the process of backing up.

Discussion on Discord

image

first listed backup is without a name configured, the second is with a name entered.

#### Further notes

Would love to be able to include this as part of 1.5.0 if possible, as it will surely come up with the ability to use Google Drive Backup plugin.


jneilliii:
Not sure why prettier failed, but appears to have happened on files I didn't touch...I did run the pre-commit script prior to posting.

cp2004:
Prettier has failed in #3817 and #3815 too, strange since it has passed for me locally too on all of them. 🤷

src/octoprint/static/js/lib/jquery/jquery-ui.js
```diff
@@ -1,6 +1,6 @@
-/*! jQuery UI - v1.12.1 - 2020-11-03
+/*! jQuery UI - v1.12.1 - 2020-11-07
 * http://jqueryui.com
-* Includes: widget.js, widgets/mouse.js
+* Includes: widget.js, position.js, data.js, disable-selection.js, focusable.js, form-reset-mixin.js, jquery-1-7.js, keycode.js, labels.js, scroll-parent.js, tabbable.js, unique-id.js, widgets/draggable.js, widgets/droppable.js, widgets/resizable.js, widgets/selectable.js, widgets/sortable.js, widgets/accordion.js, widgets/autocomplete.js, widgets/button.js, widgets/checkboxradio.js, widgets/controlgroup.js, widgets/datepicker.js, widgets/dialog.js, widgets/menu.js, widgets/mouse.js, widgets/progressbar.js, widgets/selectmenu.js, widgets/slider.js, widgets/spinner.js, widgets/tabs.js, widgets/tooltip.js, effect.js, effects/effect-blind.js, effects/effect-bounce.js, effects/effect-clip.js, effects/effect-drop.js, effects/effect-explode.js, effects/effect-fade.js, effects/effect-fold.js, effects/effect-highlight.js, effects/effect-puff.js, effects/effect-pulsate.js, effects/effect-scale.js, effects/effect-shake.js, effects/effect-size.js, effects/effect-slide.js, effects/effect-transfer.js
```
cp2004, line 3: Should this have been checked in? Or is it a way to sneak in #3799? 😜

src/octoprint/plugins/backup/__init__.py
```diff
@@ -129,6 +123,17 @@ def get_template_configs(self):
...
 
+    ##~~ Utility Methods
+
+    def build_backup_filename(self):
+        if self._settings.global_get(["appearance", "name"]) == "":
+            backup_prefix = "octoprint"
+        else:
+            backup_prefix = self._settings.global_get(["appearance", "name"])
+        return "{}-backup-{}.zip".format(
+            backup_prefix, time.strftime(BACKUP_DATE_TIME_FMT)
```
cp2004, line 134: When backed up via CLI, the settings aren't injected so this fails.
Suggest adding a way that this can be specified in the arguments, like:
```python
def build_backup_filename(self, backup_prefix=None):
        if not backup_prefix:
                if self._settings.global_get(["appearance", "name"]) == "":
                        backup_prefix = "octoprint"  # Could put the default back above, so it can be used in CLI hook
                else:
                        backup_prefix = self._settings.global_get(["appearance", "name"])
        return "{}-backup-{}.zip".format(
            backup_prefix, time.strftime(BACKUP_DATE_TIME_FMT)
```
and then specify the default prefix in the CLI hook below. There might be another way to tell, but either way this needs fixing.

cp2004 (changes requested):
CLI backup fails now since the settings are not injected. See comments for suggestions 🙂

src/octoprint/static/js/lib/jquery/jquery-ui.js
jneilliii: oops...my bad. wasn't paying attention when I committed. Since the CLI is broken, I'll revert that and re-commit and update title to WIP.

cp2004 (approved):
👍

src/octoprint/plugins/backup/__init__.py
```diff
@@ -129,6 +123,18 @@ def get_template_configs(self):
             {"type": "wizard", "name": gettext("Restore Backup?")},
         ]
 
+    ##~~ Utility Methods
+
+    def build_backup_filename(self, backup_prefix=None):
```
foosel, line 128: As this is being used from both CLI and full context, this should better be a class method that takes a `settings` object as parameter and operates on that. 

`self._settings` will not be defined in case of the CLI context, so it should not be access here. Turning the method from instance into class method with explicit `settings` parameter makes that clearer.

src/octoprint/plugins/backup/__init__.py
```diff
@@ -129,6 +123,18 @@ def get_template_configs(self):
...
+        if not backup_prefix:
+            if self._settings.global_get(["appearance", "name"]) == "":
+                backup_prefix = "octoprint"
+            else:
+                backup_prefix = self._settings.global_get(["appearance", "name"])
```
foosel, line 133: Hm... I'm a bit worried about using the instance name here without any kind of sanitizing. What if there's a `/` in there or anything else? Should at least be run through basic sanitization or slugification.

foosel (changes requested):

foosel:
> Would love to be able to include this as part of 1.5.0 if possible, as it will surely come up with the ability to use Google Drive Backup plugin.

I frankly don't feel comfortable with this at this point any more. Consider 1.5.0 frozen. I'm fine with fixing bugs or blatant UX issues in newly introduced functionality, and even with minimally invasive fixes for issues that aren't regressions, but that is quite an invasive change to do on an already started RC phase and thus will have to wait until 1.6.0. I'll merge it gladly though once the above issues have been resolved with it.

jneilliii:
No worries Gina, I'll try to re-work the code as you suggest.

src/octoprint/plugins/backup/__init__.py
jneilliii: I think I have updated this to what you mean, sorry for my lack of proper python programming.

src/octoprint/plugins/backup/__init__.py
jneilliii: added slugify implementation (ripped from filemanager storage) and that seems to work properly now. Anything with slashes get converted to `-`.

jneilliii:
@cp2004 if you get a chance if you could check my changes that would be great. The slugify might be overkill with emoji stuff, but it was part of the filemanager storage one and figured it would make sense to just reuse it. If there's a better approach here to import that versus re-create, please let me know.

cp2004:
Apologies, saw the initial request and promptly forgot about re-review.

To the eye, it looks ok, not had a chance to test it. Going to pass on review to @foosel since I think she'll be able to spot any problems in this better than I can... 🙂

foosel:
Refactored it a bit further by moving the sanitization for file storage, backup and announcements into its own util method, now it's all a bit less repetitive.

PR #3837, opened 2020-11-28
update Trigger docker build workflow

LongLiveCHIEF:
#### Change Description

set payload event-type key to use value from event.action

#### Change Summary

This amends the `trigger-docker` release so that the downstream build can distinguish between a release and a pre-release.  The downstream octoprint docker release workflow has been amended so that a `prereleased` event from octoprint will now build and publish a prereleased tagged docker image. 

This downstream workflow additionally has been setup so only _stable_ releases will move the `octoprint/octoprint:latest` tag.

This pull request is a paired PR for https://github.com/OctoPrint/octoprint-docker/pull/105, which closes https://github.com/OctoPrint/octoprint-docker/issues/102


cp2004:
Sorry @GitIssueBot, you'll have to stay upset for now since this *should* be going to master.

Issue #3544, opened 2020-04-29
[Request] - "Download Selected" in Timelapse


KenLucke:
Can you please add a button to "Download All Selected" timelapse videos on the Timelapse tab, instead of having to do them one at a time (unless I'm missing an existing way to do this).  The files could be selected in the already existing way as they are for the "Delete" function but instead, it would allow all selected ones to be pulled at the same time.

foosel:
1.6.0 has just been released including this.

Issue #3746, opened 2020-09-23
[Request] Download multiple logs at once.


KenLucke:
**Is your feature request related to a problem? Please describe.**
Annoyance rather than problem.  When downloading logfiles and backups, you have to do so one at a time.

**Describe the solution you'd like**
There already exists a mechanism for selecting multiple files at one time (for deletion).  Please add the ability to download the selected files all at once (preferably with an option to zipfile them prior to download, but just a bulk download would make me ecstatic).  An optional "Save To" folder on the local machine (instead of just automatically going into the Downloads folder) would be tremendous, too.

**Describe alternatives you've considered**
Continuing to do it the old, PITA way :)

**Additional context**

foosel:
1.6.0 has just been released.

PR #3856, opened 2020-12-04
Add bulk download for timelapses

Master92:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR targets issue #3544 by adding a button for downloading multiple timelapses at once. This is realized using a JavaScript zipping (using [JSZip](https://github.com/Stuk/jszip)) library, locally creating an archive of all selected timelapse files. This archive will then be downloaded to the client computer (with [FileSaver.js](https://github.com/eligrey/FileSaver.js/)).

Tested with Firefox 83.0, Google Chrome 87.0.4280.88 and Microsoft Edge 44.17763.831.0

Issue #3544

#### Further notes
The added libraries are both licensed under the MIT license. JSZip could also be used with GPLv3. I am unsure if adding libraries to this project is allowed, so please review.
Further, I don't know if translation to German is desired in a PR, so I left that out intentionally.


cp2004:
This is interesting, and we might be able to be apply it to #3746 as well (multiple log downloads). I'll give it a test, but would be interesting to know if downloading bulk time lapses should be disabled while the printer is printing. It might *potentially* cause stuttering etc. since time lapses can be quite large files to read off the disk. Don't know if you tried this?

cp2004:
> I am unsure if adding libraries to this project is allowed, so please review.
 
Forgot about this bit - might want to add a link to the licenses in [THIRDPARTYLICENSES.md](https://github.com/OctoPrint/OctoPrint/blob/master/THIRDPARTYLICENSES.md). 

In the end, it's not up to me (🙂) if the dependencies should be added, but if they are useful (& it is not a simple task *without* them), then I don't see a problem.

Master92:
> might want to add a link to the licenses in [THIRDPARTYLICENSES.md](https://github.com/OctoPrint/OctoPrint/blob/master/THIRDPARTYLICENSES.md).

Alright, will do this as soon as I get to my Computer again. By the way: what's the preferred way of doing these kind of patches - Straight forward pushes to my brach or amending and force pushing?

> It might potentially cause stuttering etc. since time lapses can be quite large files to read off the disk. Don't know if you tried this?

In fact, I didn't try that but I suppose it wouldn't hurt disabling the feature. This way you won't run into troubles in the first place. I'll have a look into this when adding the link to the licenses.
How are downloads of single timelapses handled when printing? Couldn't that potentially cause stuttering as well?

cp2004:
>Straight forward pushes to my brach or amending and force pushing?

Personal preference I suppose. The template says as few commits as possible, but leaving a more specific paper trail can be quite helpful to figure out what the intention was. Going to go with 'I don't know' since no one's asked that before. Question for @foosel when she comes to merge it I guess.

> How are downloads of single timelapses handled when printing? Couldn't that potentially cause stuttering as well?

~~I know the *configuration* of timelapses is disabled, but unsure about downloading~~ just checked while typing, and you can still download. It's a good point, although with single downloads I can see it being less of an issue but would have to test against my *actual* printer so it can show up any stuttering etc.

Master92:
> Personal preference I suppose. The template says as few commits as possible, but leaving a more specific paper trail can be quite helpful to figure out what the intention was.

Okay, I'll stick to adding new commits because I see it the same way and we use to do so at work. 👍
More code tomorrow...

Master92:
Alright, I modified my first commit containing the JS libraries because I thought adding the libraries to the markdown file would fit right in there.

The other change regarding disabling the bulk download got a new commit on its own because it changed the original behavior.

I think moving and adding this feature to other places for downloading multiple files should be a PR on its own, but I'll look into this very soon.

Master92:
@cp2004: FYI, I generalized the bulk downloading function and applied it to logs and backups as well. The result can be seen [here](https://github.com/Master92/OctoPrint/tree/improve/bulk_download_files). I suppose hijacking this PR is not desired, so I'll open a new one as soon as this one got merged.

@foosel: Or would you prefer closing this PR and opening a new one?

foosel:
> @foosel: Or would you prefer closing this PR and opening a new one?

Let's take care of this one first, and then we can tackle the next one. Certainly welcoming both 👍, just tricky to find the time for reviews these days 😅

src/octoprint/static/js/app/viewmodels/timelapse.js
```diff
@@ -377,6 +377,85 @@ $(function () {
...
+                self.requestData();
+            });
+
+            promise.then(function () {
+                zip.generateAsync({
+                    type: "blob",
+                    compression: "STORE"
+                }).then(function (blob) {
+                    saveAs(blob, "OctoPrint-Timelapses.zip");
+                    dialog.modal("toggle");
```
foosel, line 452: Better be explicit here and close the dialog instead of toggling it.

foosel (commented):
Before I properly review this (first look was great apart from on minor issue, `dialog.modal("toggle")` instead of explicit `dialog.modal("hide")`) and as much as I'd love to see bulk download get added, I'm currently a bit worried about memory consumption. What happens if someone tries to zip up 10 1GB files here? Timelapses can be big... Haven't yet gotten to actually testing this (too much going on), but maybe you have? 

Also a bit unsure about the performance while printing, that would need some testing first. It will probably be less of a problem though than if we were to do this via some kind of bulk download endpoint on the backend, which would have to be the alternative if the memory consumption turns out to be problematic.

src/octoprint/static/js/app/viewmodels/timelapse.js
Master92: I am unsure about what exactly you mean. Do you mean to "hide" the dialog or "hide" and "dispose"ing it? Because the latter fails with an `Uncaught (in promise) TypeError: data[option] is undefined`.

foosel:
@Master92 I'm seeing a significant delay between progress of 100% and a download offering of the resulting zip file here, even with just 30MB of timelapse files split across three files it took a bit over 30s. I started a process with around a GB of files over 10min ago and it's still ongoing without any indication that it's actually doing anything.

So... I honestly don't feel like client-side is the best approach here. Even if we make it display a "please wait" status during `zip.generateAsync`, it feels slow. Firefox's task manager says it's not really utilizing any CPU (and also not much memory, so that's at least something), so whatever it is doing, it seems to do more waiting than anything. Plus whenever I switch over to the task manager tab, CPU consumption drops down to pretty much nothing, which makes me think it's probably also suspending zipping if the tab is not in focus (or maybe the browser is pausing the background process).

Can you confirm this observation? If so, I don't think this is the way and we need to do this on the server side after all. I'll need to build an zip endpoint anyhow soon for another feature, so maybe that could be used here as well...

foosel:
I've switched this over to server side zipping with a streamed approach in tornado, and it seems to perform well enough now.

PR #3858, opened 2020-12-05
PR: This enables moving the webcam to another [its own] tab, without having to …

gruvin:
…clone and override `_enableWebcam()`, which although it works, makes it difficult/messy for Webcam Tab plugin to support future versions of Octoprint. Currently, Webcam Tab plugin needs to clone and override three different versions of `_enableWebcam()` to work with Octoprint v1.3, 1.4 and 1.5. Totally open to a better way to do it, if there is one. Thanks.

  * [Y -- technically doable via plugin but very messy. See notes below.] Your changes are not possible to do through a plugin and relevant to a large audience (ideally all users of OctoPrint)
  * [Y] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
  * [Y] Your PR targets OctoPrint's devel branch if it's a completely
  * [Y] Your PR was opened from a custom branch on your repository
  * [Y] Your PR only contains relevant changes: no unrelated files,
    if necessary!
  * [Y?] Your changes follow the existing coding style
  * [NA ] If your changes include style sheets: You have modified the
    lessc)
  * [Y -- manually. Trivial change.] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [NA -- trivial change (famous last words!)] You have run the existing unit tests against your changes and
    nothing broke
  * [N] You have added yourself to the AUTHORS.md file :)

Manually. It does not change any behaviour. The reviewer should see no apparent difference what-so-ever. 

The plugin Webcam Tab stopped working when Octoprint migrated from v1.3 to 1.4. The cause is that the only known method to move #webcam_container to another tab is to copy/paste `_enableWebcam` and `onTabChange` methods from `control.js` and then back reference to override and change the code, as affected in this PR. (should be obvious?)  

N/A

N/A

#### Further notes
The original author of Webcam Tab plugin is no longer reachable. I am forking his project to get it working again with Octoprint v1.4+. (This single plugin prevented me upgrading from v1.3 until now.)

If this change is deemed inappropriate (easy to see why it might) then I am willing to copy/paste all three versions of `_enableWebcam()` from Octoprint 1.3, 1.4 and 1.5 and select the appropriate version in the plugin. Thus, _technically_, the change _could_ be done entirely in the plugin.  However, it's messy and not at all future proof. I realise this is an edge case. Obviously, it's all up to you. Thanks.

Oh and the plugin is already python3 compatible (tested).

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -542,7 +542,7 @@ $(function () {
         };
 
         self.onTabChange = function (current, previous) {
-            if (current == "#control") {
+            if (current == "#control" || current == "#webcam") {
                 self._enableWebcam();
             } else if (previous == "#control") {
                 self._disableWebcam();
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -503,7 +503,10 @@ $(function () {
 
         self._enableWebcam = function () {
             if (
-                OctoPrint.coreui.selectedTab != "#control" ||
+                (
+                 OctoPrint.coreui.selectedTab != "#control" &&
+                 OctoPrint.coreui.selectedTab != "#webcam"
+                ) ||
                 !OctoPrint.coreui.browserTabVisible
             ) {
                 return;
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -503,11 +503,9 @@ $(function () {
 
         self._enableWebcam = function () {
             if (
-                (
-                 OctoPrint.coreui.selectedTab != "#control" &&
-                 OctoPrint.coreui.selectedTab != "#webcam"
-                ) ||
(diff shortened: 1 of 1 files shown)


gruvin:
Was too rushed. I'll be back.

jneilliii:
I've had to deal with these overrides before in my plugins (Consolidated Tabs, Consolidated Temp Control) and actually did a fork of webcamtab and linked to a couple of people already. Although this would resolve the issue for this specific plugin/tab, it would not resolve the issue for all plugins, which would make more sense long term and be more helpful. So if there was a way to potentially generalize this more I think it would be more helpful to all, including UI Customizer who also manipulates the webcam location.

https://github.com/jneilliii/OctoPrint-WebcamTab

gruvin:
@jneilliii agreed. However, my thinking was that the webcam feature is a stock component of Octoprint and as such is kinda of "special". Perhaps it should simply be refactored out of dependendence on control.js altogether?

That said, it's hard to come up with a scheme that doesn't make Octoprint itself just one giant repository of plugins. In any case, I found my version PR was flawed [don't take shortcuts] so I've withdrawn it for now.

PR #3878, opened 2020-12-11
Validate websocket JSON before sending

### LINKED ISSUE #3875: additional_state_data is broken with invalid JSON
#### What were you doing?

Trying to use the additional_state_data hook to send a dict with a value that is `float("inf")`.

1. Create a plugin in the normal way.
2. Add to the hooks: `"octoprint.printer.additional_state_data": __plugin_implementation__.additional_state_data`
3. Add this function:

```python
  def additional_state_data(self, initial, *args, **kwargs):
    return {"x": float("inf")}
```
#### What did you expect to happen?

I expected that the print time left would still be displayed on screen.

#### What happened instead?

Print time left is no longer displayed on screen.  Printer is crippled, unable to print from GUI.

#### Did the same happen when running OctoPrint in safe mode?

I didn't try this because this bug can only happen if plugins are enabled.

#### Version of OctoPrint

OctoPrint 1.5.1 Python 3.7.7

#### Operating System running OctoPrint

Linux

#### Printer model & used firmware incl. version

Virtual printer

#### Browser and version of browser, operating system running browser

Chrome

#### Screenshot(s)/video(s) showing the problem:

I have read the FAQ.

eyal0:
The default json library in python will send infinity values as `Infinity` as described [here](https://docs.python.org/3/library/json.html#json.dump).  Check that the JSON is strictly valid before sending it, otherwise throw an exception.

This fixes #3875

    Brainstorming ticket
  * [ ] Your PR targets OctoPrint's devel branch if it's a completely
    if necessary!
    lessc)
    have added unit tests
    nothing broke

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -18,6 +18,7 @@ import time
 from frozendict import frozendict
 from past.builtins import basestring, long
 
+import octoprint.util.json
 from octoprint import util as util
 from octoprint.events import Events, eventManager
 from octoprint.filemanager import FileDestinations, NoSuchStorage, valid_file_type
@@ -272,7 +273,13 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
             try:
                 additional = hook(initial=initial)
                 if additional and isinstance(additional, dict):
+                    octoprint.util.json.dump({name: additional})
                     plugin_data[name] = additional
+            except ValueError:
+                self._logger.exception(
+                    "Invalid additional data from plugin {}".format(name),
+                    extra={"plugin": name},
+                )
             except Exception:
                 self._logger.exception(
                     "Error while retrieving additional data from plugin {}, blacklisting it for further loops".format(
--- a/src/octoprint/server/util/sockjs.py
+++ b/src/octoprint/server/util/sockjs.py
@@ -589,6 +589,9 @@ class PrinterStateConnection(
 
     def _do_emit(self, type, payload):
         try:
+            import json
+            
+            _ = json.dumps(octoprint.util.thaw_frozendict({type: payload}), allow_nan=False)
             self.send({type: payload})
         except Exception as e:
             if self._logger.isEnabledFor(logging.DEBUG):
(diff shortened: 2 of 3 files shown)


eyal0:
Tested this with the same procedure as described in the PR.  The result is now warnings in the log:

```
2020-12-10 21:31:49,248 - octoprint.server.util.sockjs - WARNING - Could not send message to client 127.0.0.1: Object of type frozendict is not JSON serializable
2020-12-10 21:31:49,331 - octoprint.server.util.sockjs - WARNING - Could not send message to client 127.0.0.1: Out of range float values are not JSON compliant
2020-12-10 21:31:49,833 - octoprint.server.util.sockjs - WARNING - Could not send message to client 127.0.0.1: Out of range float values are not JSON compliant
2020-12-10 21:31:50,334 - octoprint.server.util.sockjs - WARNING - Could not send message to client 127.0.0.1: Out of range float values are not JSON compliant
```

I'm not sure why the first one is different from the rest.

eyal0:
That first error is causing a problem.  Not sure why.  I think that it's some interaction between frozendict and regular dict.

cp2004:
To fix the CI check errors, run `pre-commit run --hook-stage manual --all-files`

Full details about pre commit are in step 9 of the [contributing guidelines]( https://github.com/OctoPrint/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests).

foosel:
I improved on this slightly. We already do custom json encoding for the socket anyhow, I added the `allow_nan=False` there, and then went a step further and also added json validation on the data fetch from the plugins - we don't want to stop sending state updates to a connected client just because a plugin injected garbage data, so now that data will be caught right at the source and filtered out.

eyal0:
> I improved on this slightly. We already do custom json encoding for the socket anyhow, I added the `allow_nan=False` there, and then went a step further and also added json validation on the data fetch from the plugins - we don't want to stop sending state updates to a connected client just because a plugin injected garbage data, so now that data will be caught right at the source and filtered out.

Yes, that's way better. My solution was half-broken. I'm glad this is resolved.

Issue #2040, opened 2017-07-27
[Request] How to set up multiple WiFi networks to work with Octopi?


milkman333:
I have been travelling back and forth with my printer a lot lately and every time I unpack and setup the printer, I have to remove the SD card from the octopi and change the SSID login info, save, re-insert and power on. Not terrible but also less than ideal.

Is there a way to store two different sets of WiFi network credentials so that I don't have to keep editing octopi-network.txt every time I want to connect to the other network?

Hope that makes sense, let me know what I can do to help

GitIssueBot:
Hi @milkman333,

It looks like there is some **information missing** from your bug report that will be needed in order to solve the problem. Read the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) which will provide you with a template to fill out here so that your bug report is ready to be investigated (I promise I'll go away then too!).

If you did not intend to report a bug but wanted to **request a feature or brain storm** about some kind of development, please take special note of the title format to use as described in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md).

**Please do not abuse the bug tracker as a support forum** - if you have a question or otherwise need some kind of help or support refer to the [Mailinglist](https://groups.google.com/group/octoprint) or the [G+ Community](https://plus.google.com/communities/102771308349328485741) instead of here.

Also **make sure you are at the right place** - this is the bug tracker of the official version of OctoPrint, not the Raspberry Pi image OctoPi nor any unbundled third party OctoPrint plugins or unofficial versions. Make sure too that you have **read through the [Frequently Asked Questions](https://github.com/foosel/OctoPrint/wiki/FAQ)** and searched the [**existing tickets**](https://github.com/foosel/OctoPrint/search?q=&ref=cmdform&type=Issues) for your problem - try multiple search terms please.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next two weeks (until 2017-08-10 17:50 UTC) I'll close this ticket so it doesn't clutter the bug tracker. This is nothing personal, so please just be considerate and help the maintainers solve this problem quickly by following the guidelines linked above. Remember, the less time the devs have to spend running after information on tickets, the more time they have to actually solve problems and add awesome new features. Thank you!

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being, so don't expect any replies from me :) Your ticket is read by humans too, I'm just not one of them.

[...]

PR #3881, opened 2020-12-12
Fix stable releases not triggering docker build

LongLiveCHIEF:

This PR sends the _type_ of the release action downstream, instead of just the release event_name, which will always resolve to `release`.  The downstream Deploy Images workflow listens for `released` and `prereleased` event types for the `dispatch_releases` job of the Trigger Docker Build Workflow, and it was not matching since only `release` was being sent.

This time i created a repository exclusively for exploring github.event contexts, to ensure the right value was being sent downstream. You can see the relevant test log here: https://github.com/LongLiveCHIEF/github-actions-testing/runs/1542669636?check_suite_focus=true#step:2:168

This one gave me enough trouble about docs for workflows that I finally figured out what was missing that was bugging me, so I opened an issue with github/docs [#2040 | Please provide complete reference schema for each subkey of the context accessible in a workflow](https://github.com/github/docs/issues/2040)

Closes https://github.com/OctoPrint/octoprint-docker/issues/139

### DIFF
--- a/.github/workflows/trigger_docker.yml
+++ b/.github/workflows/trigger_docker.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           token: ${{ secrets.DOCKER_ACCESS_TOKEN }}
           repository: ${{ github.repository_owner }}/octoprint-docker
-          event-type: ${{ github.event_name }}
+          event-type: ${{ github.event.action }}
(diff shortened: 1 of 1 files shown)


cp2004:
Now we just need another hotfix to test it out 😉

PR #3899, opened 2020-12-19
change docs setting userFile to userfile

LongLiveCHIEF:
changes `userFile` to `userfile` in the docs, since this setting is treated as a single word in the source code.

See https://github.com/OctoPrint/OctoPrint/blob/f67c15a9a47794a68be9aed4f2d5a12a87e70179/src/octoprint/settings.py#L388  

See https://github.com/OctoPrint/OctoPrint/blob/b94424411a72b183e3be1cc99fa00ac26f98a56c/src/octoprint/access/users.py#L508

Currently, setting `userFile` in the docs fails silently/just leaves the setting set to `None`.


foosel:
Cherry-picked to `master` as 04c8d7a888d8d726fc6cdc34e78cae3c22225a51

PR #3916, opened 2020-12-31
Fix firmware info splitting

cp2004:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke *err*

This PR adjusts the regex used to split the comm layer's firmware info parsing

Not yet, that's why this is a draft 🙂 The current unit tests fail because they test keys with numbers in.  The regex may need to be adjusted further, for now this is draft. Tests are currently failing

https://github.com/cp2004/OctoPrint-EEPROM-Marlin/issues/21
https://github.com/cp2004/OctoPrint-EEPROM-Marlin/issues/16#issuecomment-744716312
https://discord.com/channels/704958479194128507/705047010641838211/788136463732047932

See above

#### Further notes

~~This PR is currently draft while I have a think (or anyone else, open to suggestions!) of a better way to work with the numbers.~~ Just here as a placeholder since I told people reporting the issue it will be fixed in OctoPrint 1.6.0 😬

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -182,7 +182,7 @@ Groups will be as follows:
   * ``value``: reported position value
 """
 
-regex_firmware_splitter = re.compile(r"\s*([A-Z0-9_]+):\s*")
+regex_firmware_splitter = re.compile(r"\s*([A-Z_]+):\s*")
 """Regex to use for splitting M115 responses."""
 
 regex_resend_linenumber = re.compile(r"(N|N:)?(?P<n>%s)" % regex_int_pattern)
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -182,7 +182,7 @@ Groups will be as follows:
   * ``value``: reported position value
 """
 
-regex_firmware_splitter = re.compile(r"\s*([A-Z_]+):\s*")
+regex_firmware_splitter = re.compile(r"\s*([A-Z][A-Z0-9_]*):\s*")
 """Regex to use for splitting M115 responses."""
 
 regex_resend_linenumber = re.compile(r"(N|N:)?(?P<n>%s)" % regex_int_pattern)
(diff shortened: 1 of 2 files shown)


foosel:
I see this is still in draft/wip... are you still looking into this?

cp2004:
Every now and then, I think about it again... I have tried several ways to not break keys with numbers in but at the same time do what the above code needs to do... Still haven't come up with anything concrete, beyond what was suggested in discord. If you've got an idea, by all means go for it since my regexp knowledge is still not all that great.

cp2004:
I think the only way I can figure out to do this would be something like this:
* The key **must** have a letter in it
* The key **can** have a number in it, but this is not required.

In the protocol of things here, would you find a key that was just numbers?

foosel:
I think we should be able to do something like

  * the key MUST only contain uppercase letters, underscore and numbers
  * the key MUST start with an uppercase letter

There's no spec anywhere as far as I know (surprise, surprise), but from what I've seen in the past we should be fine with that definition. So I think a regex along the lines of `\s*([A-Z][A-Z0-9_]*):\s*` should get us what we want.

If that however doesn't work out, we won't get around using a tokenizer instead of a regex.

foosel:
Pushed such a change... @cp2004 can you give it a look?

cp2004:
Well, the tests passed, so it's better than my attempt... Might make sense to add one that tests the case this PR will fix, to correctly identify the name from a response with the time in it. I can do that, then will mark this ready to review/merge.

foosel:
I think we should probably also add a test case for something like `1KEY` or `_KEY` getting ignored.

cp2004:
Something like this:
```
KEY1:VALUE1 _KEY2:INVALID 123:INVALID
```
is parsed to
```python
{'KEY1': 'VALUE1 _', 'KEY2': 'INVALID 123:INVALID'}
```

Is that intended? I can't tell

cp2004:
Tests that work are pushed, ready for merge 😄

cp2004:
What's the point in having pre-commit file watcher if it is disabled... *sighs* it's installing.

foosel:
That reminds me, on the weekend at PyCascades I learned about pre-commit.ci, I need to take a look at that.



> Is that intended? I can't tell

Hm... nope, it should be 

```
{"KEY1": "VALUE1 _KEY2:INVALID 123:INVALID"}
```

So that still needs some fine tuning. Hmmmm... Right now it doesn't require whitespace before the key's start. I think it should however, either that or string start. So replacing the `\s*` at the start with `(^|\s+)` should do it, but then the capture group index needs adjustment.

foosel:
I think now we got it.

PR #3923, opened 2021-01-03
Plugin version management infos added

tg44:
This is a small documentation addition. No new code, no new bugs, probably happier plugin developers.

### DIFF
--- a/docs/plugins/distributing.rst
+++ b/docs/plugins/distributing.rst
@@ -60,3 +60,22 @@ that in the `Plugin Repository's help pages <http://plugins.octoprint.org/help/r
 If you used the `OctoPrint Plugin Cookiecutter Template <https://github.com/OctoPrint/cookiecutter-octoprint-plugin>`_
 when creating your plugin, you can find a prepared registration entry ``.md`` file in the ``extras`` folder of your
 plugin.
+
+Version management after the official plugin repository release
+---------------------------------------------------------------
+
+Once your plugin is available in the official plugin repository, you probably want to create and distribute new versions. 
+For "beta" users you can use the manual file distribution method, or a more elegant release channels (see below). 
+After you finalized a new plugin version, don't forget to actually update the version in the ``setup.py``, 
+and `submit a new release on github <https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository#creating-a-release>`_. 
+
+After you published the new release, you can verify it on your installed octoprint, 
+with force checking the updates under the advanced options (in the software updates menu in the settings). 
+The new versions will appear to the plugin users in the next 24 hours (it depends on their cache refreshes).
+
+The `Software Update Plugin <https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#>`_ has options to define multiple release channels, 
+and you can let the users decide if they want to test your pre-releases or not.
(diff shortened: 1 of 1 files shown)


docs/plugins/distributing.rst
```diff
@@ -60,3 +60,8 @@ that in the `Plugin Repository's help pages <http://plugins.octoprint.org/help/r
...
+---------------------------------------------------------------
+
+Once your plugin is available in the official plugin repository, you probably want to create and distribute new versions. For "beta" users you can use the manual file distribution method, because this is the easiest way to try out a new plugin, or a new version before release. After you finalized a new plugin version, don't forget to actually update the version in the ``setup.py``, and submit a new release on github. Submitting a new release is rather easy, you should click on the releases, create/draft new release, on github. Give a tag to your release (this should be the same as the version number in the ``setup.py`` like ``1.2`` or ``v1.2``), add a name for the release (this could be whatever you want), and write a meaningful description with the changes. After you published the new release, you can check it out on your installed octoprint, with force check the updates under the advanced options (in the software updates menu in the settings). The new versions will appear to the plugin users in the next 24 hours (it depends on their cache refreshes).
```
cp2004, line 67: Couple of requests that I think would make this clearer:

* Split it across more than one line, to make it easier to maintain. The lines above should be an indicator of how long they should be, single line breaks are just for source management - when the docs are built it gets put into a paragraph and scales with screen size. Double line breaks are a change in paragraph.
* Mention the release channels option, for 'beta' users. IMO this is easier than the manual file distribution, since it uses the SW update plugins. [See this part of the SW update plugin docs](https://docs.octoprint.org/en/master/bundledplugins/softwareupdate.html#version-checks), or a [more detailed guide I wrote linked from the FAQ](https://github.com/cp2004/OctoPrint-Knowledge/blob/main/release-channels.md) on release channels.

Welcome the change, just want it to be easy to maintain!

cp2004 (changes requested):

foosel:
I agree with both points raised by @cp2004

tg44:
Me too, just had a lot on my plate :D

docs/plugins/distributing.rst
```diff
@@ -60,3 +60,25 @@ that in the `Plugin Repository's help pages <http://plugins.octoprint.org/help/r
...
 plugin.
+
+Version management after the official plugin repository release
+---------------------------------------------------------------
+
+Once your plugin is available in the official plugin repository, you probably want to create and distribute new versions. 
+For "beta" users you can use the manual file distribution method, or a more elegant release channels (see below). 
+After you finalized a new plugin version, don't forget to actually update the version in the ``setup.py``, and submit a new release on github. 
+
+Submitting a new release is `rather easy <https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository#creating-a-release>`_,
```
tg44, line 71: Probably, the whole "how to create release" part is unnecessary, and I should just make a link to the `submit a new release on github`, and keep the "After you published" part and so on?

foosel: Yeah, I'd avoid repeating GitHub's docs and just point people to them - they should also always be accurately reflecting what the user encounters, so less risk of stuff in our docs going stale if we just link to the source.

foosel:
Looks good now! I've switched it over to go right on the `master` branch so that it's going to be up right once the docs build has run through :)

PR #3924, opened 2021-01-03
Fix button state in usage tracking wizard

jasonbcox:
* Fix button state so that the tracking can be enabled/disabled after
selecting one of the options for the first time.
* Improve formatting in tracking wizard template to improve code
readability.
* Add myself to AUTHORS file.

### Test plan
```
rm -rf ~/.octoprint
octoprint serve
```
Navigate to http://0.0.0.0:5000/ and follow prompts until Anonymous Usage Tracking.
With this patch applied, you should be able to change your decision.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -130,6 +130,7 @@ date of first contribution):
   * [Sophist](https://github.com/Sophist-UK)
   * [Manuel McLure](https://github.com/ManuelMcLure)
   * ["j7126"](https://github.com/j7126)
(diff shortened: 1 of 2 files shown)


foosel:
Thanks! Should probably then also be done for blacklist processing & network connectivity check, for consistency. Do you want to send in a PR for those too?

jasonbcox:
> Thanks! Should probably then also be done for blacklist processing & network connectivity check, for consistency. Do you want to send in a PR for those too?

Yes, I will take a look since this PR was accepted without any major issue. :)

PR #3943, opened 2021-01-12
Fix button state in connectivity check subwizard

jasonbcox:
* Similar to #3924, fix button state so that the connectivity check can be enabled/disabled after
selecting one of the options for the first time.
* Improve formatting in the connectivity check wizard template to improve code readability.

### Test plan
```
rm -rf ~/.octoprint
octoprint serve
```
Navigate to http://0.0.0.0:5000/ and follow prompts until Online Connectivity Check.
With this patch applied, you should be able to change your decision.

### DIFF
--- a/src/octoprint/plugins/corewizard/templates/corewizard_onlinecheck_wizard.jinja2
+++ b/src/octoprint/plugins/corewizard/templates/corewizard_onlinecheck_wizard.jinja2
@@ -11,7 +11,7 @@
     OctoPrint comes preconfigured to perform the connectivity check every 15 minutes. You may change the value here.
 {% endtrans %}</p>
 
-<form class="form-horizontal" data-bind="with: settingsViewModel, enable: !setup(), css {disabled: setup()}" onsubmit="return false;">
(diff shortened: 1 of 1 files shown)


src/octoprint/plugins/corewizard/templates/corewizard_onlinecheck_wizard.jinja2
```diff
@@ -11,7 +11,7 @@
     OctoPrint comes preconfigured to perform the connectivity check every 15 minutes. You may change the value here.
 {% endtrans %}</p>
 
-<form class="form-horizontal" data-bind="with: settingsViewModel, enable: !setup(), css {disabled: setup()}" onsubmit="return false;">
```
jasonbcox, line 14: These enable/css properties on the forms do not appear to work. I removed them entirely because I don't see any reason why they should be disabled regardless of the choice to enable/disable the top-level feature.

foosel: I honestly can't tell you what this was supposed to accomplish either, so yeah, good riddance.

PR #3949, opened 2021-01-13
Reload logging settings when settings are viewed.

eyal0:
* [x] Your changes are not possible to do through a plugin and relevant
    to a large audience (ideally all users of OctoPrint)
  * [x] If your changes are large or otherwise disruptive: You have
    made sure your changes don't interfere with current development by
    talking it through with the maintainers, e.g. through a
    Brainstorming ticket
  * [x] Your PR targets OctoPrint's devel branch if it's a completely
    new feature, or maintenance if it's a bug fix or improvement of
    existing functionality for the current stable version (no PRs
    against master or anything else please)
  * [x] Your PR was opened from a custom branch on your repository
    (no PRs from your version of master, maintenance or devel please),
    e.g. dev/my_new_feature or fix/my_bugfix
  * [x] Your PR only contains relevant changes: no unrelated files,
    no dead code, ideally only one commit - rebase and squash your PR
    if necessary!
  * [x] Your changes follow the existing coding style
  * [x] If your changes include style sheets: You have modified the
    .less source files, not the .css files (those are generated with
    lessc)
  * [x] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [x] You have run the existing unit tests against your changes and
    nothing broke
  * [x] You have added yourself to the AUTHORS.md file :)

#### What does this PR do and why is it necessary?

This PR fetches the logging settings each time the setting dialog is opened.  This makes it so that if "Close" was previously used on the settings instead of "Save", the settings dialog will revert back to the unchanged settings the next time that the settings dialog is opened.

#### How was it tested? How can it be tested by the reviewer?

1. Open the Logging settings dialog.
2. Make some changes to the Log levels.
3. "Close" but don't "Save" the settings.
4. Re-open the Logging settings dialog and notice that the unsaved settings are reverted (previously they would not be).

#### What are the relevant tickets if any?

Fixes #3948

PR #3957, opened 2021-01-13
Automatically add logger when one is selected

### LINKED ISSUE #3947: [Request] Improve logging debug settings to be less confusing
**Is your feature request related to a problem? Please describe.**
The users of my plugin are often confused by the logging settings screen.

It sometimes appears as if debug logging has been set when it hasn't.  Here are the steps that they try:

1. Go to logging settings and set the logging level (they forget to press the plus sign):

image

2. Click Save.
3. Do something that tries logging.  Notice that no logging was done.
4. Go back to the settings and see that it's still there, so they assume that the logging is broken.  But no, it's actually the settings that are confusing the user.

**Describe the solution you'd like**

I suggest that the comboboxes for "Name" and "Level" always be blank.  And as soon as any of them are modified, a new row is created below immediately.  There is no danger that the settings will take place before clicking "Save" because the settings are only applied on "Save".  The user can still delete the row.

**Describe alternatives you've considered**

I thought about just making sure that the comboboxes for the new row are always blank when opening the settings dialog.  This would at least solve the problem where a user closes the settings dialog, opens it again, and still sees their settings.  However, it just takes more work form the user and they'll still be confused.

**Additional context**

Seen in https://github.com/OctoPrint/OctoPrint-Slic3r/issues/60

eyal0:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This fixes #3947 

I did a lot of adding and removing of log levels, sometimes closing without saving, sometimes saving.  I also sometimes would reload in the middle.

This fixes #3947 

Notice that it depends on merging #3949 and also merging #3951 . Merge those PRs before merging this one.

image

Notice that there is no blue plus symbol to the right of the final row (under the red garbage can).  Notice also that the name of the logger is blank.  When a logger is select there, for example, `octoprint`, here's what it will look like:

image

The row is added immediately a new blank one is available.

#### Further notes

As usual, changes only apply after pressing "Save".  Pressing "Close" causes no changes.

### DIFF
--- a/src/octoprint/plugins/logging/static/js/logging.js
+++ b/src/octoprint/plugins/logging/static/js/logging.js
@@ -108,6 +108,8 @@ $(function () {
 
             // loggers
             var availableLoggers = _.without(response.loggers, configuredLoggers);
+            availableLoggers.push("");
+            self.availableLoggersName("");
             self.availableLoggers(availableLoggers);
         };
 
@@ -180,6 +182,9 @@ $(function () {
             }
 
             var component = self.availableLoggersName();
+            if (component == "") {
+                return;
+            }
             var level = self.availableLoggersLevel();
 
             self.configuredLoggers.push({
--- a/src/octoprint/plugins/logging/static/js/logging.js
+++ b/src/octoprint/plugins/logging/static/js/logging.js
@@ -21,13 +21,6 @@ $(function () {
             return _.sortBy(self.availableLoggers());
         });
 
-        self.configuredLoggersSorted = ko.computed(function () {
-            return _.sortBy(self.configuredLoggers(), function (o) {
-                o.level();
-                return o.component;
-            });
-        });
-
         // initialize list helper
         self.listHelper = new ItemListHelper(
             "logFiles",
@@ -103,6 +96,9 @@ $(function () {
                 });
                 levels.push(item);
                 configuredLoggers.push(logger);
+            });
+            let sortedLevels = _.sortBy(levels, function (o) {
+                return o.component;
             });
             self.configuredLoggers(levels);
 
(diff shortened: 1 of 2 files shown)


src/octoprint/plugins/logging/static/js/logging.js
```diff
@@ -21,13 +21,6 @@ $(function () {
             return _.sortBy(self.availableLoggers());
         });
 
-        self.configuredLoggersSorted = ko.computed(function () {
-            return _.sortBy(self.configuredLoggers(), function (o) {
-                o.level();
```
eyal0, line 26: Was there any point to calling `o.level()` here?  I think that calling the knockout js element without using it has no effect.

As far as I can tell, it was added in 986591aed4c1771fe4992dae19357c83bb11e731 which was a squash, so the actual addition is:

https://github.com/OctoPrint/OctoPrint/pull/2091/commits/481aa6727a8e6f829b0c61f235f65892e15849cd#diff-ee9dc3299c3eba315a9e40fa02055530ea1aa207ac30a2a9d4cb2b6522652a0dR17

There doesn't seem to be any reason for it and the code works fine without.

src/octoprint/plugins/logging/static/js/logging.js
```diff
@@ -104,10 +97,15 @@ $(function () {
                 levels.push(item);
                 configuredLoggers.push(logger);
             });
+            let sortedLevels = _.sortBy(levels, function (o) {
+                return o.component;
+            });
```
eyal0, line 102: The sort is moved down to here and only happens when the settings dialog is first opened.

The reason is because it is inconvenient for the logging level to get sorted right after you add a new one.  For example, if you add the sixth logging level and then it sorts, it might jump to row, say, 3.  If you want to now change it from `WARNING` to `DEBUG`, you have to find it again in the list.

That's annoying and the user will not expect it.  So we'll only sort when the logging levels are retrieved from the server, which happens every time the setting dialog is opened.

src/octoprint/plugins/logging/static/js/logging.js
foosel: Probably something only @kantlivelong can answer 😅

kantlivelong: Not entirely sure either really. Likely just left it in by mistake. Good find :+1:

PR #4005, opened 2021-02-05
sockJS include when playing with jsclient lib

0r31:
Hi again @foosel ,

To complete the previous PR i sent regarding js client socket documentation and according to the [js_client bundle](https://github.com/OctoPrint/OctoPrint/blob/master/src/octoprint/server/__init__.py#L1961), i think it could be useful to add the sockJS lib include in the documentation, if needed.

Regards,
Orel


foosel (approved):

foosel:
Thank you :)

PR #4015, opened 2021-02-11
Center Viewport on gcode viewer is cased wrong

LazeMSS:
The case of the local storage parameter for gcode viewer was wrong when reading it compared to when saving it.
I decided to change the load instead of the save so it will work with people who has it stored prev. :)

On my own installation


foosel:
Typos like this will be the end of me 😅 Thank you for spotting & fixing!

PR #4017, opened 2021-02-11
Gcode viewer hide prime

LazeMSS:
Added an option for the  GCode Viewer so one can specify a string. And when provided the GCode Viewer will search for this string, and if found, skip all gcode up until this string. This can be used to skip prime nozzle gcode in the preview

On my own instance :)

Notice that the way the gcode is "sliced/cut" has been tried with many different ways for speed and mem consumption.
https://jsben.ch/9cwLJ - indexOf seems to be fastest on all browsers tested - firefox has includes 3% faster but then again chrome is the most used browser and is much faster using indexOf

The way the gcode "cut" is implemented is to find a line starting with the entered string and cut everything before that point and comment anything remaining out of that line. So if one is search for "; layer 1" it wil search for "\n; layer 1" - and if found in a string like this: "Blah\nBlah2\nBlah3\n; layer 1 Blah4 Blah\nBlah5\nBlah6" then the string sent to the viewer will be: "; Blah4 Blah\nBlah5\nBlah6". 

Converting the gcode into single lines and parsing line by line - this would cause slowness and memory consumption beyond what I prefer. 

The current solution is a compromise between something "easy" for the enduser and also fast and memory efficient.

### DIFF
--- a/docs/bundledplugins/gcodeviewer.rst
+++ b/docs/bundledplugins/gcodeviewer.rst
@@ -27,7 +27,7 @@ The plugin supports the following configuration keys:
   * ``mobileSizeThreshold``: Whether to also enable tracking on unreleased OctoPrint versions (anything not stable releases
     or release candidates). Defaults to ``false``.
   * ``sizeThreshold``: Unique instance identifier, auto generated on first activation
-  * ``skipUntilthis``: If this string is provided the GCode Viewer will search for this string, and if found, skip all gcode up until this string. This can be used to skip prime nozzle gcode in the preview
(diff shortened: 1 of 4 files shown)


shadycuz:
@LazeMSS  Could you provide a screenshot of the before and after gcode viewer in the PR description?

Would this functionality have also been doable as a plugin?

LazeMSS:
@shadycuz  yes it was original thought as a plugin but it seemed more nice to include it in the main.

Disabled
image

Enabled:
image

shadycuz:
ahh, so it brings the actual object into the center of the screen which gives a nicer field of view. Noice feature!

LazeMSS:
@shadycuz  yes I don't personally need to see the prime line in my gcode - that is not that important :)

PR #4018, opened 2021-02-12
Update bug issue form to new v2 format

foosel:

A backwards incompatible change to the issue form yaml configuration format makes deployment of a new version necessary. We now have two forms in place, one for now until the change happens, one for after the change happens. It should be a seemless migration with zero down time of the form.

Hope to get a review by some hubbers since currently there's no validator for the new format available yet.


foosel:
@deniseyu @Oneill38 Could you give this a quick look? 😅

deniseyu (approved):
I ran this through the new validation logic and it works fine! Thanks for making the changes ✨

foosel:
@deniseyu Perfect, thank you for the check!

PR #4029, opened 2021-02-24
Added missing Etag comparator method

encetamasb:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

When computed etag matches the request's header then tornado should respond with HTTP 304 instead of 200 (with full content). Tornado has builtin support for this but without a comparator method in the request handler its not functional.

Opened the web ui and checked response status codes. I have an 1.3MB packed_lib.js what gets downloaded almost every time I open the ui. After applying my changes I can see the expected status code (304) and cache works properly.

https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.check_etag_header

.

.

#### Further notes

It's possible that I miss something and its intentional that this method does not exist.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -138,6 +138,7 @@ date of first contribution):
   * [Nils Hofmann](https://github.com/Master92)
   * [Miroslav Šedivý](https://github.com/eumiro)
   * [Costas Basdekis](https://github.com/costas-basdekis)
+  * [Bence Tamas](https://github.com/encetamasb)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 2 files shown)


foosel:
Tornado actually has a default implementation of `check_etag_header`, and in the non-streamed `get` handler (that's used for webassets, among other things) that should actually already get used. And it does. However, it turns out that the ETag header set on the response was not enclosed in quotes as defined in RFC 7232, and THAT caused the already implemented ETag check to never succeed.

I've fixed THAT in ffea45ce5499a267a3b717df3c2030cd2d2ae7cb. This PR, while indeed solving the problem, does so by breaking the comparison algorithm to work around the non standard conform ETag header and thus covers the root of the issue. However, even though I'm not merging this, a big thank you for bringing this problem to my attention and also providing a fix 👍, even though it turned out to not be the best one for the problem on hand. Also, I learned something new today, so thanks for this as well 😊

PR #4031, opened 2021-02-27
Changes the default sorting of the "Install new Plugins..." modal

martellaj:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR changes the default sorting of the "Install new Plugins..." modal to "popular" instead of name. As a new OctoPi user, I was very interested in checking out available plugins to extend the functionality of the service. When I clicked "Get More...", I was surprised to see it simply sorted by name.

Sorting by name doesn't tell me much. I don't know how popular (i.e. how useful other users find each plugin) plugins are, if plugins are trustworthy or well-tested (maybe they have few users), and sorting this way also promotes very specific naming choices... (i.e. the first app listed doesn't even list its actual name as it's title, but chose an "A" word instead).

Admittedly, I didn't build the source code. I just looked at the JS file, determined which "sort" values there are and chose the corresponding one.

Context provided in the first question. 👍

None that I could find.

No UI change (just a different default setting).

#### Further notes
None.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -138,6 +138,7 @@ date of first contribution):
   * [Nils Hofmann](https://github.com/Master92)
   * [Miroslav Šedivý](https://github.com/eumiro)
   * [Costas Basdekis](https://github.com/costas-basdekis)
+  * [Joe Martella](https://github.com/martellaj)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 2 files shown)


kantlivelong:
Just my 2c. I think this PR would have a lot more value if it was implemented as a sort option instead of just replacing the hardcoded value. As the PR stands I can see other users asking for sort by name to be preferred.

martellaj:
It's already a sort option - and I believe your sort option is sticky once you change it.

The point of the PR is to change the default to avoid the "phone book" problem of people just choosing A-names for plugins so they rise to the top. 👍

jneilliii:
@martellaj is correct. This sort option is the equivalent of what's listed as `sort by active instance count` in the sort options of Get More.

kantlivelong:
Doh! I must be in the dark ages thinking it wasn't an option. 🥇 

There goes my pennies...

kantlivelong:
In that case yeah makes sense. Should help reduce "top plugins" questions as well.

foosel:
Yep, agreed, makes sense indeed!

foosel (approved):

PR #4048, opened 2021-03-13
Show meaningful message when software update is blocked

### LINKED ISSUE #4024: Wrong error message when software update is blocked by undervoltage
**The problem**

Seeing a few instances of the software update plugin failing to show the correct error message when undervoltage states are present, it shows 'Update not started. Is it already active? Please consult octoprint.log for details.'

Links:
https://discord.com/channels/704958479194128507/708230190462926898/811727359601541180
https://community.octoprint.org/t/plugin-updates-cant-be-installed/30811

There have been others, but these are two most recent I can remember.

The one in discord mentioned the undervoltage state was not showing up in the navbar either.

**Did the issue persist even in safe mode?**

**If you could not test in safe mode, please state why**

Yet to investigate/reproduce

**Version of OctoPrint**

1.5.x

**Operating system running OctoPrint**

Raspberry Pi / OctoPi 0.18, probably others

**Printer model & used firmware incl. version**

NA

**Browser and version of browser, operating system running browser**

NA

**Checklist of files to include below**

- [X] octoprint.log
- [ ] serial.log or contents of your terminal tab (always include in cases of issues with printer communication)

Posting as a bug report so that this is not forgotten about, so that the SW update plugin can be poked to make it do the right thing. Have not yet confirmed problem, but it has come up a couple of times on support channels.

cp2004:
...and log when it is undervoltage.

The problem here was twofold: first, a very generic message asking to check the log confused users. Second, there was nothing logged when it said check the logs.

So, this changes that so the message is escaped and displayed in the pnotify and when the update is blocked for undervoltage this is logged.

This does not add logging for other errors, since it seems unnecessary to me.

Closes #4024

### DIFF
--- a/src/octoprint/plugins/softwareupdate/__init__.py
+++ b/src/octoprint/plugins/softwareupdate/__init__.py
@@ -1086,10 +1086,14 @@ class SoftwareUpdatePlugin(
             and not self._settings.get_boolean(["ignore_throttled"])
         ):
             # currently throttled, we refuse to run
+            message = (
+                "System is currently throttled, refusing to update "
+                "anything due to possible stability issues"
+            )
+            self._logger.error(message)
             flask.abort(
                 409,
-                description="System is currently throttled, refusing to update "
-                "anything due to possible stability issues",
+                description=message,
             )
 
         if self._printer.is_printing() or self._printer.is_paused():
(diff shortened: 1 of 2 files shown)


foosel:
:+1:

PR #4049, opened 2021-03-13
Consistency in the settings headings

cp2004:
Each heading used it's own style or none at all, so now they are all aligned with the most common.

No issues here, this was just one thing that annoyed me :)


foosel:
Hm.... semantically, a fieldset legend is actually the right thing here. So I'm a bit unsure about this change. I'm wondering if it wouldn't be better actually to introduce fieldsets or rather legends where there aren't any yet.

cp2004:
That makes sense, I just wanted it to be consistent and chose the most common.

If you want to go with fieldsets/legends, I can do that or close this PR and go for it separately.

foosel:
Seriously undecided on this I gotta say. Tried to find some definitive guidance on the question "headline vs legends" and couldn't find anything. So I'm now actually rather leaning towards `h3` after all since it's simply less invasive... going to merge this.

Issue #3801, opened 2020-11-09
Plugin list and Software update list extends beyond the configuration window


dewi-ny-je:
I'm setting up Octoprint from the configuration window

Windows 10 resolution: 1600x900, 125% zoom 

1. Go to Plugin Manager or Software Update
2. notice the double scrollbar

#### What did you expect to happen?
There should be only a scrollbar: either the "inner" list is resized to the visible area of the configuration dialog, or the inner list (the plugins) is expanded completely

#### What happened instead?

the inner list has a fixed size which is in my case bigger than the Octopring modal configuration window, resulting in more difficult scrolling, especially in laptops (which are the computer most likely to show the problem due to the smaller resolution.

#### Did the same happen when running OctoPrint in safe mode?

yes

#### Version of OctoPrint
1.4.2

#### Operating System running OctoPrint
Windows 10 20H2

#### Printer model & used firmware incl. version
Not relevant.

#### Browser and version of browser, operating system running browser
Chrome 86

#### Link to octoprint.log
![settings](https://user-images.githubusercontent.com/2866139/98599347-1d05c900-22dc-11eb-90df-d643594ed5b0.PNG)
[octoprint.log](https://github.com/OctoPrint/OctoPrint/files/5513423/octoprint.log)

#### Link to contents of terminal tab or serial.log
not relevant

#### Link to contents of Javascript console in the browser
https://pastebin.com/uNCPz3fW

#### Screenshot(s)/video(s) showing the problem:

I have read the FAQ.

cp2004:
PRs #3835 and #3836 have been created to address this. Covering both plugin manager, and SW update tabs.

Final one is the 'Get More' tab of the plugin manager, that double scrolls too:
![image](https://user-images.githubusercontent.com/31997505/100445992-ec78aa00-30a5-11eb-8bd1-ad32871b9418.png)

My initial idea for this was to make it a larger modal (same width as settings), then divide in two, with repository on the left, and URL and file upload on the right. This idea was also with (partial) implementation of #3763 in mind, to create a full export and import feature following #3711 implementation, where the import from the json file was an extra field.

foosel:
1.6.0 has just been released.

PR #4071, opened 2021-04-01
💄  Better fix for styling of the pluginmanager

cp2004:
Better fix than my hack styling that navbar before.

Co-Authored-By: LazeMSS <69049371+LazeMSS@users.noreply.github.com>

Adjusts the styling in the pluginmanager to avoid the issue of hardcoded styles in the navbar, which break themes used. This fix gets rid of that sticky styling (which actually [isn't widely used when I looked at it](https://caniuse.com/?search=sticky) and still achieves the same thing. Styling is not my strong point 😄.

Made the changes then opened the navbar, confirmed that there were no dual scrollbars and the navbar remained, so the only scrollbar is the table below now.

#3835 and the initial suggestion #3801.

See above

image

#### Further notes

I would like this to make it to 1.6.0rc, since it is an issue brought in since 1.5.x


foosel:
Thanks for the reminder, this one completely managed to pass me by 🤨

LazeMSS:
Looks great @cp2004

PR #4079, opened 2021-04-04
Endpoint improvement: api/job#status to return PrinterState object instead of state.text

DarkGuardsman:

Changes the response of `api/job` to return the [PrinterState](https://docs.octoprint.org/en/master/api/datamodel.html#printer-state) object instead of state.text value. 

This allows for a cleaner response in terms of telling the state of the printer. As well provides more information from the endpoint on printer state.

Calling the endpoint to validate the response

Working on a java based wrapper for the API to be used with other apps. Ran into an issue with the status field on the job response being inconsistent in it's behavior. Specifically according to the documentation I was expecting a finite list of status string code. Instead it seems the status can be any value supplied to status.text. This includes errors and other values from what I can see in the python code. 

Issue with this is getting supplied something like `Offline (Error: No more candidates to test, and no working port/baudrate combination detected.)`. Which requires parsing/regex to pull out the state of `Offline` and Error of `No more candidates to test, and no working port/baudrate combination detected`. This is also a bit confusing as its unclear if the state is `Offline` or `Error`. I can apply a best guess in the logic. However, this leads to another problem of "what if the text changes?". As this seems to be a field used in the UI to display the status to the user. In theory this means a plugin or other source could change it breaking logic.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -141,6 +141,7 @@ date of first contribution):
   * [Joe Martella](https://github.com/martellaj)
   * [Scott Lahteine](https://github.com/thinkyhead)
   * [Karthikeyan Singaravelan](https://github.com/tirkarthi)
+  * [DarkGuardsman](https://github.com/darkguardsman)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/server/api/job.py
+++ b/src/octoprint/server/api/job.py
@@ -75,9 +75,6 @@ def jobState():
     response = {
         "job": currentData["job"],
         "progress": currentData["progress"],
-        "state": currentData["state"]["text"],
+        "state": currentData["state"],
     }
-    if currentData["state"]["error"]:
-        response["error"] = currentData["state"]["error"]
-
     return jsonify(**response)


DarkGuardsman:
Setting this as draft to get feedback before converting it to a merge. Especially given this is the first time pulling to this repository. Let me know if I need to do anything more or complete other steps.

cp2004:
Hey @DarkGuardsman, thank you for the contribution! However, there are some issues/comments with it:

I don't understand why this is a necessary change, and your explanation doesn't clear this up for me. There's a [Printer state](https://docs.octoprint.org/en/master/api/printer.html#retrieve-the-current-printer-state) endpoint that returns all this data already. The text string *should not* be used for parsing to work out the state, instead use the state flags that are returned.

> Specifically according to the documentation I was expecting a finite list of status string code.

The wording here is adjusted in the docs, 'i.e' does not mean finite list, but it now has additional clarification for the 1.6.0 docs (linked below)

The 'error' situation you described is also already solved, for 1.6.0 which is in release candidate stage. It splits the state and error, so that you get two separate strings. The state string becomes `Offline after error` and there is an `error` field. This can be found in the 1.6.0rc1 docs: https://docs.octoprint.org/en/1.6.0rc1/api/job.html#job-information-response

**Second**, this is a breaking change to the API, so it makes it more difficult for consumers of the API to use it:

Before:
```json
{
  "job": {
    "estimatedPrintTime": null, 
    "filament": {
      "length": null, 
      "volume": null
    }, 
    "file": {
      "date": null, 
      "name": null, 
      "origin": null, 
      "path": null, 
      "size": null
    }, 
    "lastPrintTime": null, 
    "user": null
  }, 
  "progress": {
    "completion": null, 
    "filepos": null, 
    "printTime": null, 
    "printTimeLeft": null, 
    "printTimeOrigin": null
  }, 
  "state": "Operational"
}
```
After:
```json
{
  "job": {
    "estimatedPrintTime": null, 
    "filament": {
      "length": null, 
      "volume": null
    }, 
    "file": {
      "date": null, 
      "name": null, 
      "origin": null, 
      "path": null, 
      "size": null
    }, 
    "lastPrintTime": null, 
    "user": null
  }, 
  "progress": {
    "completion": null, 
    "filepos": null, 
    "printTime": null, 
    "printTimeLeft": null, 
    "printTimeOrigin": null
  }, 
  "state": {
    "error": "", 
    "flags": {
      "cancelling": false, 
      "closedOrError": false, 
      "error": false, 
      "finishing": false, 
      "operational": true, 
      "paused": false, 
      "pausing": false, 
      "printing": false, 
      "ready": true, 
      "resuming": false, 
      "sdReady": true
    }, 
    "text": "Operational"
  }
}
```

**Third**, any changes to the API must be documented - especially a **breaking change**. 

I don't have final say on PRs here, but my vote would be no, for two reasons: This data is available elsewhere, and it is a breaking change.

DarkGuardsman:
> The 'error' situation you described is also already solved

Nice, that was one of the main issues of why I created this pull request. Apologies for not reviewing latest iteration as this might have saved me a bit of time. 

> I don't understand why this is a necessary change

The endpoint already is returning a subset of data provided by the state.... might as well return the full state object. If we want to argue another endpoint already exists then why provide a state text field? 

> Second, this is a breaking change to the API

Yep it would be a breaking change. Which I don't see as much of an issue since a few breaking changes have happened in the API already. One that comes to mind is switching from integers to floats. This caused several issues in my own consumption of the API.

> Third, any changes to the API must be documented

I'm happy to document the changes and even version the API if required.... sadly not exactly sure where the documentation is located in the repo.

foosel:
> If we want to argue another endpoint already exists then why provide a state text field?

History of that endpoint and how it got used, and a need for clients to show a human readable state. I'd love to just get rid of the `state` there altogether these days, but that would be a breaking change that would potentially cause issues for a lot of third party clients.

> Which I don't see as much of an issue since a few breaking changes have happened in the API already. One that comes to mind is switching from integers to floats.

If there was a breaking change on the API, it was unintentional and I'd like more details on examples of this so it doesn't happen again in the future. In principle I *always* aim to keep the REST API and the documented internal APIs fully backwards compatible. The few rare cases where I actually did intentionally introduce breaking changes it was due to necessity to work around certain bugs or even security issues, got a big heads-up in the release notes and if possible got preceded by a long deprecation phase. 

This change here is certainly quite disruptive for any clients that use the `job` endpoint and expect a textual representation there instead of an object. As already mentioned, with 1.6 there'll be a separation of the text and error representation, as needed, so the main pain point should be resolved here. *If* this endpoint were to be touched in a breaking change, it would be to get rid of the state altogether instead of basically copying the state endpoint -- not without API versioning however. Thus I won't merge this, sorry. 

Would still love details on breaking changes you observed in the past however. Sounds like there might be some documentation issues or bugs somewhere that need to be taken care of.

DarkGuardsman:
Fair enough, thank you for at least looking at the PR :)

> Would still love details on breaking changes you observed in the past however

Sadly I have not kept a list, the one I noticed recently was with time fields moving from integer to float. I assumed it was unintentional and likely a side effect of types in python. Only frustrating part was documentation not matching but guess this is expected for projects 

https://docs.octoprint.org/en/1.5.3/api/datamodel.html#job-information `printTime` and `printTimeLeft`

https://docs.octoprint.org/en/1.5.3/api/datamodel.html#job-information `estimatedPrintTime`, and `lastPrintTime`

I'll take a bit more of a look later to see if I can figure out when it changed. As well, since I plan to support a library for calling the API I'll also see about making automated tests for your endpoints. This will be the best option to catching changes and mitigating impact down stream.

jneilliii:
> since I plan to support a library for calling the API

Interesting, are you aware of the one that already exists? Maybe you're doing it in a different language?

https://github.com/dougbrion/OctoRest

https://github.com/OutsourcedGuru/octo-client

DarkGuardsman:
> Maybe you're doing it in a different language?

Yep, implementing mine in java for use with Spring boot and Discord4j. This is likely why I've run into some of these issues as java is more strict on data types then both python and javascript.

PR #4087, opened 2021-04-09
Improved error handling and missing check for disable webcam

LazeMSS:
- Fixes broken handling of webcam when webcam is disabled
- Improves validation of native hls playback options

- By myself


jneilliii:
PR should be made against just `maintenance` branch.

PR #4089, opened 2021-04-10
remove duplicated word

jugmac00:


cp2004:
Could you switch the target branch for this PR to `maintenance`? Since it is just a documentation, it does not need to wait for OctoPrint 2.0 on the `devel` branch, and can be part of the minor releases of the 1.x series. Thanks 🙂.

PR #4103, opened 2021-04-19
💄 Improve all dialog animations

### LINKED ISSUE #4085: [Request] Look into alternative settings dialogs animations

The default Bootstrap 2 implementation of animations are slow. I'm informed they are using JavaScript and jQuery to animate them, which is not working that well.

Most of us have noticed a bit of a slowdown when trying to open the settings dialogs, coming and going. Potentially made worse by the ever-increasing amount of markup in the settings dialog, it is kinda annoying. So I went looking for a solution...

Looking into alternative animations for the settings dialog:

* Disable animations:
  * + : Quick response to clicking the button, and the dialog *appears*
  * - : There is still a slight delay, and the feeling of the settings dialog *just appearing* is strange. Too much of a flash of content at once.

* Use an alternative CSS animation. I like this one, because it is smooth, and customizable. Some animations I tried are below, including a grow animation, and replacing the slide down from BS2 with a CSS animation, and a fade animation.

The CSS animations all start straight away when the element is created, so there is no need for extra logic. The only thing that I haven't figured out is closing the settings, there is no animation on any of them. This would probably be easy enough to implement, but for proof of concept I didn't. 

See everywhere above and below...

Slide:
settings-slidedown

Grow:
settings-grow

Fade:
settings-fade

No animation:
settings-noanimation

Default animation:
settings-default

cp2004:

Implements #4085, replaces the default slide-in-fade animation with a simple, more performant CSS animation.

Applying the patch, opening all the dialogs to view the animation. Make sure none of them are using the slide animation....

For ages, I was annoyed by the speed of the animation, looked into it to try and find the slow thing, which was jQuery & BS2's JavaScript animations.

Closes #4085 

See #4085 for details

### TODO

* [x] Apply to all core dialogs. Probably only if the point below is addressed.
* [ ] ~~Figure out the closing of the dialogs~~
   * This is more difficult to do with CSS animations, ~~but I'm working on it. If it is too difficult, we'll need to work out if it is worth it or whether instant-close is acceptable.~~


src/octoprint/static/js/app/viewmodels/settings.js
```diff
@@ -801,7 +801,6 @@ $(function () {
                     }
                 })
                 .css({
-                    "width": "auto",
```
cp2004, line 804: No clue what this was for, and it causes issues with the sizing of the dialog. Git shows it going way back to when the dialog was implemented, I noticed no adverse effects from removing it.

cp2004:
@foosel what do you think about changing all the dialogs?

At the moment I am struggling to get a fade-out animation to work without diving deep into changing the Bootstrap 2 sources. So at the moment, there is a fade in and instant out. Can't hook into the close settings function to trigger an animation in the settings view model since it only covers clicking the settings button, but not the close X, outside the dialog or <kbd>esc</kbd> key either.

I definitely think this should be done for settings, since it is the heaviest dialog and is way too slow. But the rest are not as slow because they have relatively less content. Maybe I will experiment so we can see what it feels like.

foosel (approved):

src/octoprint/static/js/app/viewmodels/settings.js
foosel: Let's just hope for the best then, I have no clue anymore either.

foosel:
Frankly, I have no qualms with applying it to all core modals. It's faster, works, and the fading backdrop ain't that important really.

That being said, wasn't there some event triggered on the closing of a modal regardless of the way it is closed? Could this maybe help?

jneilliii:
> That being said, wasn't there some event triggered on the closing of a modal regardless of the way it is closed? Could this maybe help?

I think I tried using this before in a plugin and don't think it exposed correctly the callback.

cp2004:
There is something like this:
```js
$('#modal').on("hidden", () => {
   // Do something
})
```
But I think when I tried it, it fired ,*after* Bootstrap closed the modal. By which point, it is too late to start an animation. At least I think so. Will keep looking, and try and change most of the core dialogs.

cp2004:
Having changes all the core dialogs to fade in/no animation out, it works. It feels just fine. And quicker & snappier still.

Ready for merge/review 🙂 

Oh, and I reduced the animation duration slightly. It is now 300ms, down from 500. Maybe there is an optimum value, it's easily changeable.

cp2004:
Only outstanding issue seems to be the webcam & timelapse loading - because the content gets loaded later, then the positioning is off. This is actually an issue with the current implementation if things go fast enough - the timing means I can get  it to appear below.

I'll add some small CSS tweaks to hopefully predict the sizing of the webcam image and make it the correct size.

image

foosel:
So do you consider this ready for merge or should I wait some more?

cp2004:
It can be merged, it was luck that I found the final issue, I thought I was finished...

Issue #3505, opened 2020-04-02
[Request] Add unselect to /api/files in REST API


OutsourcedGuru:
**Is your feature request related to a problem? Please describe.**
It would be great if programs could unselect a file so that file queuing mechanisms, for example, might be written. There are times when you'd like to simply unselect the currently-selected file if you made a mistake, for example.

**Describe the solution you'd like**
In the POST to `/api/files/local/filename` endpoint of the API, the `command` could be `unselect` rather than `select`. I couldn't imagine other ways of indicating this (that might be more appropriate) other than possibly using a keyword like `/api/files/local/none`, perhaps with the standard `select` as the `command`...?

It feels odd having to indicate the filename so that it could be unselected. In some ways, I think I prefer the `/none` version of this.

**Describe alternatives you've considered**
I'm currently using the internal interface of printer.unselect_file(). But this requires you to be running within the space of a plugin.

**Additional context**
N/A

NotExpectedYet:
Would totally love this myself, it's been a requested feature for OctoFarm and I'm currently trying to work away around it with the current API implementation.

cp2004:
This was implemented in 1.7.0 which has been released 🎉

PR #4111, opened 2021-04-25
/api/files: add unselect file command

eltonlaw:

https://github.com/OctoPrint/OctoPrint/issues/3505

Introduces an endpoint to unselect the selected file. The attached ticket mentions it being necessary "so that file queuing mechanisms, for example, might be written" and "to simply unselect the currently-selected file if you made a mistake" 

Started the server and then ran the following. Seemed to work as expected
```
>>> import requests
>>> api_token = ...
>>> my_gcode_file = ...
>>> headers = {"Authorization": "Bearer " + api_token}
>>> base_url = "http://localhost:5000/api/files"

# happy path, select and unselect
>>> print(requests.post(f"{base_url}/local/{my_gcode_file}", json={"command": "select"}, headers=headers).content)
b''
>>> print(requests.post(f"{base_url}/local/{my_gcode_file}", json={"command": "unselect"}, headers=headers).content)
b''

# unselect with no file selected
>>> print(requests.post(f"{base_url}/local/{my_gcode_file}", json={"command": "unselect"}, headers=headers).content)
b'Cannot unselect current file when there is no file selected'

# have selected file and try to unselect invalid filename
>>> print(requests.post(f"{base_url}/local/{my_gcode_file}", json={"command": "select"}, headers=headers).content)
b''
>>> print(requests.post(f"{base_url}/local/foo", json={"command": "unselect"}, headers=headers).content)
b'Only the currently selected file can be unselected'
```

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -142,6 +142,7 @@ date of first contribution):
   * [Scott Lahteine](https://github.com/thinkyhead)
   * [Karthikeyan Singaravelan](https://github.com/tirkarthi)
   * [Yvan Rodrigues](https://github.com/TwoRedCells)
+  * [Elton Law](https://github.com/eltonlaw)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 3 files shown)


foosel (approved):
I could have sworn I HAD an `unselect` command already 😅 Thank you for fixing this oversight! 👍

PR #4128, opened 2021-05-06
Improve/ignored commands

StevilKnevil:
Added **Ignored Commands** to the Serial Protocol settings
This is an exact duplicate of the _Blocked Commands_ setting except that emitted event is severity `info` not `warn`. The result of this is that there isn't a red popup displayed when the GCode is encountered, the information is just logged to the logfile.

This is important if you are intentionally producing GCode that contains codes that your printer cannot handle (e.g `M600` for filament change on a printer with a single extruder - see below for further context)

Change the blocked commands to be `M0` and the ignored command to be `M1`. Sending `M0` and `M1` in the terminal should have exactly the same effect, except that there should be no popup for the `M1` command

To generate multicolour prints, I can specfiy that my printer has 2 extruders so that the sliced GCode will contain command like
```
...
M600 ; Filament Change
T1 ; Tool change
...
```
The physical printer only has a single extruder, so I can set `M600` as a pausing command (with a GCODE script to move the head safe) and manually perform a filament change.
However the M600 and T1 commands in the sliced GCode will cause scary red popups to appear in the octoprint window, despite them being entirely expected and allowed for. Setting M600 and T1 as _ignored commands_ (rather than _blocked commands_) will avoid the popups being displayed

None

image

#### Further notes
Note that by default no GCode commands are set as _Ignored Commands_

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -144,6 +144,7 @@ date of first contribution):
   * [Yvan Rodrigues](https://github.com/TwoRedCells)
   * [Elton Law](https://github.com/eltonlaw)
   * ["sparxooo"](https://github.com/sparxooo)
+  * ["Stevil Knevil"](https://github.com/StevilKnevil)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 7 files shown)


StevilKnevil:
Accidentally closed this PR!

foosel:
I'm slightly worried that will once again lead to people complaining that "OctoPrint should not refuse to send commands to my printer without me knowing about it" (which is why the notifications about blocked commands was introduced in the first place), but considering that we ship it empty and a log entry and event are still generated (albeit the latter is usually ignored) this feels low risk enough.

Thanks!

PR #4132, opened 2021-05-09
⚡ Don't re-render the plugin repository if notmodified

### LINKED ISSUE #4131: [Request] Potential performance increase in plugin repository rendering

Every time the settings dialog is opened, the plugin repository is requested from the server. The frontend code then spends roughly 500ms *just* rendering the repository:
image

This results in a period where you can't scroll the settings dialog because the plugin repository is rendering. More noticeable when the dialog opens quickly as well. This time *can* be more than 500ms, I recently rebooted and everything is running *fast* because it is not weighed down with a few tabs and several PyCharm instances...

I'm wondering, could there be some sort of etag/cache control that means that if we request the repository from the server and it hasn't modified (which is unlikley) then return 304 and don't render anything. Could it be as simple as a last-modified-date-etag thing? I will admit, I don't know much about it but it seems possible to me.

Releasing 1.7.0 with only a couple frontend performance increases, no more.

Only downloading plugin repository & render on page load, but this makes the initial load longer. I vote keep it with settings open, but only once or if modified.

### Additional context

_No response_

cp2004:

Don't re-render the plugin repository if the server responds with `notmodified`. This shaves ~500ms off the total processing time, making opening the settings dialog *lightning fast* ⚡.

Using the performance profiler, first I identified the issue, then opened and closes the settings dialogs *several* times, to make sure this change was effective.

General push to improve UI performance, started with the animations, onto the data processing...

Closes #4131 

image
Look, barely a dropped frame...

Comparison to *before* this change:
117578894-824e4180-b0e8-11eb-81c0-5380ecaad93a

#### Further notes

Ok, now another question. I noticed some of the other endpoints that are processed repeatedly like this, and wondered if it would make sense to implement the same thing?

It *has* to be on a per-request basis, since adding `ifModified: true` to all request options breaks handlers that are not expecting an empty response.

### DIFF
--- a/src/octoprint/plugins/pluginmanager/static/js/pluginmanager.js
+++ b/src/octoprint/plugins/pluginmanager/static/js/pluginmanager.js
@@ -693,11 +693,15 @@ $(function () {
             }
 
             OctoPrint.plugins.pluginmanager
-                .getRepository(!!options.refresh)
+                .getRepository(!!options.refresh, {ifModified: true})
                 .fail(function () {
                     deferred.reject();
                 })
-                .done(function (data) {
+                .done(function (data, status, xhr) {
+                    // Don't update if cached - requires ifModified: true to pass through
+                    // the 304 status, otherwise it fakes it and produces 200 all the time.
+                    if (xhr.status === 304) return;
+
                     self.fromRepositoryResponse(data.repository);
                     self.online(data.online !== undefined ? data.online : true);
                     deferred.resolveWith(data);


cp2004:
> Further notes
>Ok, now another question. I noticed some of the other endpoints that are processed repeatedly like this, and wondered if it would make sense to implement the same thing?

>It has to be on a per-request basis, since adding ifModified: true to all request options breaks handlers that are not expecting an empty response.

For reference, the additional processing done here adds up to less than 100ms for me.

cp2004:
> First of all, definitely makes sense. Would it be a good idea to do that on the client lib layer instead of in the viewmodels?

The only thing about that is that it would *technically* be a breaking change - since `data` is undefined - a 304 has no content, by design. If you are expecting some data, you now don't get it. Since the change is not network/request related, but actually viewmodel rendering it made sense to me to put it here.

Off the top of my head, it was printer profiles, access control, backup and logs that refreshed themselves like this.

PR #4148, opened 2021-05-22
Added browser user agent string to tracking data

apbarratt:
  * [n/a] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
    if necessary!
  * [n/a] If your changes include style sheets: You have modified the
    lessc)
  * [not really as I don't have access to tracking.octoprint.org to my knowledge] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [sorry, no, kind of did this one a bit blind I'm afraid and hoped interested folk wouldn't mind checking it over] You have run the existing unit tests against your changes and
    nothing broke
  * [oh... er... I hadn't really figured something so small would earn me such a credit] You have added yourself to the AUTHORS.md file :)

This PR updates the pong tracking data to include the user agent of the most recent browser to have loaded the web interface.

As a plugin developer, I found myself wondering from time to time what browser compatibility coverage I should factor in.  My assumption is everyone is using modern browsers but with such a community of hackers, I wasn't sure I should make this assumption, thinking there may be some rather niche setups out there.

I tried to find this information on https://tracking.octoprint.org and https://data.octoprint.org but found that it was not available.  This is presumably because thus far only server instance information has been tracked rather than any information about the devices being used to access said instances.

Whilst this information I suppose could technically have been collected in a plugin of some form, I just couldn't see anyone really being motivated to download it with the few who did download it likely being curious developers themselves and so skewing the dataset.

I'm afraid I don't have the knowledge or confidence to test properly and am just hoping someone with the skill and expertise needed will be interested enough in this feature to test as appropriate... sorry :(

I couldn't see any but I may have missed something.

#### Further notes

I've checked the wording of the tracking consent sections and updated where appropriate as well as ensured there is a new opt out checkbox for the user agent.

I'm also confident this won't fall foul of the EU cookie law or GDPR in any way as it doesn't collect any kind of identifiable information, nor does it use cookies in anyway.  Indeed, the only thing it collects is the user agent string, which will in almost all cases be identical to thousands of other user agent strings used daily, so unless someone has really gone out of their way to make their user own user agent utterly unique, I can't see how such data would apply to any data protection laws.

### DIFF
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -133,6 +133,7 @@ class Events(object):
 
     # Settings
     SETTINGS_UPDATED = "SettingsUpdated"
+    SETTINGS_LOADED = "SettingsLoaded"
 
     @classmethod
     def register_event(cls, event, prefix=None):
--- a/src/octoprint/plugins/tracking/__init__.py
+++ b/src/octoprint/plugins/tracking/__init__.py
@@ -62,6 +62,8 @@ class TrackingPlugin(
         return {
             "enabled": None,
             "unique_id": None,
+            "user_agent_collection": True,
+            "most_recent_user_agent": None,
             "server": TRACKING_URL,
             "ping": 15 * 60,
             "pong": 24 * 60 * 60,
@@ -119,7 +121,11 @@ class TrackingPlugin(
         if not self._settings.get_boolean(["enabled"]):
             return
 
-        if event in (
+        if event == Events.SETTINGS_LOADED:
+            self._handle_settings_event(event, payload)
+        
+
+        elif event in (
             Events.PRINT_STARTED,
             Events.PRINT_DONE,
             Events.PRINT_FAILED,
@@ -519,6 +525,7 @@ class TrackingPlugin(
             "cores": self._environment["hardware"]["cores"],
             "freq": self._environment["hardware"]["freq"],
             "ram": self._environment["hardware"]["ram"],
+            "most_recent_user_agent": self._settings.get(["most_recent_user_agent"])
         }
 
         if (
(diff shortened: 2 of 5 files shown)


foosel:
The last seen user agent should definitely be kept in memory and not be stored in the settings and thus on disk. I'm also not too keen on introducing a whole event just to capture the user agent. 

Wondering if it would make sense to rather add a tracking endpoint to the plugin and use it from the frontend on load 🤔 

Given that a *full* user agent can also be used to do some fingerprinting of clients that I really don't even want to begin to make possible here, let's also think about what exactly is needed here. I'd say the browser type (firefox, chrome, edge, ...) and OS (windows, linux, mac, ...) would be of interest. Everything else I think won't be. So best restrict the tracked data to this too.

In any case, if this gets merged (which I'm not opposed to, after some changes) this also needs to be documented on OctoPrint/tracking.octoprint.org. 

Let me have a quick look whether I can just throw up a quick tracking endpoint for the frontend to utilize, then we could just use the user agent to extract browser & os and only send THAT off to be tracked.

foosel:
Also, you apparently forgot to install and run `pre-commit` since you have some failing checks here.

foosel:
Actually, while I was testing the aforementioned tacking endpoint, I also quickly implemented it 😬

```
2021-05-26 18:18:31,112 - octoprint.plugins.tracking - INFO - Sent tracking event webui_load, payload: {'browser_name': 'Firefox', 'browser_version': '88.0', 'os_name': 'Windows', 'os_version': '10'}
```

Doesn't yet get persisted on the backend however, that'll need another change on the logstash config.

foosel:
This is also now listed on tracking.octoprint.org and the intake has also been updated through https://github.com/OctoPrint/tracking.octoprint.org/commit/99e0daa7f7c3b2923076b91ca5bd88907e464900.

Closing this as it hasn't been merged due to a different implementation approach, BUT I'll of course list you as a contributor on the release notes given that it was totally your idea and impulse! Hope that approach is alright for you, it was the easier option from the maintenance side of perspective this time.

apbarratt:
Oh excellent, I'm really pleased we've got a solution even though my own code was a bit ridiculous in implementation =D
Excited to see it popping up in future data now :) Wow, excited to see browser user agent tracking data events... what have I become‽ 😅

Issue #4157, opened 2021-06-04
GcodeInterpreter does not calculate arc moves


ldursw:
### The problem

When fixing an issue with arc moves not being correctly calculated in PR OllisGit/OctoPrint-SpoolManager#182 it was pointed out in a comment that the analysis code was copied from OctoPrint and the original file also has the same issue.

https://github.com/OctoPrint/OctoPrint/blob/8db57f72ab90c7ed9ef1fa5a92d11f8ff904d40d/src/octoprint/util/gcodeInterpreter.py#L420

The plugin only cares about the extrusion length so adding `"G2", "G3"` is enough. OctoPrint also calculates the XYZ moves so it would require some special handling for `IJ` parameters.

```
# Original file without arcs
$ OctoPrint/bin/octoprint analysis gcode Quest2ControllerCoverLeft.gcode
DONE:19.82067164313048s
RESULTS:
dimensions:
  depth: 180.0
  height: 32.4
  width: 156.883
extrusion_length:
- 6933.180119999942
extrusion_volume:
- 0.0
printing_area:
  maxX: 158.983
  maxY: 200.0
  maxZ: 32.4
  minX: 2.1
  minY: 20.0
  minZ: 0.0
total_time: 112.9555746417688

# File with arcs without G2/G3 added to line 420
$ OctoPrint/bin/octoprint analysis gcode Quest2ControllerCoverLeft.aw.gcode
DONE:2.820354064926505s
RESULTS:
dimensions:
  depth: 180.0
  height: 32.4
  width: 156.372
extrusion_length:
- 374.4529999999963
extrusion_volume:
- 0.0
printing_area:
  maxX: 158.472
  maxY: 200.0
  maxZ: 32.4
  minX: 2.1
  minY: 20.0
  minZ: 0.0
total_time: 41.78244642603364

# File with arcs with G2/G3 added to line 420
$ OctoPrint/bin/octoprint analysis gcode Quest2ControllerCoverLeft.aw.gcode
DONE:3.956256238743663s
RESULTS:
dimensions:
  depth: 180.0
  height: 32.4
  width: 156.857
extrusion_length:
- 6933.180120000003
extrusion_volume:
- 0.0
printing_area:
  maxX: 158.957
  maxY: 200.0
  maxZ: 32.4
  minX: 2.1
  minY: 20.0
  minZ: 0.0
total_time: 112.24193040348119
```

[...]

PR #4159, opened 2021-06-06
Add arc calculation to gcode interpreter

ldursw:
Files with arc moves are now calculated correctly. Previously it would ignore arc moves and return wrong estimates.

Run `octoprint analysis gcode <file.gcode>` with a file without arcs, then with arcs and compare if the results are similar. The implementation from Arc Welder allows a small difference in the resulting file so the values between the files won't be identical.

The implementation was based on 1170914f626e165a1307dd50244e41b7d370a287.

Issue #4157

Without the patch
image

With the patch
image


foosel:
Thank you! 😊👍

Issue #4061, opened 2021-03-29
[maintenance] Error in all file analysis


cp2004:
### The problem

Since it was mentioned that 1.6.0rc is going to start tomorrow... Here's a bug:
```2021-03-29 14:04:23,720 - octoprint.filemanager.analysis - DEBUG - Got output: "DONE:0.0s\r\nERROR:Invalid analysis result, please create a bug report in OctoPrint's issue tracker and be sure to also include the GCODE file with which this happened\r\n"
2021-03-29 14:04:23,720 - octoprint.filemanager.analysis - ERROR - Analysis for local:Connection_Test.gcode ran into error: Invalid analysis result, please create a bug report in OctoPrint's issue tracker and be sure to also include the GCODE file with which this happened
```

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

OctoPrint 1.6.0.dev404+g631085f78

### Operating system running OctoPrint

Windows 10

### Printer model & used firmware incl. version

VIRTUAL

### Browser and version of browser, operating system running browser

Microsoft Edge Version 89.0.774.63 (Official build) (64-bit)

### Checklist of files to include below

- [X] octoprint.log
- [ ] serial.log or contents of your terminal tab (always include in cases of issues with printer communication)
- [ ] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
- [ ] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)
- [X] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

Any gcode file will do it, I'm always getting some kind of logged error. It can't process them at any time, if the server is restarted they are still not processed. 

Full log:
[octoprint - 2021-03-29T140714.476.log](https://github.com/OctoPrint/OctoPrint/files/6221893/octoprint.-.2021-03-29T140714.476.log)
[bundle viewer](https://bundleviewer.octoprint.org/?url=https%3A%2F%2Fgithub.com%2FOctoPrint%2FOctoPrint%2Ffiles%2F6221893%2Foctoprint.-.2021-03-29T140714.476.log)

gcode:
Any I have tried it with, but here are some of them:
https://github.com/OctoPrint/OctoPrint/blob/master/tests/cypress/fixtures/e2e-test.gcode

[Connection_Test (1).zip](https://github.com/OctoPrint/OctoPrint/files/6221925/Connection_Test.1.zip)
[...]

PR #4181, opened 2021-06-30
 🐛 Fix files with no dimensions breaking the files UI

### LINKED ISSUE #4180: [maintenance] File uploads broken, since empty metadata commit
### The problem

1. Upload a file
2. Server hangs after the analysis result is ready

Git bisect confirms it is https://github.com/OctoPrint/OctoPrint/commit/e98a5df85879d92cdad42235669c6bc6b9563254 that introduced the issue, previous commit is OK. ~~*edit* on looking, it seems maybe this should be 054b8568390c2a88bfcf1ffd67eb1c2da0ddc0de instead? Maybe I picked the wrong one on my bisect. Either way, one of these.~~ original one :)

This is on OctoPi 0.18 development setup, it doesn't seem to happen on my Windows development setup. Ran in safe mode, ran pip install to check dependencies and made no difference.

I hope it is not something specific to this setup - I can do more testing tomorrow. Since I am also not seeing the version number change in the UI/log despite pip telling me the correct number when it installs. 🤔 *edit* this could be due to checking out the specific commit, it is falling back to the last tag. The branch based versioning is not supposed to work here.

Yes, it did persist

_No response_

### Version of OctoPrint

OctoPrint-1.7.0.dev134+g1c70931e1

OctoPi 0.18

VIRTUAL

Edge 91

- [X] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

[octoprint-systeminfo-20210629224805.zip](https://github.com/OctoPrint/OctoPrint/files/6736726/octoprint-systeminfo-20210629224805.zip)

The file I'm trying to upload:
[xyzCalibration_cube.zip](https://github.com/OctoPrint/OctoPrint/files/6736732/xyzCalibration_cube.zip)

cp2004:
I had a response like
```
"date": 1614624977, 
      "display": "Misc_-_Happy_Birthday_1.gcode", 
      "gcodeAnalysis": {
        "analysisPending": false
      }, 
      "hash": "20af558d1fc68865333892cd9d0536293120e3b7", 
      "name": "Misc_-_Happy_Birthday_1.gcode", 
      "origin": "local", 
      "path": "Misc_-_Happy_Birthday_1.gcode", 
```
Can't go 3 levels deep in JS without checking that the object exists, which apparently in some cases it doesn't.

Fixes files UI in maintenance branch, see #4180 

Locally by fixing the previous issue encountered.

This seems to be a never ending cycle of me reporting bugs with empty analysis results... but I think we have got somewhere.

Closes #4180

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/files.js
+++ b/src/octoprint/static/js/app/viewmodels/files.js
@@ -967,8 +967,8 @@ $(function () {
             if (data["gcodeAnalysis"]) {
                 if (
                     data["gcodeAnalysis"]["_empty"] ||
-                    (data["gcodeAnalysis"]["dimensions"] &&
-                        data["gcodeAnalysis"]["dimensions"]["width"] === 0 &&
+                    !data["gcodeAnalysis"]["dimensions"] ||
+                    (data["gcodeAnalysis"]["dimensions"]["width"] === 0 &&
                         data["gcodeAnalysis"]["dimensions"]["depth"] === 0 &&
                         data["gcodeAnalysis"]["dimensions"]["height"] === 0)
                 ) {


foosel (commented):
Huh, I wonder how you ended up with a `gcodeAnalysis` like that. Was that from an earlier version?

In any case, while this fixes things, I think the logically more sound approach would be:

``` js
if (
    data["gcodeAnalysis"]["_empty"] ||
    !data["gcodeAnalysis"]["dimensions"] ||
    (data["gcodeAnalysis"]["dimensions"]["width"] == 0 &&
     data["gcodeAnalysis"]["dimensions"]["depth"] == 0 &&
     data["gcodeAnalysis"]["dimensions"]["height"] == 0)
) {
    // ...
}
```

That way we express more clearly that the model should be considered empty if it has `_empty` set OR no `dimensions` OR all dimensions are 0.

cp2004:
It was an older version (the file is from 1 March) - it could be one of the ones around the time #4061 was an issue, and it was completing the analysis but not getting the result? Is that how 1.6.0 is currently working?

That code should catch `gcodeAnalysis.dimensions` being undefined, as it would not pass the earlier check, so equivalent for this case but a bit more explicit.

PR #4192, opened 2021-07-21
✏️ Additional file information properties documented

### LINKED ISSUE #2682: [Request] Documentation for file information is missing properties
The [file-information](http://docs.octoprint.org/en/master/api/datamodel.html#file-information) documentation is missing (at least) the following properties:
- `prints`
- `statistics`
- `notes`
- `links`

I may be missing some more, this is based on the replies Octoprint `1.3.8` gives me, not on the source code itself.

I have read the FAQ.

cp2004:
Closes #2682

I couldn't find what the `notes` or `links` properties were referring to - plugins can add additional metadata so this may be where this comes from. It's also an old ticket, maybe things changed...

### DIFF
--- a/docs/api/datamodel.rst
+++ b/docs/api/datamodel.rst
@@ -365,6 +365,14 @@ Files
      - 0..1
      - :ref:`GCODE analysis information <sec-api-datamodel-files-gcodeanalysis>`
      - Information from the analysis of the GCODE file, if available. Left out in abridged version.
+   * - ``prints``
+     - 0..1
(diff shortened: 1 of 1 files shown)


foosel:
Thank you, don't know why I never documented that!

foosel:
`links` where a concept I put in when slicing was still a bigger focus. If you slice a file via OctoPrint's slicing interface, the resulting gcode will get a link to the STL (and iirc also vice versa). Not sure about the notes however...

PR #4222, opened 2021-09-06
Use C YAML loader when possible

flaviut:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This replaces the python YAML loader with the C++ version whenever possible. This is a lot faster:

calling `_get_options` before: .339s, .317, .220, .321, .315
calling `_get_options` after: .037s, .106, .037, .145, .089

Manually tested with `/api/connection?perfprofile`

### DIFF
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import io
 import logging as log
 import os
 import sys
@@ -391,10 +390,9 @@ def init_logging(
 
         config_from_file = {}
         if os.path.exists(logging_file) and os.path.isfile(logging_file):
-            import yaml
+            from octoprint.util import yaml
 
-            with io.open(logging_file, "rt", encoding="utf-8") as f:
-                config_from_file = yaml.safe_load(f)
+            config_from_file = yaml.load_from_file(path=logging_file)
 
         # we merge that with the default config
         if config_from_file is not None and isinstance(config_from_file, dict):
@@ -742,9 +740,7 @@ def get_plugin_blacklist(settings, connectivity_checker=None):
     import time
 
     import requests
-    import yaml
 
-    from octoprint.util import bom_aware_open
     from octoprint.util.version import is_octoprint_compatible, is_python_compatible
 
     logger = log.getLogger(__name__ + ".startup")
@@ -810,8 +806,9 @@ def get_plugin_blacklist(settings, connectivity_checker=None):
         if os.stat(path).st_mtime + ttl < time.time():
             return None
 
-        with bom_aware_open(path, encoding="utf-8", mode="rt") as f:
-            result = yaml.safe_load(f)
+        from octoprint.util import yaml
+
+        result = yaml.load_from_file(path=path)
 
         if isinstance(result, list):
             return result
@@ -824,8 +821,9 @@ def get_plugin_blacklist(settings, connectivity_checker=None):
 
             if cache is not None:
                 try:
-                    with bom_aware_open(cache, encoding="utf-8", mode="wt") as f:
-                        yaml.safe_dump(result, f)
+                    from octoprint.util import yaml
+
+                    yaml.save_to_file(result, path=cache)
                 except Exception as e:
                     logger.info(
                         "Fetched plugin blacklist but couldn't write it to its cache file: %s",
--- a/src/octoprint/access/groups.py
+++ b/src/octoprint/access/groups.py
@@ -5,18 +5,16 @@ __author__ = "Marc Hannappel <salandora@gmail.com>"
 __license__ = "GNU Affero General Public License http://www.gnu.org/licenses/agpl.html"
 __copyright__ = "Copyright (C) 2017 The OctoPrint Project - Released under terms of the AGPLv3 License"
 
-import io
 import logging
 import os
 from functools import partial
 
-import yaml
 from past.builtins import basestring
 
 from octoprint.access import ADMIN_GROUP, GUEST_GROUP, READONLY_GROUP, USER_GROUP
 from octoprint.access.permissions import OctoPrintPermission, Permissions
 from octoprint.settings import settings
-from octoprint.util import atomic_write
+from octoprint.util import atomic_write, yaml
 from octoprint.vendor.flask_principal import Need, Permission
 
 GroupNeed = partial(Need, "group")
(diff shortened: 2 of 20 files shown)


cp2004:
Does using this loader make a difference on the main settings/`config.yaml` file as well? Maybe it's already implemented there, I haven't looked.

I want to dig more into the profiler & see if there's anything I can do to help out with performance issues - I've dug into the frontend a fair bit for things that were dropping frames, but I want to get into some of the backend stuff too.

flaviut:
Oh, you're right, it does look like yaml is used extensively throughout the codebase. I'll revisit this later and update all those places too.

Even if it doesn't make a huge performance difference, this isn't something that hurts readability so it's nice to use it everywhere.

cp2004:
I knew OctoPrint had profiling in it somewhere, but wasn't entirely sure. Thanks for finding me a new toy to take a poke at, seems like very useful info.

flaviut:
I've updated the PR to use the new `safe_load` and `safe_dump` everywhere. I have not profiled it, but I think that this change is worth it just for the consistency.

jneilliii:
That's a good idea as I've recently been told of a vulnerability that if safe load isn't used there's a chance of a specially crafted payload to allow code execution.

flaviut:
To be clear, Octoprint always used `yaml.safe_load`. But this change makes it use `octoprint.utils.yaml_utils.safe_load`, which wraps around the yaml library, opportunistically using the C yaml parser when it is installed.

jneilliii:
Yeah, I know. Was just saying that your approach is sound doing the same with c yaml.

src/octoprint/util/yaml_util.py
```diff
@@ -0,0 +1,76 @@
...
+    return yaml.dump(
+        data,
+        stream=file,
+        Dumper=SafeDumper,
+        allow_unicode=True,  # no good reason not to allow it these days
+        **kwargs
+    )
+
+
+def safe_dump_pretty(data, file=None, path=None):
```
foosel, line 68: Turn into a `pretty` parameter instead and simply adjust `kwargs` passed to `yaml.dump` accordingly.

src/octoprint/util/yaml_util.py
```diff
@@ -0,0 +1,76 @@
...
+
+    try:
+        from yaml import CSafeLoader as SafeLoader
+    except ImportError:
+        from yaml import SafeLoader
+
+    return yaml.load(file, Loader=SafeLoader)
+
+
+def safe_dump(data, file=None, path=None, **kwargs):
```
foosel, line 34: `save_to_file`

src/octoprint/util/yaml_util.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+import io
+
+
+def safe_load(file=None, path=None):
```
foosel, line 5: `load_from_file`

foosel (changes requested):
Hi! That's a great change! In principle all in favour, just some issues and some minor nitpicking about the naming, see the code comments :)

For the naming: To be consistent with current code convention the package should rather be `octoprint.util.yaml`.

foosel:
Also thanks for the rebase to `maintenance`, you just saved me some steps :)

src/octoprint/util/yaml_util.py
```diff
@@ -0,0 +1,76 @@
...
+    are None, the value is returned as a string.
+
+    :rtype: str | None
+    :type data: object
+    :type path: str | None
+    :type file: typing.TextIO | None
+    """
+    if path is not None:
+        assert file is None
+        with io.open(path, "wt", encoding="utf-8") as f:
```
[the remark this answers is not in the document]
flaviut: I was hoping that this could be solved through user education, although it probably is less hassle to do it in code :(

Looks like even python 2.7 supports `utf-8-sig`, which deals with Notepad's dumb little thing. Is there a reason `bom_aware_open` was written instead of using that? Also, is there a particular reason `bom_aware_open`'s default encoding is ascii?

If not, then I'd love to delete that code in my next PR :grin:

src/octoprint/util/yaml_util.py
```diff
@@ -0,0 +1,76 @@
...
+            return safe_dump(data, file=f)
```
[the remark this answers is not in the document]
flaviut: whoops!

src/octoprint/util/yaml_util.py
flaviut: Done, but it feels kinda weird because most clients aren't passing a file or a path and are getting a string back.

src/octoprint/util/yaml_util.py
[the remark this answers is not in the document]
foosel: > I was hoping that this could be solved through user education, although it probably is less hassle to do it in code :(

User education is a noble goal that is also my preferred go-to, but simply fails in the field from my experience more often than not. And here Bad Stuff can be prevented easily with a bit of code, so it's frankly not worth the frustration of constantly seeing users run into this issue and trying to educate them on it again and again and again when all it takes is a few lines of code to solve it transparently. There are other battles to be fought that are worth more.

> Looks like even python 2.7 supports `utf-8-sig`, which deals with Notepad's dumb little thing. Is there a reason `bom_aware_open` was written instead of using that?

IIRC (and mind you, the introduction of `bom_aware_open` was over six years ago so my memory is a bit fuzzy), simply using `utf-8-sig` didn't work at the time because it would then *require* the presence of the BOM and thus fail on files without it, which is why I went through the hassle of first detecting if the `-sig` encoding was needed before using it (which is pretty much all that `bom_aware_open` does). However, I currently cannot reproduce this with my local 2.7 environment anymore, so either I'm misremembering things here, or this behaviour back then was actually buggy and got fixed since, or behaviour simply got changed when switching from `codecs.open` to `io.open` during the big Python 3 compatibility refactoring.

Considering that it no longer seems to be an issue, I guess `bom_aware_open` can be retired (though since there might be plugins out there using it it needs a deprecation annotation and then turned into a transparent wrapper around `io.open` with `-sig` attachment to the `encoding` value as needed).

> Also, is there a particular reason `bom_aware_open`'s default encoding is ascii?

IIRC, back then `open` calls defaulted to ascii, so I did too.

src/octoprint/util/yaml_util.py
foosel: Hm, I see what you mean. We actually have two uses here, dumping and saving to a file object or a path, and just dumping and returning the result. So maybe this should be two functions after all? 🤔

`dump` & `save_to_file` maybe?

foosel:
> To be consistent with current code convention the package should rather be `octoprint.util.yaml`.

Don't forget that one, I know it is a bit hidden up there in the comment 😅

src/octoprint/util/yaml_util.py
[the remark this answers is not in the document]
flaviut: resolved for https://github.com/OctoPrint/OctoPrint/issues/4228

foosel (approved):

PR #4223, opened 2021-09-08
Add static files to .ignore

flaviut:
This file is used by at least ripgrep and fd to know which files they should not look into when searching.

This change is very simple and only developer-facing, so I have not filled out the checklist.

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -1,0 +1,4 @@
+# ripgrep ignore file
+
+src/octoprint/static/vendor/
+src/octoprint/vendor/


foosel:
Should this maybe also contain `src/octoprint/vendor` given that that is the Python equivalent of the currently ignored JS only folder?

Also thanks for the pointer towards `ripgrep`!

flaviut:
Looks like you occasionally make changes to the code in that directory, so I probably wouldn't want to include it. I'd consider `venv` the python equivalent of this, and that's already ignored through the gitignore.

`src/octoprint/static/vendor/`, on the other hand, just has a bunch of minified files that are never manually edited, and they're a huge pain when searching because they tend to fill up my console.

foosel:
The only things I do change in the python part is compatibility fixes and backporting bug fixes from forks of the vendored in library. I occasionally (though way less often) do the same for the js side. So in that way it really is the same. 

Both are also excluded from the pre-commit runs.

flaviut:
Added the requested directory.

PR #4225, opened 2021-09-09
WebRTC webcam support

johnboiles:

This PR adds support for WebRTC webcams when a Stream URL beginning with `webrtc://` is specified in Webcam settings. This allows for HD (1080p) streaming from Raspberry Pis with < 1s of latency (I'm getting ~700ms with my Logitech c920 on a Pi 3B 1.2). This improves upon the MJPEG and HLS webcams by providing high resolution, low bandwidth, and low latency. There are other better-latency-than-HLS protocols out there but they don't have built-in browser support and since they're optimized for CDN distribution, they tend to have >1s latency.

|                     | MJPEG | HLS | WebRTC | [LL-DASH](https://www.theoplayer.com/blog/low-latency-dash) | [Apple LLHLS](https://developer.apple.com/documentation/http_live_streaming/enabling_low-latency_hls) | [Twitter/Community LHLS](https://medium.com/@periscopecode/introducing-lhls-media-streaming-eb6212948bef) |
|---------------------|-------|-----|--------|---------|-------------|--------------|
| High Resolution     | ❌     | ✅   | ✅      | ✅       | ✅           | ✅            |
| Low Bandwidth       | ❌     | ✅   | ✅      | ✅       | ✅           | ✅            |
| Low Latency         | ✅     | ❌   | ✅      | ⚠️       | ⚠️           | ⚠️            |
| Browser Support | ✅     | ✅   | ✅      | ❌       | ❌           | ❌            |

If accepted, the next step on this project will be to provide a WebRTC server as a part of OctoPi (similar to `ffmpeg_hls.service` for HLS support). I've started on those changes [here](https://github.com/guysoft/OctoPi/compare/devel...johnboiles:webrtc-support?expand=1).

I tested this on a Pi3 B 1.2 with a Logitech c920 camera. I tested in Safari, Chrome, and Firefox.

To run a WebRTC server you'll need to download my fork of `aiortc` and run it. I'm running it on the same Pi that has my OctoPrint instance. This is what will need to be bundled into OctoPi. 

```bash
# Checkout my branch of aiortc
git clone https://github.com/johnboiles/aiortc.git
cd aiortc
git checkout octoprint-webrtc-support

# Get some apt dependencies
sudo apt install python3-venv libavdevice-dev libavfilter-dev libopus-dev libvpx-dev pkg-config libsrtp2-dev

# Create a python virtual environment
python3 -m venv venv
source venv/bin/activate

# Upgrade pip because why not
pip install --upgrade pip

# Get dependencies specific to webcam.py
pip install aiohttp crc32c

# Install aiortc from the local source
pip install .

# [Optional] If you want to modify the python code, point to the source files for python and copy in a few compiled files
export  PYTHONPATH=$PWD/src
cp venv/lib/python3.7/site-packages/aiortc-1.2.1-py3.7-linux-armv7l.egg/aiortc/codecs/* src/aiortc/codecs/

# Stop anything using the webcam right now
sudo systemctl stop ffmpeg_hls
sudo systemctl stop webcamd

# Run the example
cd examples/webcam
python webcam.py --preferred-codec=video/H264

# Or if you're adventurous and want to try passing forward raw h264 data (works on old Logitcech c920s and maybe others)
python webcam.py --no-transcode --preferred-codec=video/H264 --video-options='{"video_size": "1920x1080", "framerate": "30", "input_format": "h264"}'

# If you're very adventurous and have a Pi camera, this might work for you:
v4l2-ctl -d /dev/video0 -v pixelformat=H264,width=1920,height=1080
v4l2-ctl -d /dev/video0 -c video_bitrate_mode=1,video_bitrate=4000000,h264_profile=4,h264_i_frame_period=60,h264_level=11,repeat_sequence_header=1
v4l2-ctl -d /dev/video0 -p 30
python webcam.py --no-transcode --preferred-codec=video/H264 --video-options='{"video_size": "1920x1080", "framerate": "30", "input_format": "h264", "pixelformat": "H264"}'

# Set your Webcam Stream URL in settings to point to your WebRTC server
# e.g. webrtc://octopi.local:8080/offer
```

See thread [here](https://community.octoprint.org/t/low-latency-h264-streaming-support-prob-w-webrtc/36996/8) for a bit more context.

None in the OctoPrint Repo. However, see also these relevant `aiortc` PRs that are needed to passthrough native h264 to WebRTC (required for 1080p on the Pi3):

* https://github.com/aiortc/aiortc/pull/559
* https://github.com/aiortc/aiortc/pull/564

Look at that 🔥 1080p 30fps video inside OctoPrint -- I had to zoom in so you can see it in its 1080p glory.

image

#### Further notes

The most uncertain part of this PR is how to define the Stream URL for WebRTC streams. The WebRTC entrypoint is just an HTTP url (e.g. `http://octopi.local:8080/offer`). I used `webrtc://octopi.local:8080/offer` since it was simple and straightforward in my opinion. The one problem with it though is that the current setup wouldn't allow you to use https for webrtc. This would be best to decide now since it'd be potentially a breaking change for any WebRTC users in the future. Some alternatives:
* `webrtc://http://octopi.local:8080/offer`
* `webrtc://octopi.local:8080/offer` (for http) AND `webrtcs://octopi.local:8080/offer` (for https)
* `webrtc:http://octopi.local:8080/offer`

Long-term (after a lot of testing) I think WebRTC could potentially be a better default than MJPEG. For low resolutions like 640x480, the Pi can handle transcoding to h264 just fine (`aiortc` supports using the `h264_omx` hardware encoder) so it should be able to support a wide variety of camera hardware at similar video quality to mjpegstreamer (and use less bandwidth).

### DIFF
--- a/src/octoprint/server/api/settings.py
+++ b/src/octoprint/server/api/settings.py
@@ -129,6 +129,7 @@ def getSettings():
             "streamUrl": s.get(["webcam", "stream"]),
             "streamRatio": s.get(["webcam", "streamRatio"]),
             "streamTimeout": s.getInt(["webcam", "streamTimeout"]),
+            "streamWebrtcIce": s.get(["webcam", "streamWebrtcIce"]),
             "snapshotUrl": s.get(["webcam", "snapshot"]),
             "snapshotTimeout": s.getInt(["webcam", "snapshotTimeout"]),
             "snapshotSslValidation": s.getBoolean(["webcam", "snapshotSslValidation"]),
@@ -549,6 +550,8 @@ def _saveSettings(data):
             s.set(["webcam", "streamRatio"], data["webcam"]["streamRatio"])
         if "streamTimeout" in data["webcam"]:
             s.setInt(["webcam", "streamTimeout"], data["webcam"]["streamTimeout"])
+        if "streamWebrtcIce" in data["webcam"]:
+            s.set(["webcam", "streamWebrtcIce"], data["webcam"]["streamWebrtcIce"])
         if "snapshotUrl" in data["webcam"]:
             s.set(["webcam", "snapshot"], data["webcam"]["snapshotUrl"])
         if "snapshotTimeout" in data["webcam"]:
--- a/src/octoprint/server/api/settings.py
+++ b/src/octoprint/server/api/settings.py
@@ -129,7 +129,7 @@ def getSettings():
             "streamUrl": s.get(["webcam", "stream"]),
             "streamRatio": s.get(["webcam", "streamRatio"]),
             "streamTimeout": s.getInt(["webcam", "streamTimeout"]),
-            "streamWebrtcIce": s.get(["webcam", "streamWebrtcIce"]),
+            "streamWebrtcIceServers": s.get(["webcam", "streamWebrtcIceServers"]),
             "snapshotUrl": s.get(["webcam", "snapshot"]),
             "snapshotTimeout": s.getInt(["webcam", "snapshotTimeout"]),
             "snapshotSslValidation": s.getBoolean(["webcam", "snapshotSslValidation"]),
@@ -550,8 +550,13 @@ def _saveSettings(data):
             s.set(["webcam", "streamRatio"], data["webcam"]["streamRatio"])
         if "streamTimeout" in data["webcam"]:
             s.setInt(["webcam", "streamTimeout"], data["webcam"]["streamTimeout"])
-        if "streamWebrtcIce" in data["webcam"]:
-            s.set(["webcam", "streamWebrtcIce"], data["webcam"]["streamWebrtcIce"])
+        if "streamWebrtcIceServers" in data["webcam"] and isinstance(
+            data["webcam"]["streamWebrtcIceServers"], (list, tuple)
+        ):
+            s.set(
+                ["webcam", "streamWebrtcIceServers"],
+                data["webcam"]["streamWebrtcIceServers"],
+            )
         if "snapshotUrl" in data["webcam"]:
             s.set(["webcam", "snapshot"], data["webcam"]["snapshotUrl"])
         if "snapshotTimeout" in data["webcam"]:
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1685,10 +1685,11 @@ var negotiateWebRTC = function (streamUrl) {
         });
 };
 
-var startWebRTC = function (videoElement, streamUrl) {
+var startWebRTC = function (videoElement, streamUrl, iceServers) {
+    iceServers = iceServers || ["stun:stun.l.google.com:19302"];
     var config = {
         sdpSemantics: "unified-plan",
-        iceServers: [{urls: ["stun:stun.l.google.com:19302"]}]
+        iceServers: [{urls: iceServers}]
     };
     pc = new RTCPeerConnection(config);
     pc.addEventListener("track", function (evt) {
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1686,4 +1686,5 @@ var startWebRTC = function (videoElement, streamUrl) {
     });
 
     negotiateWebRTC(streamUrl);
+    return pc;
 };
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1543,7 +1543,7 @@ var copyToClipboard = function (text) {
 
 var determineWebcamStreamType = function (streamUrl) {
     if (streamUrl) {
-        if (streamUrl.startsWith("webrtc://")) {
+        if (streamUrl.startsWith("webrtc")) {
             return "webrtc";
         }
 
@@ -1663,9 +1663,11 @@ var negotiateWebRTC = function (streamUrl) {
         })
         .then(function () {
             var offer = pc.localDescription;
-            streamUrl = streamUrl.slice("webrtc://".length);
+            // webrtc://host.com becomes http://host.com
+            // webrtcs://host.com becomes https://host.com
+            streamUrl = "http" + streamUrl.slice("webrtc".length);
             return $.ajax({
-                url: "http://" + streamUrl,
+                url: streamUrl,
                 type: "POST",
                 dataType: "json",
                 data: JSON.stringify({
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1651,22 +1651,19 @@ var negotiateWebRTC = function (streamUrl) {
         .then(function () {
             var offer = pc.localDescription;
             streamUrl = streamUrl.slice("webrtc://".length);
-            return fetch("http://" + streamUrl, {
-                body: JSON.stringify({
+            return $.ajax({
+                url: "http://" + streamUrl,
+                type: "POST",
+                dataType: "json",
+                data: JSON.stringify({
                     sdp: offer.sdp,
                     type: offer.type
                 }),
-                headers: {
-                    "Content-Type": "application/json"
-                },
-                method: "POST"
+                contentType: "application/json; charset=UTF-8"
             });
         })
         .then(function (response) {
-            return response.json();
-        })
-        .then(function (answer) {
-            return pc.setRemoteDescription(answer);
+            return pc.setRemoteDescription(response);
         })
         .catch(function (e) {
             console.error(e);
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1625,7 +1625,16 @@ var deepMerge = function (target, source) {
 };
 
 var isWebRTCAvailable = function () {
-    return typeof RTCPeerConnection == "function";
+    return (
+        typeof RTCPeerConnection == "function" &&
+        typeof RTCPeerConnection.prototype.addEventListener == "function" &&
+        typeof RTCPeerConnection.prototype.addTransceiver == "function" &&
+        typeof RTCPeerConnection.prototype.createOffer == "function" &&
+        typeof RTCPeerConnection.prototype.setLocalDescription == "function" &&
+        typeof RTCPeerConnection.prototype.removeEventListener == "function" &&
+        typeof RTCPeerConnection.prototype.addEventListener == "function" &&
+        typeof RTCPeerConnection.prototype.setRemoteDescription == "function"
+    );
 };
 
 var negotiateWebRTC = function (streamUrl) {
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1543,6 +1543,10 @@ var copyToClipboard = function (text) {
 
 var determineWebcamStreamType = function (streamUrl) {
     if (streamUrl) {
+        if (streamUrl.startsWith("webrtc://")) {
+            return "webrtc";
+        }
+
         var lastDotPosition = streamUrl.lastIndexOf(".");
         var firstQuotationSignPosition = streamUrl.indexOf("?");
         if (
@@ -1618,4 +1622,68 @@ var deepMerge = function (target, source) {
     });
 
     return target;
+};
+
+var negotiateWebRTC = function (streamUrl) {
+    pc.addTransceiver("video", {direction: "recvonly"});
+    pc.addTransceiver("audio", {direction: "recvonly"});
+    return pc
+        .createOffer()
+        .then(function (offer) {
+            return pc.setLocalDescription(offer);
+        })
+        .then(function () {
+            // Wait for ICE gathering to complete
+            return new Promise(function (resolve) {
+                if (pc.iceGatheringState === "complete") {
+                    resolve();
+                } else {
+                    function checkState() {
+                        if (pc.iceGatheringState === "complete") {
+                            pc.removeEventListener("icegatheringstatechange", checkState);
+                            resolve();
+                        }
+                    }
+                    pc.addEventListener("icegatheringstatechange", checkState);
+                }
+            });
+        })
+        .then(function () {
+            var offer = pc.localDescription;
+            streamUrl = streamUrl.slice("webrtc://".length);
+            return fetch("http://" + streamUrl, {
+                body: JSON.stringify({
+                    sdp: offer.sdp,
+                    type: offer.type
+                }),
+                headers: {
+                    "Content-Type": "application/json"
+                },
+                method: "POST"
+            });
+        })
+        .then(function (response) {
+            return response.json();
+        })
+        .then(function (answer) {
+            return pc.setRemoteDescription(answer);
+        })
+        .catch(function (e) {
+            console.error(e);
+        });
+};
+
+var startWebRTC = function (videoElement, streamUrl) {
+    var config = {
+        sdpSemantics: "unified-plan",
+        iceServers: [{urls: ["stun:stun.l.google.com:19302"]}]
+    };
+    pc = new RTCPeerConnection(config);
+    pc.addEventListener("track", function (evt) {
+        if (evt.track.kind == "video") {
+            videoElement.srcObject = evt.streams[0];
+        }
+    });
+
+    negotiateWebRTC(streamUrl);
 };
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1624,6 +1624,10 @@ var deepMerge = function (target, source) {
     return target;
 };
 
+var isWebRTCAvailable = function () {
+    return typeof RTCPeerConnection == "function";
+};
+
 var negotiateWebRTC = function (streamUrl) {
     pc.addTransceiver("video", {direction: "recvonly"});
     pc.addTransceiver("audio", {direction: "recvonly"});
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1686,11 +1686,12 @@ var negotiateWebRTC = function (streamUrl) {
 };
 
 var startWebRTC = function (videoElement, streamUrl, iceServers) {
-    iceServers = iceServers || ["stun:stun.l.google.com:19302"];
     var config = {
-        sdpSemantics: "unified-plan",
-        iceServers: [{urls: iceServers}]
+        sdpSemantics: "unified-plan"
     };
+    if (iceServers) {
+        config.iceServers = [{urls: iceServers}];
+    }
     pc = new RTCPeerConnection(config);
     pc.addEventListener("track", function (evt) {
         if (evt.track.kind == "video") {
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -43,6 +43,7 @@ $(function () {
         self.webcamHlsEnabled = ko.observable(false);
         self.webcamWebRTCEnabled = ko.observable(false);
         self.webcamError = ko.observable(false);
+        self.webRTCPeerConnection = null;
 
         self.keycontrolActive = ko.observable(false);
         self.keycontrolHelpActive = ko.observable(false);
@@ -739,7 +740,22 @@ $(function () {
         self._switchToWebRTCWebcam = function () {
             var video = document.getElementById("webcam_webrtc");
 
-            startWebRTC(video, self.settings.webcam_streamUrl());
+            // Close any existing, disconnected connection
+            if (
+                self.webRTCPeerConnection != null &&
+                self.webRTCPeerConnection.connectionState != "connected"
+            ) {
+                self.webRTCPeerConnection.close();
+                self.webRTCPeerConnection = null;
+            }
+
+            // Open a new connection if necessary
+            if (self.webRTCPeerConnection == null) {
+                self.webRTCPeerConnection = startWebRTC(
+                    video,
+                    self.settings.webcam_streamUrl()
+                );
+            }
 
             self.webcamMjpgEnabled(false);
             self.webcamHlsEnabled(false);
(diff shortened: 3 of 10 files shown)


src/octoprint/static/js/app/helpers.js
```diff
@@ -1616,3 +1620,67 @@ var deepMerge = function (target, source) {
...
+                        }
+                    }
+                    pc.addEventListener("icegatheringstatechange", checkState);
+                }
+            });
+        })
+        .then(function () {
+            var offer = pc.localDescription;
+            streamUrl = streamUrl.slice("webrtc://".length);
+            return fetch("http://" + streamUrl, {
```
cp2004, line 1651: OctoPrint generally supports older browsers still, not sure if fetch is generally able to be used or not. @foosel  has never given me a definitive list of what browsers OctoPrint will and won't support...

Most of the rest of the codebase uses jQuery/ajax rather than fetch.

cp2004:
I'll admit I was not expecting this PR to happen with as few lines as this.

jneilliii:
If we're going to have this many layers of webcam options wouldn't it be more appropriate to split out the type into a pull-down/select list? This way if some other technology wants to be integrated, it should be a little easier to add long-term. I'm curious since both webrtc and hls use video tag, if it would be even easier to just use the same container and adjust the bindings for visibility. 

BTW, I could easily wrap this in a plugin as well like I did for HLS to allow for easier testing.

johnboiles:
> I'll admit I was not expecting this PR to happen with as few lines as this.

Haha me neither! But a lot of the work was done in `aiortc` to get it to efficiently pass through h264 without transcoding.

> wouldn't it be more appropriate to split out the type into a pull-down/select list

Probably! That's probably a cleaner solution than inventing unique Stream URL schemes.

src/octoprint/static/js/app/helpers.js
johnboiles: Ah good catch, would a  `try`/`catch` around the body of `startWebRTC` be enough do you think?

johnboiles: (I'm kinda new at JS, apologies if this question is basic)

johnboiles:
> wouldn't it be more appropriate to split out the type into a pull-down/select list

Who is the right decision maker to decide the right approach here? Is it just @foosel or are there others who have authority to make this sort of product decision?

jneilliii:
foosel will definitely have the ultimate decision and will comment eventually. A lot of the times PR review happens at the beginning of the week, but she's currently in the process of prepping 1.7.0rc3 for release after just getting back from vacation so it may be a bit.

johnboiles:
Sounds good! Thanks for the info!

foosel:
Just to confirm, I've seen this, it's on my TODO list, hope to get around to it today or tomorrow.

src/octoprint/static/js/app/helpers.js
foosel: > @foosel has never given me a definitive list of what browsers OctoPrint will and won't support...

Goal for now: ES5 compatibility. That seems to be the deciding factor for many old tablets out there. So:

https://caniuse.com/?search=es5

Whatever else is supported under these goes. The fetch API sadly doesn't, but it shouldn't be too big of an issue to achieve the same with the jquery `ajax` functions if I'm not mistaken.

src/octoprint/static/js/app/viewmodels/control.js
```diff
@@ -524,6 +526,8 @@ $(function () {
                 self._switchToMjpgWebcam();
             } else if (streamType == "hls") {
                 self._switchToHlsWebcam();
+            } else if (streamType == "webrtc") {
```
foosel, line 529: This needs a compatibility check to determine if webrtc is supported. Only then should we attempt to switch.

src/octoprint/static/js/app/viewmodels/settings.js
```diff
@@ -453,6 +454,16 @@ $(function () {
                     hls.loadSource(self.webcam_streamUrl());
                     hls.attachMedia(video_element);
                 }
+            } else if (streamType == "webrtc") {
```
foosel, line 457: Compatibility check.

src/octoprint/static/js/app/viewmodels/control.js
```diff
@@ -729,6 +734,32 @@ $(function () {
 
             self.webcamMjpgEnabled(false);
             self.webcamHlsEnabled(true);
+            self.webcamWebRTCEnabled(false);
+        };
+
+        self._switchToWebRTCWebcam = function () {
+            var video = document.getElementById("webcam_webrtc");
```
foosel, line 741: Compatibility check.

foosel (changes requested):
Have a bunch of points, see below.

About the select, IMHO for now we can still go with the URL scheme/format based approach. Should be straightforward enough and save us the hassle of implementing and then cramming in yet another setting.

johnboiles:
Thanks @foosel!

Any preference on how to adopt the scheme to support https? Some ideas:

* `webrtc://octopi.local:8080/offer` (for http) AND `webrtcs://octopi.local:8080/offer` (for https)
* `webrtc:http://octopi.local:8080/offer`

src/octoprint/static/js/app/helpers.js
```diff
@@ -1616,3 +1620,78 @@ var deepMerge = function (target, source) {
 
     return target;
 };
+
+var isWebRTCAvailable = function () {
```
johnboiles, line 1624: From my Googling, there didn't seem to be a canonical way of checking for WebRTC compatibility. The most fully-featured option seems to be [DetectRTC.js](https://github.com/muaz-khan/DetectRTC) but that sure seems like a heavy dependency to add for a simple check. I also saw articles like [this](https://blog.ivrpowers.com/post/coding/browser-webrtc/) but they seemed a little sketchy (ie does the presence of `navigator.getUserMedia` definitively mean WebRTC is supported and vice versa?).

So I opted for the simplest solution -- just check if RTCPeerConnection and all the methods we're going to call exist.

Related: There's also [adapter.js](https://github.com/webrtcHacks/adapter) which attempts to patch naming disparities in the WebRTC implementations and is recommended in the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API). If we start seeing wider WebRTC adoption in OctoPrint, then it might be worth bringing that dependency in to support a few more browsers.

johnboiles:
@foosel requested changes are ready for review.

foosel:
> Any preference on how to adopt the scheme to support https?

My first intuition after seeing the question was `webrtcs;//`, and since that is also your first suggestion I would say let's go for that ;)

src/octoprint/static/js/app/helpers.js
foosel: I would have gone with that solution as well given the situation 👍

foosel (approved):
Looking good now!

foosel:
Just one minor change: This should go into `maintenance` instead of `devel`. It's an improvement of an existing feature, not a full blown rewrite of something, so there's nothing against including it in 1.8.0. If you could rebase your changes against `maintenance`, I'll be happy to merge :)

foosel (changes requested):
(sorry for the chaos, saw the target branch too late, I need to add some labeling for that I think)

johnboiles:
Working on it now. I'll target `maintenance` and add `webrtcs`!

johnboiles:
@foosel done: Ready for review.

johnboiles:
Also FYI I snuck in one more commit that adds the `controls` attribute to the `<video>` tag. This allows the browser to show controls on the video which enables platform specific video features (like picture-in-picture, or fullscreen).

image

image

johnboiles:
Ah 'E2E tests` are getting cancelled because [ubuntu-16.04 is going away in GH actions next week](https://github.com/actions/virtual-environments/issues/3287). So I made #4234. Let's see if the build succeeds there.

src/octoprint/static/js/app/helpers.js
```diff
@@ -1619,3 +1623,80 @@ var deepMerge = function (target, source) {
...
+        })
+        .catch(function (e) {
+            console.error(e);
+        });
+};
+
+var startWebRTC = function (videoElement, streamUrl) {
+    var config = {
+        sdpSemantics: "unified-plan",
+        iceServers: [{urls: ["stun:stun.l.google.com:19302"]}]
```
foosel, line 1691: Hm... should we maybe turn this into an (advanced) option? There are people out there who prefer no contact with google services of any kind, so allowing to specific the stun server to use would be a good idea.

foosel: I've pushed a commit to address this, please check on your end as well.

I tested this based on the instructions from your OP and the feed works. However, I'm seeing an interesting issue where it only works *once*. I have to restart the server side between page loads or all that I get is a blank video. No idea if that is an issue on my end or with the server. Tested with a C270 and the transcoding command line.

johnboiles: Good call thanks for handling that!

It's almost certainly a problem with the server, that thing is still rough around the edges. Did you pull the latest from my `aiortc` `octoprint-webrtc-support` branch? I made a few tweaks on Sunday. Anything interesting in stdout from the server?

The good thing about the front end code in this PR is that it should be pretty agnostic to server implementation so we should be able to swap in any WebRTC server if `aiortc` doesn't fit our needs (Janus for example is the incumbent option).

foosel: Nothing on stdout, looks just like it should work. And this was a clone from this morning. Eh 🤷‍♀️

src/octoprint/templates/snippets/settings/webcam/webcamStreamWebrtcIce.jinja2
```diff
@@ -0,0 +1,7 @@
+<div class="control-group" title="{{ _('Endpoint to use for WebRTC ICE')|edq }}">
```
johnboiles, line 1: I think `ICE server` would be slightly more consistent terminology. It could also be made optional. Taking a stab at updating the copy here: `ICE server (STUN and/or TURN) to use for WebRTC (optional)`

if a STUN server isn't used clients may still be able to negotiate a direct connection but it depends how what layer of network are between them. Even on my LAN I've noticed that STUN makes connections more reliable.


https://datatracker.ietf.org/doc/html/rfc8489
https://developer.mozilla.org/en-US/docs/Web/API/RTCConfiguration/iceServers

foosel (approved):

src/octoprint/templates/snippets/settings/webcam/webcamStreamWebrtcIce.jinja2
foosel: Rephrased, does that look better?

src/octoprint/templates/snippets/settings/webcam/webcamStreamWebrtcIce.jinja2
```diff
@@ -0,0 +1,7 @@
...
+    <label class="control-label" for="settings-webcamStreamWebrtcIce">{{ _('WebRTC ICE endpoint') }}</label>
+    <div class="controls">
+        <input type="text" class="input-block-level" data-bind="value: webcam_streamWebrtcIce, valueUpdate: 'afterkeydown'" id="settings-webcamStreamWebrtcIce">
+        <span class="help-block"><span class="label">{{ _('WebRTC') }}</span> {% trans %}Only used for WebRTC streams. Provide the endpoint to use for WebRTC's Interactive Connectivity Establishment (ICE). Usually a STUN server. If unsure, leave at the default of <code>stun:stun.l.google.com:19302</code>.{% endtrans %}</span>
```
johnboiles: This can be a `STUN` or `TURN` server. [Docs](https://developer.mozilla.org/en-US/docs/Web/API/RTCConfiguration/iceServers)

`Only used for WebRTC streams. Provide the ICE or STUN server to use for WebRTC's Interactive Connectivity Establishment (ICE). If unsure, leave the default of <code>stun:stun.l.google.com:19302</code>.`

johnboiles: The short explanation is that a STUN server just helps negotiate NAT traversal for UDP while a TURN server relays all the traffic. Either can be used to get around NATs, but TURN servers aren't free because of the bandwidth usage.

src/octoprint/server/api/settings.py
```diff
@@ -129,6 +129,7 @@ def getSettings():
             "streamUrl": s.get(["webcam", "stream"]),
             "streamRatio": s.get(["webcam", "streamRatio"]),
             "streamTimeout": s.getInt(["webcam", "streamTimeout"]),
+            "streamWebrtcIce": s.get(["webcam", "streamWebrtcIce"]),
```
johnboiles, line 132: You might consider naming this `streamWebrtcIceServer`. Your call!

src/octoprint/static/js/app/helpers.js
```diff
@@ -1685,10 +1685,11 @@ var negotiateWebRTC = function (streamUrl) {
         });
 };
 
-var startWebRTC = function (videoElement, streamUrl) {
+var startWebRTC = function (videoElement, streamUrl, iceServers) {
+    iceServers = iceServers || ["stun:stun.l.google.com:19302"];
     var config = {
         sdpSemantics: "unified-plan",
-        iceServers: [{urls: ["stun:stun.l.google.com:19302"]}]
+        iceServers: [{urls: iceServers}]
```
johnboiles, line 1692: If you make this optional, you should probably break this into a separate check and don't set `iceServers` at all if the setting is unset.

For example:
https://github.com/aiortc/aiortc/blob/main/examples/webcam/client.js#L49

src/octoprint/static/js/app/helpers.js
johnboiles: 🤷 interesting. You can always add a `-vv` flag to the `webcam.py` command to get verbose output. But I think fine to punt on it for now -- I'll test your commit here on my end as well.

johnboiles:
(Apologies for the slow drip of feedback -- I originally thought I had just one comment)

src/octoprint/templates/snippets/settings/webcam/webcamStreamWebrtcIce.jinja2
johnboiles: Looks great!

johnboiles:
Tested on my end and everything is working just fine!

foosel:
All good ^^ But you frankly are way more knowledgeable about all this than I, so I suggest *you* give the copy and logic a final touch and then we merge this.

johnboiles:
Will do. While we're here I might make it also support a comma separated list of ICE servers for completeness. LMK if you have any issue with that.

foosel:
Make it comma separated on the frontend only, should be a list in the backend. 

AFAICR just needs additions to 

https://github.com/OctoPrint/OctoPrint/blob/0e1ada7a0e8445a86da329ebd222fa03020c6e90/src/octoprint/static/js/app/viewmodels/settings.js#L993

and

https://github.com/OctoPrint/OctoPrint/blob/0e1ada7a0e8445a86da329ebd222fa03020c6e90/src/octoprint/static/js/app/viewmodels/settings.js#L1223

foosel (approved):

johnboiles:
Oh oops sorry i force pushed over you. We spotted the same bug :)

foosel:
As long as everything is green ;)

johnboiles:
🎉 🎉 🎉 Thanks @foosel! I'll keep working on buttoning up the server for distribution in OctoPi.

kennethjiang:
Great job guys, especially @johnboiles . I implemented a similar thing for TSD but it's a lot uglier since it was not integrated with octopi. Looking forward to this feature making its way to the next octopi release!

QuinnDamerell:
+1 this is great stuff and will be an amazing improvement for the entire maker community! 🎉

Triodes:
@johnboiles @foosel Hey hey! First of all: love the work! ❤️ 

### Background
I stumbled across this pull request while exploring my options for using (old) mobile phones as camera's for OctoPrint. I currently have one set up using an ip webcam that exposes the camera as an mjpg stream, but it's slow and on higher resolutions will cause the phone to become hot and throttle. Bandwith is also an issue when the wifi signal is not as good.

So when I saw this I jumped on it and tried to see if I could make an android app that would properly respond to the implementation in this PR. However I ran into some issues with the signalling mechanism that's being used. 

Some context: The webRTC standard does not specify how clients should pass connection offers/answers/candidates to eachother. The end user is responsible for implementing this "signalling" mechanism. This PR has a very simple solution:

1. Send an offer via an ajax call
2. Recieve the answer in the response of this ajax call

### The problem: 
The answer created by the official Android webRTC library contains no ICECandidates. Even though these candidates will start generating as soon as you create an answer and an event is fired every time a new candidate is found, these are not added to the answer that was initially created, nor have I figured out how to add them to the answer.

So when I send my answer (which contains no candidates) and octoprint receives it, the connection fails. 

Commonly a signalling server is used to send offers/answers/candidates between clients. Looking at the 2 flows in the [MDN docks](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Signaling_and_video_calling) it seems that the exchange of candidates is not possible with this PR since there is only 1 ajax request with an offer and answer in the corresponding response. So when my candidates become available I have no way of sending them back to the browser.

### The goal
Have a discussion about how we want to implement the signalling for OctoPrint. i.e.

- What should the protocol look like?
- Who is responsible for implementing the signalling server?
  - The camera side?
  - OctoPrint client side?
  - OctoPrint server side?

I look forward to your reply!

Kind regards,
Aron

johnboiles:
Huh weird that android doesn't send ICE candidates. How does it expect to negotiate a connection if it doesn't communicate how it can be reached.

This looks like the same issue
https://github.com/aiortc/aiortc/issues/333

Does this help?
https://stackoverflow.com/a/69544931

Surely there's a way to make this work on the android side. Seems like an android quirk to solve if Android native webrtc is behaving significantly differently than web browser webrtc.

Triodes:
> Huh weird that android doesn't send ICE candidates. How does it expect to negotiate a connection if it doesn't communicate how it can be reached.

That's because the candidates are sent using ICE trickling. Basically after the offer or answer is sent the signalling mechanism is responsible for sending the candidates in separate messages. [This document](https://webrtc.org/getting-started/peer-connections) explains it very clearly

> This looks like the same issue [aiortc/aiortc#333](https://github.com/aiortc/aiortc/issues/333)

Yep, exactly the same

> Does this help? https://stackoverflow.com/a/69544931

I'm gonna see.

> Surely there's a way to make this work on the android side. Seems like an android quirk to solve if Android native webrtc is behaving significantly differently than web browser webrtc.

I don't think it's behaving differently. The signalling is simply not part of the protocol, so we have to do that ourselves. In this case the implementation does not seem to support trickling.

This part of the docs is the most relevant I think:
>Once a RTCPeerConnection object is created, the underlying framework uses the provided ICE servers to gather candidates for connectivity establishment (ICE candidates). The event icegatheringstatechange on RTCPeerConnection signals in what state the ICE gathering is (new, gathering or complete).
>
>While it is possible for a peer to wait until the ICE gathering is complete, it is usually much more efficient to use a "trickle ice" technique and transmit each ICE candidate to the remote peer as it gets discovered. This will significantly reduce the setup time for the peer connectivity and allow a video call to get started with less delays.
>
>To gather ICE candidates, simply add a listener for the icecandidate event. The RTCPeerConnectionIceEvent emitted on that listener will contain candidate property that represents a new candidate that should be sent to the remote peer (See Signaling).

So I know I could wait until gathering candidates is complete before sending the answer from android to octoprint. But I haven't figured out yet how I need to add candidates to the answer....

Triodes:
>>Does this help? https://stackoverflow.com/a/69544931
>
>I'm gonna see.

I did not set these values, but changing them seemed to make things worse. At that point the ICE candidate gathering process would not even start.

Triodes:
I made some progress: I managed to wait with sending the answer until the ICE gathering was complete. At that point I need to "regenerate" the answer and then it will have some ICE candidates in the SDP.

When I send out that answer as a reponse to the ajax call I actually get an event in my android app that the connection has succeeded and in octoprint we no longer have a black window but a "loading spinner"
image

Still no video though...
There is absolutely nothing coming into the chrome dev tools console regarding this. No warnings, no logs. Nothing about the RTCPeerConnection in general.

So I'm stuck again...

Triodes:
In the meantime I have also tried to update the octoprint code to use a socket.io based signalling mechanism:
``` javascript
var negotiateWebRTC = function (streamUrl) {
    pc.addTransceiver("video", {direction: "recvonly"});
    pc.addTransceiver("audio", {direction: "recvonly"});

    streamUrl = "http" + streamUrl.slice("webrtc".length);
    var sock = io(streamUrl);

    sock.on("connect", () => {
        sock.emit("ident", "client");
        return pc
            .createOffer()
            .then(function (offer) {
                return pc.setLocalDescription(offer);
            })
            .then(function () {
                var offer = pc.localDescription;
                sock.emit("message", pc.localDescription.toJSON());
            })
            .catch(function (e) {
                console.error(e);
            });
    });

    // Use a queue because sometimes candidates arrive before the initial answer
    var queue = []

    sock.on("message", function(message) {
        if (message.type == "answer") {
            pc.setRemoteDescription(message);
            queue.forEach(candidate => pc.addIceCandidate(candidate));
        }
        if (message.type == "candidate") {
            if(!pc || !pc.remoteDescription || !pc.remoteDescription.type){
                queue.push(message);
            } else {
                pc.addIceCandidate(message);
            }            
        }
    })

    sock.connect();
};
```
Between the android app and octoprint is a small node.js server that relays the socket messages between the two. My goal is to integrate that part into the android app as wel eventually.

Lo and behold:
image

jneilliii:
It may make more sense for you guys to continued this development discussion on the forum or discord, not here in a merged PR.

gbeatty:
hey @johnboiles , with this PR merged into Octoprint and now officially released with 1.8, is there any update on the PR to get the WebRTC server into OctoPi?

jneilliii:
It's been discussed further in the Discord server with others, but haven't seen John around in a while. Specifically cryxsd (OctoApp developer), cp2004, and Triodes has played around with it some.

johnboiles:
@gbeatty I haven't had time to land this and it's a busy season for me now. My progress is [here](https://github.com/guysoft/OctoPi/compare/devel...johnboiles:webrtc-support?expand=1) and I'd love some help getting it across the finish line.

Triodes:
@johnboiles I would love to see your aiortc setup in action sometimes and maybe I can help on the final touches. Send me a ping on the discord if you're interested

PR #4229, opened 2021-09-11
Switch bom_aware_open to use utf-8-sig by default

### LINKED ISSUE #4228: Use utf-8-sig for bom_aware_open
This simplifies our code and allows us to avoid opening the same file twice.

Context:

> Looks like even python 2.7 supports `utf-8-sig`, which deals with Notepad's dumb little thing. Is there a reason `bom_aware_open` was written instead of using that?

IIRC (and mind you, the introduction of `bom_aware_open` was over six years ago so my memory is a bit fuzzy), simply using `utf-8-sig` didn't work at the time because it would then *require* the presence of the BOM and thus fail on files without it, which is why I went through the hassle of first detecting if the `-sig` encoding was needed before using it (which is pretty much all that `bom_aware_open` does). However, I currently cannot reproduce this with my local 2.7 environment anymore, so either I'm misremembering things here, or this behaviour back then was actually buggy and got fixed since, or behaviour simply got changed when switching from `codecs.open` to `io.open` during the big Python 3 compatibility refactoring.

Considering that it no longer seems to be an issue, I guess `bom_aware_open` can be retired (though since there might be plugins out there using it it needs a deprecation annotation and then turned into a transparent wrapper around `io.open` with `-sig` attachment to the `encoding` value as needed).

> Also, is there a particular reason `bom_aware_open`'s default encoding is ascii?

IIRC, back then `open` calls defaulted to ascii, so I did too.

_Originally posted by @foosel in https://github.com/OctoPrint/OctoPrint/pull/4222#discussion_r706026141_

flaviut:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This changes the API of bom_aware_open to match the default python
encoding of utf-8.

It also avoids the double open from before by using the `utf-8-sig`
encoding

Fixes #4228

### DIFF
--- a/src/octoprint/util/__init__.py
+++ b/src/octoprint/util/__init__.py
@@ -1078,27 +1078,11 @@ else:
             self.cleanup()
 
 
-def bom_aware_open(filename, encoding="ascii", mode="r", **kwargs):
-    import codecs
-
+def bom_aware_open(filename, encoding="utf-8", mode="r", **kwargs):
     assert "b" not in mode, "binary mode not support by bom_aware_open"
 
-    codec = codecs.lookup(encoding)
-    encoding = codec.name
-
-    if kwargs is None:
-        kwargs = {}
-
-    potential_bom_attribute = "BOM_" + codec.name.replace("utf-", "utf").upper()
-    if "r" in mode and hasattr(codecs, potential_bom_attribute):
-        # these encodings might have a BOM, so let's see if there is one
-        bom = getattr(codecs, potential_bom_attribute)
-
-        with io.open(filename, mode="rb") as f:
-            header = f.read(4)
-
-        if header.startswith(bom):
-            encoding += "-sig"
+    if encoding in {"utf-8", "utf8"} and "r" in mode:
+        encoding = "utf-8-sig"
 
     return io.open(filename, encoding=encoding, mode=mode, **kwargs)
 
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -810,7 +810,7 @@ def get_plugin_blacklist(settings, connectivity_checker=None):
         if os.stat(path).st_mtime + ttl < time.time():
             return None
 
-        with bom_aware_open(path, encoding="utf-8", mode="rt") as f:
+        with bom_aware_open(path, mode="rt") as f:
             result = yaml.safe_load(f)
 
         if isinstance(result, list):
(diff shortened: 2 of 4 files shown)


src/octoprint/util/__init__.py
```diff
@@ -1078,27 +1078,11 @@ def __exit__(self):
...
-        # these encodings might have a BOM, so let's see if there is one
-        bom = getattr(codecs, potential_bom_attribute)
-
-        with io.open(filename, mode="rb") as f:
-            header = f.read(4)
-
-        if header.startswith(bom):
-            encoding += "-sig"
+    if encoding in {"utf-8", "utf8"} and "r" in mode:
+        encoding = "utf-8-sig"
```
foosel, line 1085: This turns this into a breaking change. The old code will work with any encoding that supports a BOM (as retrieved from `codecs`), this code here only works with `utf8`. While it is extremely unlikely that there is client code out there using anything but UTF8 with `bom_aware_open`, we can't rule it out and thus this change would go against semver.

Suggestion:

```suggestion
    import codecs
    codec = codecs.lookup(encoding)
    potential_bom_attribute = "BOM_" + codec.name.replace("utf-", "utf").upper()
    if hasattr(codecs, potential_bom_attribute):
        encoding = codec.name + "-sig"
```

src/octoprint/util/__init__.py
```diff
@@ -1078,27 +1078,11 @@ def __exit__(self):
             self.cleanup()
 
 
-def bom_aware_open(filename, encoding="ascii", mode="r", **kwargs):
-    import codecs
-
+def bom_aware_open(filename, encoding="utf-8", mode="r", **kwargs):
```
foosel, line 1081: Given that the long time goal here should be to get rid of `bom_aware_open` altogether, how about:

```suggestion
@deprecated("Please use io.open with '-sig' encoding instead", since="1.8.0")
def bom_aware_open(filename, encoding="utf-8-sig", mode="r", **kwargs):
```

foosel (changes requested):
Two open points, see above.

src/octoprint/util/__init__.py
flaviut: This isn't quite right, because only utf-8 supports the `-sig`. [UTF-16 is required to handle a BOM by default, and so is UTF-32, although it isn't typically used.](https://en.wikipedia.org/wiki/Byte_order_mark#Usage)

foosel:
Agreed. Closing this, will deprecate `bom_aware_open` in 1.8, slated for removal with 2.0 and switch the calls to it in core to `io.open`.

PR #4230, opened 2021-09-11
Improve settings processing time

flaviut:
This brings the load time of /api/settings on my crappy ARM board from
1s to 0.4s

This is done by

- optimizizing the hot loop inside deep_dict, which provides most the
gain (to 0.46s)
- using pickle for deep-copying, which is fast when doing lots of
pointer chasing, but slow when doing long copies (to 0.37s)

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -546,7 +546,9 @@ class HierarchicalChainMap(ChainMap):
     def deep_dict(self):
         def deep_dict_inner(root):
             return {
-                key: deep_dict_inner(root[key]) if isinstance(value, dict) else value
+                key: deep_dict_inner(self.__class__._get_next(key, root))
+                if isinstance(value, dict)
+                else value
                 for key, value in root.items()
             }
 
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -22,7 +22,6 @@ __author__ = "Gina Häußge <osd@foosel.net>"
 __license__ = "GNU Affero General Public License http://www.gnu.org/licenses/agpl.html"
 __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms of the AGPLv3 License"
 
-import copy
 import fnmatch
 import io
 import logging
@@ -56,6 +55,7 @@ from octoprint.util import (
     CaseInsensitiveSet,
     atomic_write,
     dict_merge,
+    fast_deepcopy,
     generate_api_key,
     is_hidden_path,
 )
@@ -543,17 +543,14 @@ class DuplicateFolderPaths(InvalidSettings):
 
 
 class HierarchicalChainMap(ChainMap):
-    def deep_dict(self, root=None):
-        if root is None:
-            root = self
+    def deep_dict(self):
+        def deep_dict_inner(root):
+            return {
+                key: deep_dict_inner(root[key]) if isinstance(value, dict) else value
+                for key, value in root.items()
+            }
 
-        result = {}
-        for key, value in root.items():
-            if isinstance(value, dict):
-                result[key] = self.deep_dict(root=self.__class__._get_next(key, root))
-            else:
-                result[key] = value
-        return result
+        return deep_dict_inner(self)
 
     def has_path(self, path, only_local=False, only_defaults=False):
         if only_defaults:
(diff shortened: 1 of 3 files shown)


src/octoprint/settings.py
```diff
@@ -1790,19 +1787,13 @@ def _get_value(
                 except KeyError:
                     raise NoSuchSettingsPath()
 
-            if preprocessors is not None:
```
flaviut, line 1793: note: this has no significant effect on performance, but it is redundant.

src/octoprint/settings.py
```diff
@@ -543,17 +543,14 @@ def __str__(self):
...
-            root = self
 
-        result = {}
-        for key, value in root.items():
-            if isinstance(value, dict):
-                result[key] = self.deep_dict(root=self.__class__._get_next(key, root))
-            else:
-                result[key] = value
-        return result
+    def deep_dict(self):
```
flaviut, line 547: I looked into caching the results of this calculation, but I seemed like the cache was not being used. I'm not sure if it was being invalidated, or if we construct a bunch of these from scratch.

flaviut:
There's a couple failing tests, but I'm not sure if there's a bug here or not. I would have expected the last three statements too all be equal, either all returning `True`, or all throwing a `KeyError`:

```python
>>> self._map = HierarchicalChainMap(
...   {'api': {'key': 'test'}},
...   {'api': {'enabled': True, 'key': None}})
>>> self._map.get_by_path(['api'])
{'key': 'test'}
>>> self._map.get_by_path(['api', 'enabled'])
True
>>> self._map.get_by_path(['api'])['enabled']
{KeyError}'enabled'
>>> self.effective['api']['enabled']
{KeyError}'enabled'
```

foosel (changes requested):
See below for comments.

src/octoprint/settings.py
flaviut: I stored the cached value as `self._deep_dict_cache`, and cleared it whenever a write operation was performed. I saw no significant performance improvement.

foosel: Can you take a look at eb69bd6fa90a01258063f3049f98a0335cbf7dbe that I just pushed and give it a try? I think that should achieve some caching, at least I saw several hits during page loads.

foosel:
> There's a couple failing tests, but I'm not sure if there's a bug here or not. I would have expected the last three statements too all be equal, either all returning `True`, or all throwing a `KeyError`:

I agree, that look counter intuitive and needs some deep dive on my end.

src/octoprint/settings.py
flaviut: That doesn't invalidate the cache often enough, sometimes `Settings` writes to `ChainMap.maps`. I implemented the caching in https://github.com/flaviut/OctoPrint/commit/788f6f97143ff3207663bda565a8713c2eb7b04d, and didn't see an improvement.

I also re-wrote the settings storage to essentially flatten the dictionaries as

```
{ a: b, c: {d: 1, e: 2}, f: {} }
a = b
c.d = 1
c.e = 2
f = {}
```

in https://github.com/flaviut/OctoPrint/commit/34004a299e8d59ba593363618e2dc7c754edd21e, but I haven't gotten a chance to test the performance of that, since I haven't been able to make it pass all the tests

foosel: Oh right, I totally forgot about the overlays. I rolled the caching back for now. Let's concentrate on the failing tests.

foosel:
> I agree, that look counter intuitive and needs some deep dive on my end.

Correction, that is correct behaviour and the reason for the existence of the `merge` parameter.

What makes the two tests fail is that 

``` python
    def deep_dict(self):
        def deep_dict_inner(root):
            return {
                key: deep_dict_inner(root[key]) if isinstance(value, dict) else value
                for key, value in root.items()
            }

        return deep_dict_inner(self)
```

isn't logically equivalent to 

``` python
    def deep_dict(self, root=None):
        if root is None:
            root = self

        result = {}
        for key, value in root.items():
            if isinstance(value, dict):
                result[key] = self.deep_dict(root=self.__class__._get_next(key, root))
            else:
                result[key] = value
        return result
```

This however apparently is and makes the tests pass:

``` python
    def deep_dict(self):
        def deep_dict_inner(root):
            return {
                key: deep_dict_inner(self.__class__._get_next(key, root))
                if isinstance(value, dict)
                else value
                for key, value in root.items()
            }

        return deep_dict_inner(self)
```

And I will admit that I have no idea right now why because from where I stand, this should behave the same 🤔

Still going to push that change for now.

foosel (approved):

PR #4234, opened 2021-09-14
Move away from `ubuntu-16.04` in GH actions

johnboiles:
Because it's going to be gone on September 20th 2021: https://github.com/actions/virtual-environments/issues/3287

I'm conservatively just bumping to `ubuntu-18.04` since there was probably a reason this was set to `ubuntu-16.04` while all the other actions run on `ubuntu-latest`


foosel:
TBH, I have no idea why it was set to that. Might have been that it was the first ever workflow I built and back then this was part of the examples 😅

PR #4235, opened 2021-09-14
Send Server-Timing header

flaviut:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Sends the Server-Timing header with each api request.

This is useful to be able to tell when a performance issue is due to the
network or due to the actual execution time

There's hardly any overhead, so send it with every request.

screenshot of app timing

I had octoprint make some requests with the network debugger pane open.

This is helpful in understanding long load times. Turns out that that in my situation, many of the requests being sent take a long time to return not because they're slow, they're < 10ms, but because the system I'm running them on has very poor WiFi latency.

### DIFF
--- a/src/octoprint/server/__init__.py
+++ b/src/octoprint/server/__init__.py
@@ -1392,6 +1392,10 @@ class Server(object):
         @app.before_request
         def before_request():
             g.locale = self._get_locale()
+
+            # used for performance measurement
+            g.start_time = octoprint.util.monotonic_time()
+
             if self._debug and "perfprofile" in request.args:
                 try:
                     from pyinstrument import Profiler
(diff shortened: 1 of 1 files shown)


foosel (changes requested):
Good idea, but should be using a monotonic time source.

I'm wondering if it would make sense to add a similar wrapper around the Tornado context, right now it will only measure the requests against the flask based middleware.

flaviut:
> I'm wondering if it would make sense to add a similar wrapper around the Tornado context, right now it will only measure the requests against the flask based middleware.

I looked at the RequestHandler interface, and I'm not sure it's always possible. I think tornado allows streaming the output, which wouldn't allow us to know the duration up front.

foosel:
Ah, good point. In that case let's leave it like it is for now.

foosel (approved):

PR #4236, opened 2021-09-15
settings: reimplement storage chainmap

flaviut:
I have a lot of trouble understanding the settings implementation, and I feel like it can be simplified.

I don't expect this to be merged, so I'm not looking for any concrete code suggestions. My goal is to discuss this a bit.

This implementation stores the settings as a chainmap of full paths rather than as a set of nested chainmaps. This means that `{"a": "a", "b": {"c": "c"}, "d": {}}` is stored as:

```
a = a
b.c = c
d = {}
```

There are still layers--but they are not nested in any way, so they are much easier to reason about.

Behavior changes:

- `HierarchicalChainMap.get(asdict=True)` is eliminated. It behaves as if this is always true.
- I'd suggest that `HierarchicalChainMap.get(merged=True)` be eliminated and behave as if it is always true, as it is confusing how `settings.effective['foo']['bar'] != settings.get(['foo', 'bar'], merged=False)` can be true. `merged=False` actually requires special handling in this implementation.
- `Settings.set([...], None)` no longer calls `HierarchicalChainMap.del_by_path()`. I'm not sure what it was deleting before, I find it very difficult to keep trees in my head, but setting the path to `None` and deleting all subchildren makes sense in this implementation.

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -542,101 +542,232 @@ class DuplicateFolderPaths(InvalidSettings):
         )
 
 
-class HierarchicalChainMap(ChainMap):
-    def deep_dict(self):
-        def deep_dict_inner(root):
-            return {
-                key: deep_dict_inner(self.__class__._get_next(key, root))
-                if isinstance(value, dict)
-                else value
-                for key, value in root.items()
-            }
+_CHAINMAP_SEP = "\x1f"
 
-        return deep_dict_inner(self)
+
+class HierarchicalChainMap(object):
+    """
+    Stores a bunch of nested dictionaries in chain map, allowing queries of nested values
+    work on lower directories. For example:
+
+    Example:
+        >>> example_dict = {"a": "a", "b": {"c": "c"}}
+        >>> hcm = HierarchicalChainMap({"b": {"d": "d"}}, example_dict)
+        >>> cm = ChainMap({"b": {"d": "d"}}, example_dict)
+        >>> cm["b"]["d"]
+        'd'
+        >>> cm["b"]["c"]
+        Traceback (most recent call last):
+            ...
+        KeyError: 'c'
+        >>> hcm.get_by_path(["b", "d"])
+        'd'
+        >>> hcm.get_by_path(["b", "c"])
+        'c'
+    """
+
+    @staticmethod
+    def _unflatten(kv_pairs, prefix=""):
+        """
+        :type kv_pairs: Iterable[Tuple[str, any]]
+        """
+        result = dict()
+        for key, value in kv_pairs:
+            if not key.startswith(prefix):
+                continue
+            subkeys = key[len(prefix) :].split(_CHAINMAP_SEP)
+            current = result
+            for subkey in subkeys[:-1]:
+                if subkey not in current:
+                    current[subkey] = {}
+                current = current[subkey]
+            current[subkeys[-1]] = value
+
+        return result
+
+    def __init__(self, *maps):
+        self._chainmap = ChainMap(*map(self._flatten, maps))
+
+    @staticmethod
+    def from_layers(*layers):
+        result = HierarchicalChainMap()
+        result._chainmap.maps = layers
+        return result
+
+    @staticmethod
+    def _flatten(d, parent_key=""):
+        if d is None:
+            return {}
+
+        items = []
+        for k, v in d.items():
+            new_key = parent_key + _CHAINMAP_SEP + k if parent_key else k
+            if v and isinstance(v, dict):
+                items.extend(HierarchicalChainMap._flatten(v, new_key).items())
+            else:
+                items.append((new_key, v))
+        return dict(items)
+
+    def deep_dict(self):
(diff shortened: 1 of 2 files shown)


foosel:
> I have a lot of trouble understanding the settings implementation, and I feel like it can be simplified.

I wrote it, I'm still on the same page as you (the implementation was written in large parts in 2016 and I have forgotten a ton about it), so I'm all for this discussion!



> This implementation stores the settings as a chainmap of full paths rather than as a set of nested chainmaps.

I really like that idea.



>     * `HierarchicalChainMap.get(asdict=True)` is eliminated. It behaves as if this is always true.

This turns this into something that needs to target devel/2.0 then. Backwards incompatible change. I agree that it makes sense to do this however since the current implementation is just needlessly complicated. We'd need a migration guide for plugin authors however (their settings use this setting implementation under the hood). 



>     * I'd suggest that `HierarchicalChainMap.get(merged=True)` be eliminated and behave as if it is always true, as it is confusing how `settings.effective['foo']['bar'] != settings.get(['foo', 'bar'], merged=False)` can be true. `merged=False` actually requires special handling in this implementation.

I'm wondering if that might cause issues in some workflows. IIRC this is primarily used to get only the subset stored in config.yaml in some scenarios (e.g. some bundled plugins), so we'd need to create a way to still access that. `only_local` or something like that comes to mind.

Ideally we'd screen the bundled plugins and raise a bit of ruckus about this in the plugin author community as well in the hopes that people who'd be negatively effected by this/require it for specific workflows would speak up.

>     * `Settings.set([...], None)` no longer calls `HierarchicalChainMap.del_by_path()`. I'm not sure what it was deleting before, I find it very difficult to keep trees in my head, but setting the path to `None` and deleting all subchildren makes sense in this implementation.

Can't remember what was going on there either. The problem with all of this is that it was implemented in 2016 and then added to over the years. So now you have this organically grown ... thing that is harder than needed to understand the inner workings of.

tldr; I'm all for some refactoring and making stuff easier, we just need to see how we can do that without making it too hard on the plugin devs relying on the settings implementation behaving as it does now. 

Also: if possible it would be lovely if you could file the refactoring of the HierarchicalChainmap API to include `top_layer` and such properties, and the settings manager making use of that as a separate PR. This one here might take a while, but I really like this refactoring, makes things way cleaner already, would love to see that in 1.8.0.

foosel:
@flaviut have you had a chance to take a look at the splitting of this PR into a 1.8.0 and 2.0.0 subset as mentioned above?

flaviut:
I have not. I'll review this issue later today, are there any rough deadlines I should be aware of?

foosel:
Nope, I was really just wondering if this was still on your mind 😅

flaviut:
Hey! So I finally got a chance to take a look at this, and I have a 80%-finished version here: https://github.com/flaviut/op-hierarchical-chainmap

Basically, I'd like better performance now and would prefer not to wait until 2.0. I wrote this in a new project because python packaging, especially for native extensions, is very difficult. I'd be happy to transfer it across github namespaces.

The native version of `deep_dict` is 100x faster on the default Octporint config than the python version on x64, I haven't tested ARM yet.

I'd appreciate feedback on this and if it's something you're interested in. After that, I'd be happy to go through and polish it up to get it included in Octoprint!

foosel:
First of all, this is discussion relating to both this PR as well as #4338, since I understand the latter to be meant as a temporary speedup only, until this one can make it into OctoPrint proper.

I have enormous reservations towards introducing a native extension that is *very* specific to the project, would need a full build pipeline established, preferably with wheels for all major expected target platforms to not require compilation on every single instance out there, and would require all that effort only as a workaround. So instead of that, let's see how we can get this into 1.8.0 after all.

I've looked through the code and also the docs, and then also the tests, and have so far been unable to determine the `asdict` behaviour change you quoted in actual code. There never was `HierarchicalChainMap.get` nor an `asdict` parameter there. Where that comes into play is `Settings.get`, and as far as I can see (and also can confirm in the test suite) you have not changed any behaviour there at all. The only changes to actual behaviour on the settings that I have seen so far is one change related to how empty dictionaries under a key are now handled in the top layer - old behaviour was keeping an empty dictionary, which was more of a bug than a feature, new behaviour is cleaning up, and I'm pretty sure that was what the `del_by_path` call in `Settings.set` was supposed to accomplish but failed to achieve. Apart from one failing unit test that still needed adjustment I don't see an issue with this minimal behaviour change (that was undocumented anyhow).

> * I'd suggest that `HierarchicalChainMap.get(merged=True)` be eliminated and behave as if it is always true, as it is confusing how `settings.effective['foo']['bar'] != settings.get(['foo', 'bar'], merged=False)` can be true. `merged=False` actually requires special handling in this implementation.

So THIS would be the only backwards incompatible change (that again, I'm all in favour for doing longterm, just not in 1.x). So how about we throw a deprecation warning in here if merged is False, directing people to make use of a to be created way to access the top layer (which this was originally meant to do, in case you need to know if something is actually in config.yaml or not), and then throwing that out in 2.0 but leaving it in for now.

PR #4280, opened 2021-10-26
Websocket: Allow client to selectively subscribe to messages

crysxd:
This PR adds a `subscribe` message to the web socket connection which can be sent by the client in order to selectively subscribe to only parts of what the web socket currently sends. Reducing the websocket traffic is important for mobile applications in order to reduce bandwidth but can also help clients with low performance. 

This change is **not breaking** as existing clients will see no changes in behavior.

The `connected` message now includes an additional field `supports_selective_subscription` (always `True`) which can be used by a client to determine wether the server supports the new protocol extension. Clients can now include a `uses_selective_subscription` flag in the `auth` message (defaults to `False` for compatibility with old clients) which will tell the server to not send any messages before the client does not subscribe to them.

After requesting  the selective subscription with the `auth` message, clients can send a `subscribe` message to specify in which messages they are interested in. The `subscribe` message can be sent multiple times to update the subscription. If a new `subscribe` message is received, old values are discarded and not merged with the new message.

The subscribe message can contain following fields:

- `current` is a boolean value to indicate wether the current message shall be sent. If the client was not subscribed to the current message before, a `history` message will be sent. This flag defaults to `False`
- `logsFilterRegex` is a string that can be used to filter terminal logs included in the `history` and `current` message. If the field is not provided, no filters are applied
- `plugins` can be either `True` to subscribe to all plugin messages or `False` to not receive any plugin messages. It can also be an array of strings where each string is a plugin name. The array indicates a list of plugins the client wants to receive messages from. Plugins listed here do not need to be installed. The field defaults to `False`
- `events` can be either `True` to subscribe to all events or `False` to not receive any event messages. It can also be an array of strings where each string is an event name. The array indicates a list of events the client wants to receive messages for. The field defaults to `False`

Following is an example flow:

```
OctoPrint: {"connected": {"supports_selective_subscription": true, ...}
Client: {"auth": "...", "uses_selective_subscription": true}
Client: {"subscribe": { "current": true, "plugins": ["PSUControlPlugin"], events: true, logsRegexFilter: "^Recv:.*M290.*$"}
OctoPrint: ...
Client: {"subscribe": { "current": true, "plugins": false, events: ["PrintCompleted"] }
OctoPrint: ...
```

I tested the change by validating the web interface as a "legacy" client is still fully functional and receives all messages, confirming the non-breaking nature of the changes. The filter options were tested by adopting OctoApp to use the `subscribe` method in different configurations, using timers to subscribe and unsubscribe at intervals.

See discussion on [Discord](https://discord.com/channels/704958479194128507/705047010641838211/900636949264625704)

/

/

#### Further notes
/


foosel:
I've refactored this a bit and added docs and also a filter for messages. Can you take a look @crysxd if that still matches your idea? I've put it through some testing from my end with a quickly whipped up client, but it would also be nice if you could take a look with your own testing.

crysxd:
Looks great! Much more elegant code :D 

~The only concern I have is that you removed the flag from the `auth`  message...but after thinking about it a bit, the client can also send the `subscribe` message before sending the `auth` message to ensure it never receives a `history` message. Or am I wrong with this assertion?~ Read to the end before complaining

The only thing I'd miss in the documentation is a reference to the `supports_selective_subscription` flag in the `connected` message as this might be an easier way to determine support compared to checking the OctoPrint version. Not sure if you planned on remove this flag? If not, I can add a reference to the documentation and resolve conflicts tomorrow.

foosel:
Hah, the flag went right past me, I was too focused on the other parts. Rename it to `subscriptions_supported` and document it and we have a deal ^^

crysxd:
I think we don't need this flag. There is no penalty for sending the subscription message to older instances. You will get more than you asked for....but there should be no functional impacts. If this really would be a problem for the application for some reason, a version check is always possible.

Still have to do the testing tonight or tomorrow morning, decided to have a work-free weekend 😃

crysxd:
@foosel Everything is tested and works well. I removed the `supports_selective_subscription` flag.

While testing, one thing I noticed is the inverted logic of `filter_logs` and `filter_messages` compared to `plugins` and `events`. For the first two `True` means you get nothing, for the latter two `False` means you get nothing. This is a bit confusing on first glance. We could invert the logic of `plugins` and `events` to be `filter_plugins` and `filter_events` to have a consistent logic. And that point, I'm wondering wether `subscribe` is the best name for the message itself or if `filter` would be better. We could also invert `filter_logs` and `filter_messages` and keep `subscribe`.

What do you think? Functionally, this is ready to be merged.

foosel:
That a really good point. I like the subscribe point of view, so I've simply now reversed the logs and message logic. `logs: true` means "give me everything", `logs: false` means "give me nothing", and `logs: <regex>` means "give me only that which matches". So now it's basically the same as for events and plugins. Also adjusted the docs.

crysxd:
🥳 Will do a simple integration on the weekend and do some more testing. Thanks for your support!

Issue #4299, opened 2021-11-03
PrintFailed event is fired rapidly


fritzw:
### The problem

1. Use Klipper
2. Turn on the printer, but do not issue a FIRMWARE_RESET
3. Try to start a print
4. Check the terminal to see OctoPrint sending many G-Code lines and Klipper responding with the same error message each time
5. Observe that the PrintFailed event is fired many times. Probably once for each G-Code line sent (for example by using the OctoPrint-IFTTT plugin to send a push message to your phone).

**Expected Behavior**: PrintFailed is only fired _once_ for each print when it fails, not dozens of times.

Example terminal output:
```
Send: G1 E-5 F2000
Recv: // Lost communication with MCU 'mcu'
Recv: // Once the underlying issue is corrected, use the
Recv: // "FIRMWARE_RESTART" command to reset the firmware, reload the
Recv: // config, and restart the host software.
Recv: // Printer is shutdown
Recv: !! Lost communication with MCU 'mcu'
Recv: ok
Send: G1 X-10 F8000
Recv: // Lost communication with MCU 'mcu'
Recv: // Once the underlying issue is corrected, use the
Recv: // "FIRMWARE_RESTART" command to reset the firmware, reload the
Recv: // config, and restart the host software.
Recv: // Printer is shutdown
Recv: !! Lost communication with MCU 'mcu'
Recv: ok
Send: G1 Y200
Recv: // Lost communication with MCU 'mcu'
Recv: // Once the underlying issue is corrected, use the
Recv: // "FIRMWARE_RESTART" command to reset the firmware, reload the
Recv: // config, and restart the host software.
Recv: // Printer is shutdown
Recv: !! Lost communication with MCU 'mcu'
Recv: ok
Send: M106 S0
Recv: // Lost communication with MCU 'mcu'
Recv: // Once the underlying issue is corrected, use the
Recv: // "FIRMWARE_RESTART" command to reset the firmware, reload the
Recv: // config, and restart the host software.
Recv: // Printer is shutdown
Recv: !! Lost communication with MCU 'mcu'
Recv: ok
Send: G92 E0
```

### Did the issue persist even in safe mode?

I cannot test this issue in safe mode (state why below)

### If you could not test in safe mode, please state why

Requires plugins

### Version of OctoPrint

1.7.2

### Operating system running OctoPrint

octopi 0.18.0

### Printer model & used firmware incl. version

[...]

PR #4324, opened 2021-11-27
✨ virtual: Add printer lock feature

cp2004:
For testing behaviour when the printer is locked. Marlin feature, enabled with `PASSWORD_FEATURE`.

Can be enabled on startup, where no communication is possible until unlock. In Marlin, it accepts a 1-9 digit password.

Also added/fixed OKs not being sent on Klipper MCU errors. Possible resolution to the hanging experienced debugging https://github.com/OctoPrint/OctoPrint/issues/4299#issuecomment-978112886

Tested locally, Windows 10/Python 3.9.


foosel:
Ahh, the missing ok might indeed have been the culprit there, thanks for spotting that.

PR #4336, opened 2021-12-08
Check if paused for printer.is_ready

kohend:
Checks for paused print in the is_ready method.

With the BetterHeaterTimeout plugin, it can be tested by anything that checks the is_ready API while the printer is ready and when it's paused.

Found it when using BetterHeaterTimeout plugin

image

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -921,7 +921,8 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
     def is_ready(self, *args, **kwargs):
         return (
             self.is_operational()
-            and not self.is_printing()
+            and not self._comm.isBusy()
+            # isBusy is true when paused
             and not self._comm.isStreaming()
         )
 
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -157,6 +157,7 @@ date of first contribution):
   * [Christian Würthner](https://github.com/crysxd)
   * [Maciej Urbański](https://github.com/rooterkyberian)
   * [Adam Wolf](https://github.com/adamwolf)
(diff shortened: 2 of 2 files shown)


src/octoprint/printer/standard.py
```diff
@@ -927,6 +927,7 @@ def is_ready(self, *args, **kwargs):
         return (
             self.is_operational()
             and not self.is_printing()
+            and not self.is_paused()
             and not self._comm.isStreaming()
         )
```
foosel, line 932: Good find! However, in order to not duplicate logic here, I'd rather go with:

```suggestion
        return (
            self.is_operational()
            and not self._comm.isBusy()
            and not self._comm.isStreaming()
        )
```

Do you want to do the honors or shall I?

foosel (changes requested):

src/octoprint/printer/standard.py
kohend: I did, and added a comment since comm being busy is not obvious that it includes paused.

src/octoprint/printer/standard.py
```diff
@@ -927,6 +927,8 @@ def is_ready(self, *args, **kwargs):
         return (
             self.is_operational()
             and not self.is_printing()
+            and not self._comm.isBusy()
```
foosel, line 930: If you leave `is_printing` in there this is redundant. Just `self._comm.isBusy` covers all printing and paused states.

foosel (changes requested):

src/octoprint/printer/standard.py
kohend: OK, removed, and fixed the merge conflict, I was using a stale fork.

foosel (approved):

PR #4337, opened 2021-12-11
Create new plugin version list endpoint

crysxd:
This PR introduces a new permission `LIST_PLUGINS` and a new endpoint `/plugin/pluginmanager/plugins/versions` which returns a compact list of all enabled plugins and their versions. This can be used by 3rd party systems that want to provide similar functionality to the web interface or integrate with plugins via their API. The new endpoint requires the `LIST_PLUGINS` permission which is grated by default to read only users, operators and admins. Guests do not have access to this endpoint (403). 

At the moment, only users holding the `MANAGE_PLUGIN` permission have the ability to see which plugins are installed. Other users can already "probe" for installed plugins by e.g. looking at the settings or trying to request API endpoints provided by plugins. This means the information is already available to malicious parties and the new endpoint does not expose additional information.

I opted to return a very compact data set as this is all that an external client would need. Additional information like e.g. safe mode victim, pending disabled etc. is not included as this data is only needed for the intend of managing the plugins - thus should be restricted to users holding the matching permission.

The endpoint returns a JSON object:

```
{
    "PrintTimeGenius": "2.2.8",
    "action_command_prompt": null,
    "announcements": null,
    "appkeys": null,
    "backup": null,
    "corewizard": null,
    "discovery": null,
    "errortracking": null,
    "eventmanager": null,
    "file_check": "2021.2.23",
    "firmware_check": "2021.10.11",
    "gcodeviewer": null,
    "logging": null,
    "resource_monitor": "0.3.5",
    "softwareupdate": null,
    "tracking": null,
    "virtual_printer": null
}
```

Postman & web interface

[Discord discussion](https://discord.com/channels/704958479194128507/705047010641838211/914807730982772777)

I know we said we wouldn't need to filter for enabled plugins only, but I wanted to make the response as simple as possible. Now all listed plugins are available to be used. Hope that's ok, happy to change the structure if requested!

/ 

#### 
Screenshots (if appropriate)
Screenshot 2021-12-11 at 15 09 59

#### Further notes
/

### DIFF
--- a/src/octoprint/plugins/pluginmanager/__init__.py
+++ b/src/octoprint/plugins/pluginmanager/__init__.py
@@ -27,7 +27,7 @@ from past.builtins import basestring
 
 import octoprint.plugin
 import octoprint.plugin.core
-from octoprint.access import ADMIN_GROUP
+from octoprint.access import ADMIN_GROUP, READONLY_GROUP, USER_GROUP
 from octoprint.access.permissions import Permissions
 from octoprint.events import Events
 from octoprint.server import safe_mode
@@ -226,6 +226,13 @@ class PluginManagerPlugin(
 
     def get_additional_permissions(self):
         return [
+            {
+                "key": "LIST",
+                "name": "List plugins",
+                "description": gettext("Allows to list installed plugins."),
+                "default_groups": [READONLY_GROUP, USER_GROUP, ADMIN_GROUP],
+                "roles": ["manage"],
+            },
             {
                 "key": "MANAGE",
                 "name": "Manage plugins",
@@ -461,6 +468,13 @@ class PluginManagerPlugin(
             "online": self._connectivity_checker.online,
         }
 
+    def _plugin_versions_response(self):
+        plugins = dict()
+        for plugin in self._get_plugins():
+            if plugin["enabled"] is True:
+                plugins[plugin["key"]] = plugin["version"]
+        return plugins
+
     @octoprint.plugin.BlueprintPlugin.route("/plugins")
     @Permissions.PLUGIN_PLUGINMANAGER_MANAGE.require(403)
     def retrieve_plugins(self):
@@ -485,6 +499,38 @@ class PluginManagerPlugin(
             hash_update(repr(self._notices))
             hash_update(repr(safe_mode))
             hash_update(repr(self._connectivity_checker.online))
+            hash_update(repr(_DATA_FORMAT_VERSION))
+            return hash.hexdigest()
+
+        def condition():
+            return check_etag(etag())
+
+        return with_revalidation_checking(
+            etag_factory=lambda *args, **kwargs: etag(),
+            condition=lambda *args, **kwargs: condition(),
+            unless=lambda: refresh,
+        )(view)()
+
+    @octoprint.plugin.BlueprintPlugin.route("/plugins/versions")
+    @Permissions.PLUGIN_PLUGINMANAGER_LIST.require(403)
+    def retrieve_plugin_list(self):
+        refresh = request.values.get("refresh", "false") in valid_boolean_trues
+        if refresh or not self._is_notices_cache_valid():
+            self._notices_available = self._refresh_notices()
+
+        def view():
+            return jsonify(**self._plugin_versions_response())
+
+        def etag():
+            import hashlib
+
+            hash = hashlib.sha1()
+
+            def hash_update(value):
+                value = value.encode("utf-8")
+                hash.update(value)
+
+            hash_update(repr(self._plugin_versions_response()))
             hash_update(repr(_DATA_FORMAT_VERSION))
             return hash.hexdigest()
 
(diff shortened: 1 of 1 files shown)


src/octoprint/plugins/pluginmanager/__init__.py
```diff
@@ -497,6 +511,38 @@ def condition():
...
+        def etag():
+            import hashlib
+
+            hash = hashlib.sha1()
+
+            def hash_update(value):
+                value = value.encode("utf-8")
+                hash.update(value)
+
+            hash_update(repr(self._plugin_versions_response()))
```
foosel, line 533: This will behave non-deterministically depending on the used python versions, because the order of items returned from a dict is not guaranteed to match insertion order in Python < 3.7.

Better:

``` python
hash_update(repr(list(map(lambda k, v: "{}={}".format(k, v), self._plugins_versions_response()))))
```

Alternatively you could also use `OrderedDict` for building the data structure.

foosel, line 533: This will behave non-deterministically depending on the used python versions, because the order of items returned from a dict is not guaranteed to match insertion order in Python < 3.7.

Better:

``` python
hash_update(repr(list(map(lambda k, v: "{}={}".format(k, v), self._plugins_versions_response()))))
```

Alternatively you could also use `OrderedDict` for building the data structure.

src/octoprint/plugins/pluginmanager/__init__.py
```diff
@@ -461,6 +468,13 @@ def _plugin_response(self):
             "online": self._connectivity_checker.online,
         }
 
+    def _plugin_versions_response(self):
+        plugins = dict()
+        for plugin in self._get_plugins():
+            if plugin["enabled"] is True:
+                plugins[plugin["key"]] = plugin["version"]
+        return plugins
+
```
foosel, line 477:
```suggestion
    def _get_plugins_with_versions(self, only_enabled=True):
        return dict((p["key"], p["version"]) for p in self._get_plugins() if p["enabled"] or not only_enabled)
```

src/octoprint/plugins/pluginmanager/__init__.py
```diff
@@ -497,6 +511,38 @@ def condition():
             unless=lambda: refresh,
         )(view)()
 
+    @octoprint.plugin.BlueprintPlugin.route("/plugins/versions")
+    @Permissions.PLUGIN_PLUGINMANAGER_LIST.require(403)
+    def retrieve_plugin_list(self):
+        refresh = request.values.get("refresh", "false") in valid_boolean_trues
+        if refresh or not self._is_notices_cache_valid():
+            self._notices_available = self._refresh_notices()
```
foosel, line 519: I think this doesn't belong here.

src/octoprint/plugins/pluginmanager/__init__.py
```diff
@@ -497,6 +511,38 @@ def condition():
...
+                hash.update(value)
+
+            hash_update(repr(self._plugin_versions_response()))
+            hash_update(repr(_DATA_FORMAT_VERSION))
+            return hash.hexdigest()
+
+        def condition():
+            return check_etag(etag())
+
+        return with_revalidation_checking(
```
foosel, line 540: Considering that the full response is required in order to determine the ETag needed for revalidation checking, I wouldn't bother with caching here since it simply doesn't add anything in terms of saving resources.

If I need to fetch all enabled plugins & versions to figure out if they changed in order to respond with a `Not Modified` header, there's not much point in caching to begin with.

I'd rather suggest to get rid of the additional response generating method defined above altogether and inlining that here.

foosel (changes requested):
Found a couple of points that need to be taken care of.

Quick heads-up, both a booster shot and a christmas break are coming up on my end, so it might take a bit until this gets merged depending on when changes are made.

crysxd:
Will work through the changes today or tomorrow! Thanks for taking the time to check this out 
🙃 No time pressure here...so if we won't merge it before the new year that's totally fine...even it won't make it in 1.8

src/octoprint/plugins/pluginmanager/__init__.py
crysxd: So basically we can get rid of the entire ETag/caching and just return a simple json response?

foosel: IMHO yes. I mean, we don't really gain any performance from caching here, and in worst case (cache is stale) will actually have to do the work needed to preparing the response twice (once for the ETag comparison and once for the actual response generation).

crysxd: Totally agree...I basically just copied it from the other endpoints but guess the situation there is a bit different. Will remove it, also passively resolves some of the other points!

crysxd:
Removed Etag as discussed and inlined the data generation

crysxd:
Seems like npm has a bad day...I tried building it twice but installing eslint fails with 503

Will trigger it again later today

foosel:
Hm, GitHub is still showing me the old code... did something go wrong with the force pushes there?

crysxd:
Seems like I did something wrong when "restarting" the build this morning....sorry for the extra round :(

foosel (approved):

PR #4338, opened 2021-12-11
Use C++ HierarchicalChainMap

flaviut:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This replaces the HierarchicalChainMap with a version written in C++. That means that Octoprint's settings functionality gets faster throughout, without any breaking changes.

On my machine, this improves the response time of /api/settings from 1.6s to .4s.

- unit tests on the implementation of the ChainMap
- unit tests comparing the python HierarchicalChainMap to the C++
- manual testing by loading into octoprint
- running the above under valgrind to test for memory safety issues (you must use `./configure --with-valgrind ...` when building python)

https://github.com/OctoPrint/OctoPrint/pull/4236 is a re-write of this functionally in a simpler & more logical way, but isn't compatible with the existing code in certain edge cases.

before
after

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,7 @@ INSTALL_REQUIRES_PYTHON3 = [
     "colorlog>=5.0.1,<6",
     # vendor bundled dependencies
     "unidecode",  # dependency of awesome-slugify, leave w/o range to not cause dependency hell w/ Octolapse deps
+    "op_hierarchical_chainmap @ git+https://github.com/flaviut/op-hierarchical-chainmap.git@1e61f5e941552838393525032f59de7085936485#egg=op_hierarchical_chainmap",
 ]
 
 # OSX specific requirements
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -705,7 +705,15 @@ class Settings(object):
 
         assert isinstance(default_settings, dict)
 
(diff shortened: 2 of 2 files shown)


foosel:
As mentioned on Discord, that will most likely not be reviewed until after my holidays break. IF I go for a native implementation of Chainmap (I will admit that right now I'm very cautious about this since it adds complexity and potential maintenance struggles) it would have to be available on PyPI however, and properly maintained - are you able to commit to this or would this be more a case of "here, take this code and maintain it yourself"?

flaviut:
> are you able to commit to this or would this be more a case of "here, take this code and maintain it yourself"?

I'd be happy to commit to maintaining this, with the understanding that the end goal here is to do something that's a lot less work (like my previous PR) and eliminate this code.

foosel:
While this is a great implementation feat, I'm not going to go for this. A project specific native extension that would require a build pipeline to compile native wheels for the most common deployment targets, would need to be pushed to pypi even though VERY project specific just to be able to install without compilation across instances out there, and all of that only as a temporary workaround - that's not a good use of both our time I fear. So instead let's see that we merge your end game changes from #4236 instead, and upon further investigation, it seems like that might be possible for 1.8.0 after all, see my comment there.

PR #4339, opened 2021-12-11
Use pickle for announcements cache

flaviut:
Parsing this XML file is very slow, so there's no reason to do it more than we have to.

See [4480-2021-12-11T13:23:48-05:00.json](https://github.com/OctoPrint/OctoPrint/files/7697670/4480.zip), MainThread -> parse. Use https://www.speedscope.app/ to view the file.

On my system, this reduces the time for /plugin/announcements/channels from 250ms CPU time to 50ms.

The rough improvement in this specific function call is from 8% to 0.32% of total runtime. (see [6529-2021-12-11T14:49:20-05:00.json](https://github.com/OctoPrint/OctoPrint/files/7697676/6529.zip) for after)

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Manually tested with existing cache files. If existing XML files are in place, the pickle parser will fail & give up.


foosel:
Using pickled data for on-disk caching/session persistence always makes me very very worried about potential abuse vectors opened up by this. Just imagine someone, e.g. through abusing a security hole in a third party plugin, being able to replace the cache data here with a prepared payload that opens a reverse shell ([example](https://davidhamann.de/2020/04/05/exploiting-python-pickle/)).

See also the big warning on the top of the python docs page about `pickle`:

image

Data here really might have been tampered with.

The other PR by you using pickle immediately pickled and unpickled something in memory again, so I was okay with it. I would be fine with this change too if you'd use json, or if you used a [custom hardened unpickler](https://docs.python.org/3/library/pickle.html#restricting-globals).

Side note, I thought webassets was working around this with a custom "safe" pickle loader, at least they called it `safe_unpickle`, but upon further investigation I just had to realize that they in fact don't really make anything safe here from a security point of view but rather an error handling one, and I'll need to harden the code there in that respect by implementing a custom unpickler in OctoPrint anyhow.

foosel (changes requested):
See my comment above

foosel:
As discussed on Discord, custom unpickler will never be as safe as desired, so JSON serialization it should be. I've disabled the bundle part caching on webassets, as it was only used during initial bundling anyhow, and on refreshing during development, so that stuff should be fine now, and for this cache here as discussed we should go with JSON. You might want to look into the two new helpers `octoprint.util.[de]serialize` that I've pushed in 008df45

flaviut:
So this is very WIP, since the following occurs:

```
-           'updated_parsed': time.struct_time(tm_year=2010, tm_mon=9, tm_mday=6, tm_hour=0, tm_min=1, tm_sec=0, tm_wday=0, tm_yday=249, tm_isdst=0)},
+           'updated_parsed': [2010, 9, 6, 0, 1, 0, 0, 249, 0]},
```

The problem here is that `time.struct_time` is secretly a tuple, not a class, so json never bothers calling the default method to serialize it in a more special way. Anyway, potential solutions:

- recurse through the object pre-serialization & replace `struct_time` with something else.
- use an external library like `orjson`, which promises to call `default` more often.
- use an external library like `quickle`, which is basically pickle except it doesn't work on arbitrary types.

I'd appreciate your input on what you'd prefer here!

foosel:
Since I'd prefer not having to pull in extra dependencies just for this, I'd vote for the first option for now.

foosel:
@flaviut just a heads-up, I'll look into wrapping up 1.8.0 soon

flaviut:
Appreciate the heads up. I'll put this on the weekend's to do list.

tests/util/test_comprehensive_json.py
```diff
@@ -0,0 +1,39 @@
...
+from frozendict import frozendict
+
+from octoprint.util import comprehensive_json
+
+
+def test_bytes_roundtrip():
+    val = b"\x00\x01\x02\x03"
+    res = comprehensive_json.dumps(val)
+    assert res == '"\\u0000\\u0001\\u0002\\u0003"'
+    assert comprehensive_json.loads(res) == "\x00\x01\x02\x03"
```
flaviut, line 15: I'm not super happy about `bytes` not round-tripping properly, but I've refrained from changing this in case there are compatibility concerns elsewhere.

src/octoprint/util/comprehensive_json.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @classmethod
+    def remove_decoder(cls, classname):
+        try:
+            del cls.decoders[classname]
+        except KeyError:
+            pass
+
+    @classmethod
+    def encode(cls, val):
```
flaviut, line 73: this may be a bit overkill, but I've basically replicated the way that the json encoder walks over the value tree here, with the exception of the cycle checking that it does.

src/octoprint/util/json/__init__.py
```diff
@@ -1,46 +1,13 @@
...
-    def encode(cls, obj):
-        for type, encoder in cls.encoders.items():
-            if isinstance(obj, type):
-                return encoder(obj)
-        raise TypeError
-
-
-JsonEncoding.add_encoder(frozendict, lambda obj: dict(obj))
-JsonEncoding.add_encoder(bytes, lambda obj: to_unicode(obj))
+dump = deprecated(
```
flaviut, line 9: is this module part of the public octoprint API?

src/octoprint/util/json/encoding.py
```diff
@@ -0,0 +1,66 @@
...
+from collections import OrderedDict
+from typing import Any, Callable, Dict
+
+from frozendict import frozendict
+
+from octoprint.util import to_unicode
+
+
+class JsonEncoding:
+    # actually OrderedDict, but typing annotation for that is only available from
```
flaviut, line 14: didn't realize it--I thought python only used patch releases for bug fixes, not new stuff.

How about this instead?

```
try:  # py3.7.{0,1} support
  from typing import OrderedDict
except ImportError:
  OrderedDict = Dict
```

src/octoprint/util/json/serializing.py
```diff
@@ -0,0 +1,156 @@
...
+from typing import Any, Callable, Dict, List
+
+from frozendict import frozendict
+
+from octoprint.util import to_bytes, to_unicode
+
+from .encoding import JsonEncoding
+
+
+class SerializableJsonEncoding(JsonEncoding):
```
flaviut, line 17: I like this separation. I was filling iffy before about how the previous interface filled too many goals in one place, but this is a lot better.

tests/util/test_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+def test_encoding_dumps_typeerror():
+    with pytest.raises(TypeError):
+        json.encoding.dumps(SomeClass())
+
+
+@pytest.mark.parametrize(
+    "val",
+    [
+        pytest.param({"foo": "bar"}, id="dict"),
```
flaviut, line 56: I also like this way of doing the tests! Didn't realize it was an option.

foosel:
While reviewing this I realized that the new serialization focused encoding class was now being used everywhere, the old `octoprint.util.json` stuff was still left but unused and also some backwards incompatibility managed to creep in (`dump` vs `dumps`). That made me think about whether we weren't mixing up concerns here and indeed we were, so I've refactored things.

Everything is now under `octoprint.util.json` but split between encoding (= we want to transfer stuff to json so it can be pushed out on an API) and serialization (= we need a way to save a data structure to disk somewhere and want to be able to get it back as it was). The former is the same as before, the latter is what you added. I adjusted the usages accordingly and also added a deprecation wrapper for the `dump` situation. Finally, some more tests where thrown in for good measure.

Could you take a look at it and see if I missed something in the process of this?

foosel:
Damn, you were faster than I was in writing my comment XD

src/octoprint/util/json/__init__.py
foosel: Not per se documented, but as with everything under util, `dump` might be in use by third party plugins and I prefer to at least give them a warning before just throwing stuff out. Was still even used in core OctoPrint itself btw, also fixed that.

src/octoprint/util/json/encoding.py
foosel: Good idea, going to do that

tests/util/test_json.py
foosel: I only recently learned about that approach (so far have been using `ddt` for this kinda stuff), but find it really nice. Decided to try it here ^^

tests/util/test_comprehensive_json.py
foosel: It now roundtrips properly on the serializing implementation 🥳

src/octoprint/util/comprehensive_json.py
foosel: Felt like overkill, but I couldn't find a better approach either.

foosel (approved):
From my perspective this should be fine now, some extra set of eyes are welcome however.

flaviut:
Yup. looks good! I fixed a bug in the bytes encoding, but I'm happy with it as it is right now.

foosel:
Thank you, that was a lovely collaboration experience 😊

flaviut:
It was! Thank you for working on this with me!

PR #4349, opened 2021-12-21
Improve timelapse documentation

crysxd:
Small additions to the Timelapse API docs to ensure everything is matching the implementation. DELETE and POST requests returns the response body of the GET request with updated information and the `unrendered` flag is also applying to them. Also the `save` flag used to make the config persist across restarts was not documented.

Checking the code :)

/

/

[Timelapse — OctoPrint 2.0.0.dev1534+g6e47596e3.dirty documentation.pdf](https://github.com/OctoPrint/OctoPrint/files/7755814/Timelapse.OctoPrint.2.0.0.dev1534%2Bg6e47596e3.dirty.documentation.pdf)

#### Further notes
/


docs/api/timelapse.rst
```diff
@@ -27,12 +27,17 @@ Retrieve a list of timelapses and the current config
...
 
-   Delete the timelapse ``filename``.
+   Delete the timelapse ``name``.
+   
+   Returns a :ref:`timelapse list <sec-api-timelapse-datamodel-list>` in the
+   response body with the updated information.
 
    Requires the ``TIMELAPSE_DELETE`` permission.
 
+   :param unrendered: If provided and true, also include unrendered timelapses
```
cp2004, line 39:
```suggestion
   :param unrendered: If provided and true, also include unrendered timelapses in the response
```

docs/api/timelapse.rst
```diff
@@ -56,21 +61,33 @@ Delete an unrendered timelapse
 
    Delete the unrendered timelapse ``name``.
 
+   Returns a :ref:`timelapse list <sec-api-timelapse-datamodel-list>` in the
+   response body.
+
    Requires the ``TIMELAPSE_DELETE`` permission.
 
+   :param unrendered: If provided and true, also include unrendered timelapses
```
cp2004, line 69:
```suggestion
   :param unrendered: If provided and true, also include unrendered timelapses in the response
```

docs/api/timelapse.rst
```diff
@@ -56,21 +61,33 @@ Delete an unrendered timelapse
...
+   configuration to ``true`` in order to make the new configuration persist accross restarts.
 
    The configuration is expected as the request body.
 
+   Returns a :ref:`timelapse list <sec-api-timelapse-datamodel-list>` in the
+   response body.
+
    Requires the ``TIMELAPSE_ADMIN`` permission.
 
+   :param unrendered: If provided and true, also include unrendered timelapses
```
cp2004, line 88:
```suggestion
   :param unrendered: If provided and true, also include unrendered timelapses in the response
```

Not as needed here since it is more obvious but if we're doing the other ones...

cp2004 (commented):
My suggestions are just suggestions...

I felt like since the response is not the subject of the request (eg. deleting or changing config) then it should be more explicit, but then it felt repetitive.

PR #4358, opened 2021-12-31
Adding shutdown action command.

### LINKED ISSUE #4357: [Request] Marlin 2.0.9.3 Host shutdown

Hi,

since Marlin 2.0.9.3, there is an option (HOST_SHUTDOWN_MENU_ITEM in Configuration_adv.h), that allows you to shut down the host through Action Commands. 

It emits: action:shutdown

Could this be supporting in Octoprint?

Support for this in Octoprint itself, ie not through the Action Commands plugin.

I am currently using Action Commands.  It seems like a feature of Marlin deserves more integrated support.

### Additional context

_No response_

JoveToo:
- [ ] Your changes are not possible to do through a plugin and relevant to a large audience (ideally all users of OctoPrint) 
- [x] If your changes are large or otherwise disruptive: You have made sure your changes don't interfere with current development by talking it through with the maintainers, e.g. through a Brainstorming ticket
- [x] Your PR targets OctoPrint's devel branch if it's a completely new feature, or maintenance if it's a bug fix or improvement of
existing functionality for the current stable version (no PRs against master or anything else please)
- [x] Your PR was opened from a custom branch on your repository (no PRs from your version of master, maintenance or devel please), e.g. dev/my_new_feature or fix/my_bugfix
- [x] Your PR only contains relevant changes: no unrelated files, no dead code, ideally only one commit - rebase and squash your PR if necessary!
- [x] Your changes follow the existing coding style If your changes include style sheets: You have modified the .less source files, not the .css files (those are generated with lessc)
- [x] You have tested your changes (please state how!) - ideally you have added unit tests
- [ ] You have run the existing unit tests against your changes and nothing broke
- [x] You have added yourself to the AUTHORS.md file :)

This PR adds support for "shutdown" in the default action commands. This is now supported by an optional menu item in Marlin.

It allows to shut down the OctoPi from the printer LCD menu, which is very handy for headless installations. One less excuse to not neatly shutting down your pi :).

I have tested this with my printer running the 2.0.9.3 Marlin with the option HOST_SHUTDOWN_MENU_ITEM in Configuration_adv.h.

https://github.com/MarlinFirmware/Marlin/pull/22908

Closes #4357 

#### Further notes
Not sure if this can call sarge directly or should use a wrapper.


foosel:
Took the liberty to rebase this on `maintenance`, as it makes more sense that way (this is not a breaking change of any kind, [so `devel` is the wrong target](https://github.com/OctoPrint/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests)). 

Review incoming

src/octoprint/util/comm.py
```diff
@@ -2321,6 +2322,16 @@ def busyIntervalSet():
                         elif action_name == "disconnect":
                             self._log("Disconnecting on request of the printer...")
                             self._callback.on_comm_force_disconnect()
+                        elif action_name == "shutdown":
+                            sarge.run(
+                                settings().get(
+                                    ["server", "commands", "systemShutdownCommand"]
+                                ),
```
foosel, line 2329: This can be unset, in which case this call would error out and cause the whole loop to error out. That's definitely not what we want, needs error handling + check for preconditions.

And  calling system commands from within the comm layer is also something that somehow makes me queasy. It might be time to introduce a manager type of class responsible for handling the restart OctoPrint/restart system/shutdown system commands and rather use that here. That then could also take care of error handling at a central place, and be reused in other places too (system API, software update and plugin manager plugin come to mind).

foosel (changes requested):

src/octoprint/util/comm.py
JoveToo: I can definitely add the error handling.

I may not be the right person to make a manager class although I am willing to try. My python experience is limited.

JoveToo: In any case, I share your queasiness about directly calling sarge and system commands form the inner loop. 

As for input, if you point me to a good class to take as example for the system command wrapper, I will give it a shot. If I have questions, I will come to the #dev-core discord channel.

foosel: Just remember that you might also want to use [octoprint.util.commandline](https://github.com/OctoPrint/OctoPrint/blob/master/src/octoprint/util/commandline.py) for this. Maybe.

JoveToo:
I have added error handling, at least. The try/except might be overkill.

I have tested with a working command, a non-existing setting, an empty setting and a command that fails. Nothing triggers the except clause.

@foosel do you want the command subsystem in this PR or should I start a new one?

JoveToo:
First attempt, please review.

TODO:
- [ ] Integrate this with plugin manager as suggested, [here](https://github.com/OctoPrint/OctoPrint/blob/master/src/octoprint/server/__init__.py#L476)
- [ ] Replace all occurrences of running of system commands with call to this module.

src/octoprint/systemcommands/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    SERVER_RESTART_COMMAND = "serverRestartCommand"
+    SYSTEM_RESTART_COMMAND = "systemRestartCommand"
+    SYSTEM_SHUTDOWN_COMMAND = "systemShutdownCommand"
+
+    def __call(self, command):
+        if not command:
+            return
+
+        try:
+            sarge.run(
```
foosel, line 43: This should also pass `async_=True` so the command only gets triggered and we don't wait for it to complete.

src/octoprint/__init__.py
```diff
@@ -184,6 +184,11 @@ def init_platform(
     if callable(after_environment_detector):
         after_environment_detector(**kwargs)
 
+    try:
+        init_systemcommands()
+    except Exception as ex:
+        raise FatalStartupError("Could not initialize system commands", cause=ex)
+
```
foosel, line 191: This shouldn't get initialized here but rather only as needed. If adjusted as mentioned, there's no need for explicit initialization anyhow.

src/octoprint/__init__.py
```diff
@@ -893,6 +898,12 @@ def init_environment_detector(plugin_manager):
     return EnvironmentDetector(plugin_manager)
 
 
+def init_systemcommands():
+    from octoprint.systemcommands import systemcommands
+
+    return systemcommands(init=True)
+
+
```
foosel, line 906: See above, can be removed.

foosel (changes requested):
Looking good already! There's some implementation issues that need to be tackled, but this is the right way!

It will also obviously need to be used throughout the code base, but it is probably way easier for me to find the places it needs to be used than for you, so leave that to me :) 

It probably should be part of the injected properties provided to plugin implementations as well, so take a look into `octoprint.server.__init__.py` on how to do that (it's just an additional entry in a `dict` really)

src/octoprint/systemcommands/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    return _instance
+
+
+class SystemCommands:
+
+    SERVER_RESTART_COMMAND = "serverRestartCommand"
+    SYSTEM_RESTART_COMMAND = "systemRestartCommand"
+    SYSTEM_SHUTDOWN_COMMAND = "systemShutdownCommand"
+
+    def __call(self, command):
```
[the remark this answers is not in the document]
JoveToo: I googled private and found the __, so I used it.  Protected is better. 

I didn't know about __call__ :)

I've added the return value.

src/octoprint/systemcommands/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+                close_fds=CLOSE_FDS,
+                stdout=sarge.Capture(),
+                stderr=sarge.Capture(),
+                shell=True,
+            )
+        except Exception:
+            logging.getLogger(__name__).exception("Error while executing system command")
+
+    def __getcmd(self, cmd):
+        return settings().get(["server", "commands", cmd])
```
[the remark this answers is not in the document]
JoveToo: It looked pretty but one less stack frame, as you wish :)

JoveToo:
Comments, applied.

foosel (approved):
Thanks! I refactored things a bit to make it usable from other places in the code (and remove some code duplication in the process). I also made the support for this command something that needs to be enabled. The reason for that is that shutting down the system is something that not everyone might want to give their printer access to out of the box (I know I don't), and it opens up an DOS attack vector via manipulated firmware or GCODE (which some people are apparently downloading instead of slicing themselves, and an `M118 //action:shutdown` is easily added), so this is not something that should ship enabled by default but something power users who WANT this kind of functionality have to enable themselves. In fact, after thinking about it a ton the past few days, I'm actually still on the fence about maybe actually extracting it into a plugin due to that. For now though, this gets merged, and the changes to the system command handling this brought with it are going to stay in any case!

Issue #3589, opened 2020-06-02
[Request] "Update When Print Ends" option


KenLucke:
In addition to the "Ignore" and "Update Now" (which does not work while printing, of course) options for software and plugin updates, it would be nice to have a "Update When Print Ends" option that would show up if the dialog is presented during a print.  Clicking that option would wait until the current print job finishes, then do the updates and its restart, after which it would notify the user of the updates installed with an additional dialog.

combs:
I was just coming to open a similar feature request. This would fit nicely with how I, at least, use OctoPrint! Thank you!

foosel:
Just realized this is actually a duplicate of #3320, but this one has seen more upvotes, so I'm keeping it open and will close #3320 instead.

foosel:
Implemented by @jneilliii in #4364 and will be part of 1.8.0

Issue #3585, opened 2020-05-29
[Request] staging plugin updates


mrseeker:
**Is your feature request related to a problem? Please describe.**
I am always frustrated when I have to wait for a print to finish before I can update or install new plugins. This is especially frustrating when the print takes around 20-30h to finish.

**Describe the solution you'd like**
A way to schedule commands (install, update, remove) while prints are in progress, and execute them when the print is finished.

**Describe alternatives you've considered**
Writing it down on a stickynote, and installing them when the print is finished.

GitIssueBot:
This issue has been mentioned on **OctoPrint Community Forum**. There might be relevant details there:

https://community.octoprint.org/t/feature-req-queue-plugin-install/26820/2

jneilliii:
duplicate of #3589 and #3320

Issue #3320, opened 2019-10-27
[Request] Update after print finishes


tjjfvi:
**Is your feature request related to a problem? Please describe.**
The main reason I go to octoprint is to print something or to check on a print. Sometimes it notifies me of updates. At that point, I either don't want to wait or can't update, because a print is running. 

**Describe the solution you'd like**
It would be nice if I could tell it to "update after next/current print" (or similar).

**Describe alternatives you've considered**
Optional auto-updates could also be a solution.

**Additional context**
N/A

foosel:
While #3589 is a duplicate of that, I didn't realize it until it had received a bunch of upvotes already. Since we can't transfer them, I'm closing this request as a duplicate of #3589 (even though it was earlier) and suggest anyone interested in this a) give #3589 a thumbsup on the issue text and b) follow that instead.

PR #4364, opened 2022-01-08
adds an update queue to allow starting updates while printing, #3320,…

jneilliii:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

While printing if you receive the update prompt and press update now the available updates will be pushed into a queue to be run after the print is finished or cancelled. 

Ran locally in dev environment. 

#3320, #3589, #3585, #4431

image


cp2004 (changes requested):
Tested this PR today, to see what happened...

There's still this notice at the top of the UI, which is not accurate any more:
image

I have a few other comments as well. As it stands, this PR doesn't really work as it's being intended to, and needs some work to sort out the user experience as well as some bugs.

The confirmation dialog needs work as well, since it makes no mention of the fact updates won't actually happen and won't disrupt the print job, and be queued.

image

image

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -1331,6 +1328,20 @@ def perform_update(self):
 
         force = json_data.get("force", "false") in octoprint.settings.valid_boolean_trues
 
+        if self._printer.is_printing() or self._printer.is_paused():
+            # do not update while a print job is running
+            # store targets to be run later on print done event
+            if self._queued_updates is not None:
+                self._queued_updates = dict_merge(
+                    self._queued_updates, {"targets": targets, "force": force}
+                )
```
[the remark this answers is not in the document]
jneilliii: Thanks, that's my misunderstanding of how dict_merge worked relative to lists. Interesting during my testing I didn't have any issues but can't remember if I tried more than just one at a time. Will adjust.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
```diff
@@ -723,6 +722,25 @@ $(function () {
...
+                            "</pre>";
+                        self._showPopup({
+                            title: gettext("Updates Queued"),
+                            text: message,
+                            type: "info",
+                            buttons: {
+                                sticker: false
+                            }
+                        });
+                        return;
```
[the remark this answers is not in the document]
jneilliii: I assume that only really matters for a plugin that doesn't force a restart at the end of the process.

jneilliii:
Will include UI adjustments when I fix the other stuff.

src/octoprint/plugins/softwareupdate/__init__.py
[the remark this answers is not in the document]
jneilliii: slightly different approach because your method would allow duplicates.

```
            if self._queued_updates is not None:
                self._queued_updates["targets"] = list(
                    set(self._queued_updates["targets"] + targets)
                )
            else:
                self._queued_updates = {"targets": targets, "force": force}
```

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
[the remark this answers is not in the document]
jneilliii: Added toggle back to false

jneilliii:
Requested changes completed. I think I remember why I missed these now. I was doing all my testing from the Update Now pop up rather than Software Update settings.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
```diff
@@ -76,7 +77,7 @@ $(function () {
                 !self.updateInProgress &&
                 self.environmentSupported() &&
                 self.storageSufficient() &&
-                !self.printerState.isPrinting() &&
+                /*!self.printerState.isPrinting() &&*/
```
foosel, line 80: Commented code be gone ;)

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
```diff
@@ -834,13 +843,8 @@ $(function () {
                     "Be sure to read through any linked release notes, especially those for OctoPrint since they might contain important information you need to know <strong>before</strong> upgrading."
                 ) +
                 "</p>" +
-                "<p><strong>" +
-                gettext("This action may disrupt any ongoing print jobs.") +
-                "</strong></p>" +
                 "<p>" +
-                gettext(
-                    "Depending on your printer's controller and general setup, restarting OctoPrint may cause your printer to be reset."
-                ) +
```
foosel, line 843: So, the reason for this back then when was the following scenario:

User has SD print ongoing, OctoPrint currently not connected via serial but physically. User clicks update. Update runs, server reboots. Server is set to connect to printer on startup. Server connects, printer reboots (it's still one of THOSE), boom, print kaputt.

So I'm not so sure right now this should be taken out completely here, even with a queue this COULD happen.

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -1492,6 +1503,17 @@ def get_wizard_details(self):
...
+            event == Events.PRINT_DONE
+            or event == Events.PRINT_FAILED
+            and self._queued_updates is not None
+        ):
+            self.perform_updates(
+                targets=self._queued_updates["targets"],
+                force=self._queued_updates["force"],
+            )
+            self._queued_updates = None
+
```
foosel, line 1516: I'm leaning towards not immediately starting the update but having a countdown here to allow the user to still intervene.

foosel (changes requested):
So there are still some things I found and there's also some things that need discussion here I think.

First of all, this is all in the frontend. So if the user closes the browser the enqueued updates will never take place. And if the user accesses OctoPrint from two browser sessions and enqueues different updates, only one request will win and the others be locked out and fail silently. So the queue should be kept in the backend and managed centrally. 

Second of all, there's currently no way to abort the queue. As a user I'd expect a way to change my mind until the print is actually done and the update started. I'd also be able to deselect already selected updates again.

Thirdly, already mentioned in a comment, I don't think the update should just start right away on print finish/failure. What if due to a failure I still need to investigate something in the terminal or the cam or whatnot, and now the thing is updating, locking me out and after everything is gone. Give it a timeout, and only update after a *successful* & uncancelled print, otherwise show a notification with the queue maybe and wait for more manual input and also allow to throw the queue away at this point.

Otherwise I see this causing quite a few ruffled feathers and confusion, as much as I'd like to see this added.

cp2004:
> First of all, this is all in the frontend. So if the user closes the browser the enqueued updates will never take place. And if the user accesses OctoPrint from two browser sessions and enqueues different updates, only one request will win and the others be locked out and fail silently. So the queue should be kept in the backend and managed centrally.

It's in the backend the queue is managed there, responding with a 202 and the content of the queue when it can't be done immediately.

foosel:
... I feel stupid now. Maybe I shouldn't be doing code reviews this late in the day, for some reason I was 100% sure I was only looking at js files there. Disregard that point please 😐

jneilliii:
> First of all, this is all in the frontend. So if the user closes the browser the enqueued updates will never take place. And if the user accesses OctoPrint from two browser sessions and enqueues different updates, only one request will win and the others be locked out and fail silently. So the queue should be kept in the backend and managed centrally.
> 

As already figured out, this is handled on server side. However, if another session is opened I don't think the buttons are switching to queued until after an Update button is clicked and gets the data from the server side. Would the best way to handle this be an onSettingsShown callback to load the current queue or injected with on_settings_load server side? 

> Second of all, there's currently no way to abort the queue. As a user I'd expect a way to change my mind until the print is actually done and the update started. I'd also be able to deselect already selected updates again.
> 

Agreed, I noticed after changing the button from Update to Queued it would make sense to make that button cancel from being queued, which would have to call back-end and remove target from list. I assume a `Cancel All Queued` button at the top with Update All would also make sense if there are queued updates?

> Thirdly, already mentioned in a comment, I don't think the update should just start right away on print finish/failure. What if due to a failure I still need to investigate something in the terminal or the cam or whatnot, and now the thing is updating, locking me out and after everything is gone. Give it a timeout, and only update after a _successful_ & uncancelled print, otherwise show a notification with the queue maybe and wait for more manual input and also allow to throw the queue away at this point.
> 

I'll add a cancel popup similar to the Tasmota idle timeout abort popup on print success. On print cancelled show pop up that requires user to click to start updates and not automatically do it. Both with options to clear the queue.

> So I'm not so sure right now this should be taken out completely here, even with a queue this COULD happen.
>

Will revert this change and put warning back in about potential printer resets.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
jneilliii: Put back in, feel free to adjust the language to your liking.

src/octoprint/plugins/softwareupdate/__init__.py
jneilliii: added timer, but still want to re-do the way the PNotify is getting made and utilize the _showPopup function.

jneilliii:
@cp2004 would love to get some feedback/testing on the latest changes. I'm sure there are some efficiencies that could be gained somewhere. I do want to re-do the PNotify calls for the timer and cancel messages to utilize the internal _showPopup functions. Only weird thing I noticed with dev environment was when proceeding the update complete notice wouldn't show and I'm not quite sure why since I'm calling the same function as the default update function is. Any pointers greatly appreciated.

jneilliii:
I have no idea what happened with precommit. Something related to pulling in maintenance changes I think. I apologize in advance for my github noobness.

cp2004:
Will take a look at some point (when I finish my own PRs 😜) - the pre-commit changes do make it much more difficult to see the actual code changes. Everything in the pre-commit-config probably needs updating (which is what happened by the looks of it), but maybe in a separate PR? I would just remove that last commit, everything with pre-commit is a bit out of sync since all the pyupgrade changes for Python 3 and stuff.

jneilliii:
reset to before merge and forced push.

cp2004 (approved):
Looking great 👍 

Have tested as many scenarios as I can, all seem to be working well. Cancelling, adding more, cancelling prints, pausing countdowns. Progress bar looks nice.

Took a brief look through the code, don't see any issues. IMO ready to go.

cp2004 (changes requested):
Actually - there is one small thing - the countdown timer should be cancelled if a new print is started. At the moment I can:
* start print
* queue updates
* print finishes, update countdown starts
* start new print
* update happens anyway

The countdown probably just needs resetting, then it can start again when the next print is done.

jneilliii:
Good catch. I'll adjust that tonight after work.

jneilliii:
sorry for the delay, work went sideways, cancel abort timer on print start added

cp2004 (approved):

jneilliii:
small change incorporated to wait for MOVIE_DONE event to start timer if timelapse is enabled.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
```diff
@@ -99,6 +108,10 @@ $(function () {
             );
         });
 
+        // Hack to remove automatically added Cancel button
+        // See https://github.com/sciactive/pnotify/issues/141
+        PNotify.prototype.options.confirm.buttons = [];
+
```
foosel, line 114: This just tickled my memory, and then I remember that I had already added a workaround for that:

https://github.com/OctoPrint/OctoPrint/blob/1d4043736774dd9b4ea3eb8381100a38d8943fd4/src/octoprint/static/js/app/main.js#L283-L319

So just use `PNotify.singleButtonNotify` if you want to get rid of this button. What you do here will influence the defaults for any notifications spawned after first invocation of this line and thus might have nasty side effects.

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -1331,9 +1329,71 @@ def perform_update(self):
...
+        self._send_client_message(
+            "queued_updates",
+            {"targets": self._queued_updates["targets"]},
+        )
+
+        return (
+            flask.jsonify({"queued": self._queued_updates.get("targets", False)}),
+            202,
+        )
+
```
foosel, line 1396: This endpoint somehow feel icky still. And I'm talking a purely API design perspective here.

I think it would be better to have this endpoint be located at `/update/queued` and implement the command pattern also used in the `SimpleApiPlugin` mixin. So a `POST`, yes, but taking commands, only one for now, `cancel`, with optional parameter `targets`. The logic otherwise is sound.

Something like this maybe? (just written in review, untested & no pre-commit here)

```python
    @octoprint.plugin.BlueprintPlugin.route("/update/queued", methods=["POST"])
    @no_firstrun_access
    @Permissions.PLUGIN_SOFTWAREUPDATE_UPDATE.require(403)
    def cancel_queued(self):
        if "application/json" not in flask.request.headers["Content-Type"]:
            flask.abort(400, description="Expected content-type JSON")

        json_data = flask.request.get_json(silent=True)
        if json_data is None:
            flask.abort(400, description="Invalid JSON")
        
        command = json_data.get("command")
        if "command" == "cancel":
            targets = [x.strip() for x in json_data.get("targets", [])]

            if targets:
                self._queued_updates["targets"] = [x for x in self._queued_updates["targets"] if x not in targets]
            else:
                self._queued_updates["targets"] = []
                
            if (
                not self._queued_updates["targets"]
                and self._queued_updates_abort_timer
            ):
                self._queued_updates_abort_timer.cancel()
                self._queued_updates_abort_timer = None

            self._send_client_message(
               "queued_updates",
                {"targets": self._queued_updates["targets"]},
            )

            return (
                flask.jsonify({"queued": self._queued_updates["targets"]}),
                202,
            )

        else:
            flask.abort(400, description="Expected a valid command")
```

Also makes things a bit more pythonic.

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -1492,6 +1552,34 @@ def get_wizard_details(self):
...
+        if (
+            event == Events.PRINT_FAILED
+            and self._settings.global_get(["webcam", "timelapse", "type"]) == "off"
+            and len(self._queued_updates.get("targets", [])) > 0
+        ):
+            self._send_client_message(
+                "queued_updates",
+                {"print_failed": True, "targets": self._queued_updates["targets"]},
+            )
+
```
foosel, line 1582: Should be `if ... elif ...` given that there's no way more than one of these can ever be true.

src/octoprint/plugins/softwareupdate/static/clientjs/softwareupdate.js
```diff
@@ -13,6 +13,7 @@
         this.updateUrl = url + "update";
         this.configureUrl = url + "configure";
         this.updatelogUrl = url + "updatelog";
+        this.cancelQueuedURL = url + "cancel_queued";
```
foosel, line 16: If adjusted as suggested above, this would turn into

```suggestion
        this.queuedURL = this.updateUrl + "/queued";
```

src/octoprint/plugins/softwareupdate/static/clientjs/softwareupdate.js
```diff
@@ -72,6 +73,36 @@
...
+        targets = targets || [];
+        if (typeof targets === "string") {
+            targets = [targets];
+        }
+
+        var data = {
+            targets: targets
+        };
+        return this.base.postJson(this.cancelQueuedURL, data, opts);
+    };
```
foosel, line 91:
```suggestion
    OctoPrintSoftwareUpdateClient.prototype.cancelQueued = function (targets, opts) {
        if (arguments.length === 1 && _.isObject(arguments[0])) {
            var params = arguments[0];
            targets = params.targets;
        }

        targets = targets || [];
        if (typeof targets === "string") {
            targets = [targets];
        }

        var data = {
            command: "cancel",
            targets: targets
        };
        return this.base.postJson(this.queuedURL, data, opts);
    };
```

src/octoprint/plugins/softwareupdate/static/clientjs/softwareupdate.js
```diff
@@ -72,6 +73,36 @@
...
+            var params = arguments[0];
+            opts = params.opts;
+        }
+
+        var data = {
+            cancelAll: true
+        };
+
+        return this.base.postJson(this.cancelQueuedURL, data, opts);
+    };
```
foosel, line 104:
```suggestion
    OctoPrintSoftwareUpdateClient.prototype.cancelAllQueued = function (opts) {
        if (arguments.length === 1 && _.isObject(arguments[0])) {
            var params = arguments[0];
            opts = params.opts;
        }

        var data = {
            command: "cancel"
        };

        return this.base.postJson(this.queuedURL, data, opts);
    };
```

foosel (changes requested):
Got some more nitpicking, more related to API design and minor stuff though. See below.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
jneilliii: So I tried this and pushed the changes up, but it has the nasty side effect of not handling the Pnotify.update at line 1361. The initial popup doesn't have the default cancel button, but as soon as the update happens the button comes back. Maybe a better hack would be to add a hidden class to the empty cancel button?

src/octoprint/plugins/softwareupdate/__init__.py
jneilliii: These changes were made. Totally makes sense and allows for expanding in the future if necessary.

src/octoprint/plugins/softwareupdate/static/js/softwareupdate.js
jneilliii: I switched it over to a hidden button instead. Achieves the same goal with all the hackery.

foosel: Good solution, I didn't realize this would break on updating the popup. Might need a solution in core but for now this works.

foosel (approved):

foosel:
🥳 Thank you!

Issue #4375, opened 2022-01-19
Rotate/Mirror options not applied to HLS webcam stream


crysxd:
### The problem

1. Setup HLS webcam in OctoPi
2. Enable the rotate option in settings
3. Go back to webcam

Expected: Image rotated
Actual: Image not rotated

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

maintenance branch

### Operating system running OctoPrint

OctoPi

### Printer model & used firmware incl. version

/

### Browser and version of browser, operating system running browser

Safari, Chrome

### Checklist of files to include below

- [X] [Systeminfo Bundle (always include!)](https://github.com/OctoPrint/OctoPrint/files/7894646/octoprint-systeminfo-20220119063500.zip)
- [X] [Contents of the JavaScript browser console (always include in cases of issues with the user interface)](https://github.com/OctoPrint/OctoPrint/files/7894641/console.log)
- [X] [Screenshots and/or videos showing the problem (always include in case of issues with the user interface)](https://user-images.githubusercontent.com/5859228/150077449-3cd9606c-31c7-4817-940a-b192f7e9753d.mov)
- [ ] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

### Additional information & file uploads

Happy to pick this up myself! Just wanted to  raise the issue here so it is documented for others. Most likely also applies to RTSP

cp2004:
1.8.0 has been released 🎉

PR #4378, opened 2022-01-22
HLS & WebRTC webcam rotation and mirroring

crysxd:
This PR adds support for the rotate and mirror options in the webcam settings to HLS and WebRTC video playback. If the rotation option is disabled, the behaviour is identical as before. If the rotate option is enabled the video is rotated 90 degrees and resized to fit.

The native video controls present for WebRTC content did not work with the rotation and mirroring. Thus I added controls to mute/unmute the video and to launch fullscreen or Picture in Picture (shown if supported). HLS did not have video controls so far, I now added the same controls there.

Tested in Chrome and Safari with OctoPi HLS stream and @Triodes WebRTC app

/

#4375

Screenshot 2022-01-23 at 09 01 32

https://user-images.githubusercontent.com/5859228/150669987-1988ebff-68ac-4a6f-8dde-ef63b67fb2ba.mp4

#### Further notes
Not quite in scope of this PR, but the settings pane should maybe address the differences in how MJPEG and HLS/WebRTC aspect ratio is handled as MJPEG uses a fixed aspect ratio from settings with letterboxing and HLS/WebRTC always use the source aspect ratio. Potentially we can also apply the logic introduced in this PR to MJPEG and remove the fixed aspect ratio to unify the behaviour.


src/octoprint/static/js/app/viewmodels/control.js
```diff
@@ -765,6 +782,37 @@ $(function () {
             self.webcamHlsEnabled(false);
             self.webcamWebRTCEnabled(true);
         };
+
+        self._updateVideoTagWebcamLayout = function (containerId) {
```
foosel, line 786: This needs more comments tbh. I see myself looking at this code in half a year and having NO idea anymore what the sections are supposed to achieve.

src/octoprint/templates/tabs/control.jinja2
```diff
@@ -1,10 +1,46 @@
...
+                <!-- /ko -->
+            </div>
+        </div>
+        <div data-bind="css: { webcam_rotated: settings.webcam_rotate90(), webcam_unrotated: !settings.webcam_rotate90() }">
+            <div class="rotation_target">
+                <div data-bind="css: { flipH: settings.webcam_flipH(), flipV: settings.webcam_flipV() }">
+                    <video id="webcam_hls" muted autoplay playsinline style="width: 100%"></video>
+                </div>
+            </div>
+        </div>
```
foosel, line 43: All of this is pretty much duplicated. This screams for the use of a ko template to reduce duplication, and the logic in the click handlers should definitely go into descriptively named methods on the view model as well.

crysxd: I was actually wondering if we need that all twice...webrtc and HLS both use `<video>` so I think we can just have one set of elements for them. Are you aware of any reason we have this all twice?

src/octoprint/static/js/app/viewmodels/control.js
crysxd: Will do :)

crysxd:
The controls are not really a new feature as they were already present before (not for HLS but for WebRTC). The problem is that I have to remove the native controls for the rotation as the controls are rotated/mirrored as well and generally glitch out on Chrome. This means I removed the native controls and in order to not cut back on functionality, I need to replace them. 

Safari displays the native controls fine where rotated (but rotated), Chrome creates some weird layout bugs:
Screenshot_2022-01-22_at_15 12 23

I can still make it a separate PR if you want :)

src/octoprint/templates/tabs/control.jinja2
foosel: It was less duplication before, so it didn't feel as bad as now 😅 I'd suggest to either deduplicate it by reusing the same snippet for both types or at least use a template to not *actually* duplicate the source.

foosel:
> The controls are not really a new feature as they were already present before (not for HLS but for WebRTC).

Aaah, gotcha! In that case please disregard the comment, reimplementing that as part of this PR makes sense then.

src/octoprint/templates/tabs/control.jinja2
crysxd: Then let me not duplicate anything :D Sharing the video element caused some glitches when switching from HLS to webRTC, so I still have two of them but the rest is all shared.

src/octoprint/static/js/app/viewmodels/control.js
crysxd: Added

foosel (changes requested):

src/octoprint/templates/tabs/control.jinja2
```diff
@@ -1,10 +1,26 @@
...
+    <div id="webcam_video_container" tabindex="0" data-bind="visible: webcamWebRTCEnabled() || webcamHlsEnabled(), css: { webcam_rotated: settings.webcam_rotate90(), webcam_unrotated: !settings.webcam_rotate90() }">
+        <div class="video-controls btn-group action-buttons">
+            <div class="btn btn-mini" data-bind="click: function(data, event) { if (webcamWebRTCEnabled()) { document.getElementById('webcam_webrtc').requestFullscreen() } else { document.getElementById('webcam_hls').requestFullscreen() } }"><i class="fas fa-expand"></i></div>
+            <div class="btn btn-mini" data-bind="visible: document.pictureInPictureEnabled, click: function(data, event) {  if (webcamWebRTCEnabled()) { document.getElementById('webcam_webrtc').requestPictureInPicture() } else { document.getElementById('webcam_hls').requestPictureInPicture() } }"><i class="fas fa-images"></i></div>
+            <div class="btn btn-mini" data-bind="click: function(data, event) { webcamMuted(!webcamMuted()); document.getElementById('webcam_webrtc').muted = webcamMuted(); document.getElementById('webcam_hls').muted = webcamMuted() }">
```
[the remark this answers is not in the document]
crysxd: Will do!

crysxd:
Extracted the click handler and also got rid of the magic style selector to get the active video element 🙂

PR #4382, opened 2022-01-24
Adding event markings on the temperature graph

### LINKED ISSUE #3261: Marks on the temperature graph [Request]
It would be nice to tell how the temperature relates with pausing/stopping prints, so marks/lines indicating when the print was started/paused/stopped etc would be useful. This would be done by waiting for specific gcode commands, and logging them with the time to display them on the temperature graph.

Thanks, Quentin

surdu:
    Brainstorming ticket
  * [ ] Your PR targets OctoPrint's devel branch if it's a completely
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Adds event markings on the temperature graph so it will be easier to correlate temperature variations with printing events.

Using the virtual printer, started and stopped a bunch of prints. Left the marking reach the end of the graph and made sure they will disappeared once they get at the end of the graph.

None
Closes #3261 
Screenshot 2022-01-24 at 23 55 07

#### Further notes
None

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -64,10 +64,14 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         self._targetTemp = None
         self._targetBedTemp = None
         self._targetChamberTemp = None
-
-        self._temps = TemperatureHistory(
+        
+        self._temps = DataHistory(
             cutoff=settings().getInt(["temperature", "cutoff"]) * 60
         )
+        self._markings = DataHistory(
+            cutoff=settings().getInt(["temperature", "cutoff"]) * 60
+        )
+
         self._messages = deque([], 300)
         self._log = deque([], 300)
 
@@ -2084,11 +2088,11 @@ class StateMonitor:
         }
 
 
-class TemperatureHistory(InvariantContainer):
+class DataHistory(InvariantContainer):
     def __init__(self, cutoff=30 * 60):
-        def temperature_invariant(data):
+        def data_invariant(data):
             data.sort(key=lambda x: x["time"])
             now = int(time.time())
             return [item for item in data if item["time"] >= now - cutoff]
 
-        InvariantContainer.__init__(self, guarantee_invariant=temperature_invariant)
+        InvariantContainer.__init__(self, guarantee_invariant=data_invariant)
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -718,6 +718,15 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 
         return self._comm.getFilePosition()
 
+    def addMarking(self, type):
+        self._markings.append({
+            "type": type,
+            "time": int(time.time() * 1000),
+        })
+
+    def get_markings(self):
+        return self._markings
+
     def start_print(self, pos=None, user=None, *args, **kwargs):
         """
         Starts the currently loaded print job.
@@ -739,6 +748,7 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         self._lastProgressReport = None
         self._updateProgressData()
         self._setCurrentZ(None)
+        self.addMarking("Print")
         self._comm.startPrint(
             pos=pos,
             user=user,
@@ -754,6 +764,8 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 
         if self._comm.isPaused():
             return
+
+        self.addMarking("Pause")
 
         self._comm.setPause(
             True,
@@ -771,6 +783,8 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         if not self._comm.isPaused():
             return
 
+        self.addMarking("Resume")
+
         self._comm.setPause(
             False,
             user=user,
@@ -783,6 +797,8 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         """
         if self._comm is None:
             return
+
+        self.addMarking("Cancel")
 
         # tell comm layer to cancel - will also trigger our cancelled handler
         # for further processing
@@ -1390,6 +1406,7 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
                 temps=list(self._temps),
                 logs=list(self._log),
                 messages=list(self._messages),
+                markings=list(self._markings),
             )
 
             plugin_data = self._get_additional_plugin_data(initial=False)
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -750,7 +750,6 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         self._lastProgressReport = None
         self._updateProgressData()
         self._setCurrentZ(None)
-        self.addMarking("Print")
         self._comm.startPrint(
             pos=pos,
             user=user,
@@ -766,8 +765,6 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
 
         if self._comm.isPaused():
             return
-
-        self.addMarking("Pause")
 
         self._comm.setPause(
             True,
@@ -785,8 +782,6 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
         if not self._comm.isPaused():
             return
 
-        self.addMarking("Resume")
-
         self._comm.setPause(
             False,
             user=user,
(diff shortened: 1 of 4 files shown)


src/octoprint/printer/standard.py
```diff
@@ -742,6 +743,17 @@ def get_file_position(self):
 
         return self._comm.getFilePosition()
 
+    def addMarking(self, type):
+        self._markings.append(
+            {
+                "type": type,
+                "time": int(time.time() * 1000),
+            }
+        )
```
foosel, line 752: This needs some way to age out old markings. Imagine the printer is online for a month or something (there's nothing that would speak against that), and being used. The memory would slowly get filled with ancient graph markings that aren't even displayed anymore but will still be kept and also sent to the printer on every single state update. That's bad.

Limit this to the temperature cutoff time (`settings().getInt(["temperature", "cutoff"]) * 60`, see `self._temps`), ideally create another `InvariantContainer` like `TemperatureHistory` (or possibly just reuse that by instancing it again, after renaming it to something like `TimedHistory`).

Also, stick to `time.time()` timestamps on the backend, translate to JS-site timestamps on the frontend.

src/octoprint/server/util/sockjs.py
```diff
@@ -449,6 +449,7 @@ def on_printer_send_current_data(self, data):
                 "logs": self._filter_logs(logs),
                 "messages": messages,
                 "busyFiles": busy_files,
+                "markings": self._printer.get_markings(),
```
foosel, line 452: This is sending all markings, every time. Once the mentioned cutoff is implemented, that shouldn't be such a big of an issue resource wise, however it still diverts from the per-connection backlog implementation used for temperatures, messages and logs and thus should probably be aligned.

src/octoprint/static/js/app/viewmodels/temperature.js
```diff
@@ -200,13 +200,28 @@ $(function () {
...
+            PAUSE: "Pause",
+            RESUME: "Resume",
+            CANCEL: "Cancel"
+        };
+
+        var markingColors = {};
+        markingColors[self.MARKING_TYPE.PRINT] = "#218656";
+        markingColors[self.MARKING_TYPE.PAUSE] = "#FDC02F";
+        markingColors[self.MARKING_TYPE.RESUME] = "#27CAEE";
+        markingColors[self.MARKING_TYPE.CANCEL] = "#DA3749";
```
foosel, line 215: I'd suggest the following here... 

```suggestion
        self.marking_props = {
            "start": gettext("Start"),
            "pause": gettext("Pause"),
            "resume": gettext("Resume"),
            "cancel": gettext("Cancel"),
        }
```

Then switch the backend markings to lower case names (& also add a regular `done` one because that's kinda missing right now) and use them to create an additional css class attached to each label (`tempMarkingLabel-{type}`). Adjust the less file, include the color definitions for the classes in there (because that way any third party themes can also override the colors).

src/octoprint/printer/standard.py
```diff
@@ -808,6 +825,8 @@ def cancel_print(self, user=None, *args, **kwargs):
         if self._comm is None:
             return
 
+        self.addMarking("Cancel")
```
foosel, line 828: `cancel`

src/octoprint/printer/standard.py
```diff
@@ -795,6 +810,8 @@ def resume_print(self, user=None, *args, **kwargs):
         if not self._comm.isPaused():
             return
 
+        self.addMarking("Resume")
```
foosel, line 813: `resume`

src/octoprint/printer/standard.py
```diff
@@ -779,6 +792,8 @@ def pause_print(self, user=None, *args, **kwargs):
         if self._comm.isPaused():
             return
 
+        self.addMarking("Pause")
```
foosel, line 795: `pause`

src/octoprint/printer/standard.py
```diff
@@ -763,6 +775,7 @@ def start_print(self, pos=None, user=None, *args, **kwargs):
         self._lastProgressReport = None
         self._updateProgressData()
         self._setCurrentZ(None)
+        self.addMarking("Print")
```
foosel, line 778: `start`

And where's `done`?

foosel (changes requested):
I like the idea! 👍

However, I've got a bunch of points, see the review notes.

Additionally, this should not hook into the printer methods but instead hook into the `on_comm_printjob_*` callbacks instead. The reason being that action commands or other events on the printer might change states, and while the state would be synced and the events for that would be triggered, the corresponding printer methods won't be called in all cases, and you'd miss marks that way. In other words, `printer.{start|pause|resume|cancel}_print` aren't the only reasons a state change will happen. However, the callbacks *will* always be called when the printer state changes, and are actually what also triggers the system wide events, so that's the right place for any such handling.

Happy to clarify or elaborate if needed 😊

src/octoprint/static/js/app/viewmodels/temperature.js
surdu: Unfortunately, the lines are drawn on the canvas, only the labels are html elements, so I don't believe I'll be able to make the colors CSS customizable 😞

foosel: Ah, good point. Then leave it with the colors hard coded and I'll see if I can come up with something long term.

src/octoprint/static/js/app/viewmodels/temperature.js
```diff
@@ -200,13 +200,40 @@ $(function () {
...
+        self.marking_props[self.MARKING_TYPE.DONE] = gettext("Done");
+
+
+        self.markingColors = {};
+        self.markingColors[self.MARKING_TYPE.PRINT] = "#218656";
+        self.markingColors[self.MARKING_TYPE.PAUSE] = "#FDC02F";
+        self.markingColors[self.MARKING_TYPE.RESUME] = "#27CAEE";
+        self.markingColors[self.MARKING_TYPE.CANCEL] = "#DA3749";
+        self.markingColors[self.MARKING_TYPE.DONE] = "#1B72F9";
+
```
foosel, line 228:
```suggestion
        self.marking_props = {
            "print": {
                "label": gettext("Start"),
                "color": "#218656"
            },
            "pause": {
                "label": gettext("Pause"),
                "color": "#FDC02F"
            },
            "resume": {
                "label": gettext("Resume"),
                "color": "#27CAEE"
            },
            "cancel": {
                "label": gettext("Cancel"),
                "color": "#DA3749"
            },
            "done": {
                "label": gettext("Done"),
                "color": "#1B72F9"
            },
        }
```

And then use with `self.marking_props[mark.type].label` or `self.marking_props[mark.type].color`. Way less repetition and only one props object needed instead of multiple.

src/octoprint/static/js/app/viewmodels/temperature.js
surdu: I _could_ try some hacks like: creating the labels, getting the computed styles for the labels and determine the color for each and then remove the temporary labels

foosel: Not important enough (for now) to introduce hacks ^^

src/octoprint/printer/standard.py
```diff
@@ -718,6 +722,17 @@ def get_file_position(self):
 
         return self._comm.getFilePosition()
 
+    def addMarking(self, type):
```
foosel, line 725: Sorry for not catching that earlier, but this should actually be `add_marking`. Yes, a lot of OctoPrint's older methods, e.g. the one referenced just above this, still make use of camelCase, but new stuff should stick to Python's established standard of snake_case (and I'm also trying to make old stuff fit whenever I do refactoring in the general region as well).

foosel (changes requested):
Could you take care of the pre-commit failure and the one thing I missed last time? Thank you!

surdu:
No problem. Hope that you're ok with the new name for the `TemperatureHistory` class. If not, please let me know 😉

foosel:
As you write protected this PR I could not push some final changes here before squashing and merging it, so instead I've squashed and merged everything locally as 9d0b2953119cce993b90b4c912591af8f5bf8574 and pushed it to `maintenance`. So this PR *is* merged, even though GitHub is not able to automatically detect that. Thank you!

surdu:
Sorry, it was the default option and was not aware to change it 😞 

Glad I could help!

foosel:
Might actually have been my fault after all, in either case, no need to apologize, all good :)

PR #4384, opened 2022-01-27
Added Try/Except test to prevent I/O errors…

GonzoDMX:
This change handles serial posix I/O errors generated when attempting to connect to a nonexistent or busy serial port.
If a SerialException occurs when attempting to connect to a port it is handled and a log of the failed connection attempt is generated, finally None is returned, which is handled properly by the existing checks.

Tested in OctoPrint dev environment. The bug can be reproduced by setting the serial port to AUTO and connecting the virtual printer in a dev environment. In the real world it could potentially occur with a faulty serial connection, if linux fails to update port availability in a timely manner or when a port is improperly closed.

#### Any background context you want to provide
I am developing a plugin for automating serial communications between OctoPrint and peripheral (non-printer) devices. I stumbled across this problem while testing my plugin.

None, I fixed the issue before thinking to submit a ticket.

Best regards,
-Andrew O'Shei

This is my first PR on OctoPrint, I took a look through the guidelines but I apologize in advance if I slipped up somewhere.

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -3745,7 +3745,7 @@ class MachineCom(object):
                 serial_obj.open()
 
             except serial.SerialException:
-                self._logger.info("Failed to connect: Port {} does not exist".format(p))
+                self._logger.info("Failed to connect: Port {} is busy or does not exist".format(p))
                 return None
 
             # Set close_exec flag on serial handle, see #3212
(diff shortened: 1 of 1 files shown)


GonzoDMX:
Sorry, I forgot to check the target before submitting the PR. It should be good now.
-Andrew

cp2004 (approved):
Seems to make sense based on my reading/understanding of the code here.

foosel (approved):

PR #4389, opened 2022-01-30
Addresses issue #4362

### LINKED ISSUE #4362: GitHub tags are only read up to the first `-` character, breaking tags such as `1.0.0-rc1`
### The problem

The software update plugin does a load of parsing to compare the GitHub tag version to the one from the installed Python packages. I unwrapped some of it to figure out what was going on here.

We have a tag on GitHub with the name `1.0.0-rc1`, which according to [PEP440](https://www.python.org/dev/peps/pep-0440/#pre-release-separators) is allowed but would be normalized to `1.0.0rc1`, which is how it comes through if this tag is set in setup.py.

The plugin runs it through (as far as I could trace) `octoprint.plugins.softwareupdate.version_checks.github_release.get_comparable_version`. This calls some functions in `octoprint.util.version` including `normalize_version`, which only reads the tag up to and including the `-`, stripping it down to just `1.0.0` in this case, which would be seen as 'ahead' of `1.0.0rc1` of course.

Some testing to demonstrate, the first line is the problem one:
```python
>>> get_comparable_version("1.1.1-rc1")
<Version('1.1.1')>
>>> get_comparable_version("1.1.1rc1")
<Version('1.1.1rc1')>
>>> get_comparable_version("1.1.1.rc1")
<Version('1.1.1rc1')>
>>> get_comparable_version("1.1.1_rc1")
<Version('1.1.1rc1')>
```

They should all be parsed equally according to PEP440.

This resulted in an update loop where `1.0.0-rc1` was set the same in both `setup.py` and the GitHub tag, but it became GitHub tag = 1.0.0 and setup.py = 1.0.0rc1 when parsed and the comparison would always show an update available.

**TLDR**; everything after the `-` in a GitHub tag is ignored, when it should be included.

Discussion on discord: https://discord.com/channels/704958479194128507/708230829050036236/928402397435420775

Yes, it did persist

NA

### Version of OctoPrint

1.7.2

All

All

All

_No response_

GonzoDMX:
Fixes the normalize_version function in src/util/version.py so that it returns the correct version when the input version string is a release candidate containing the characters "-rc".

Tested in OctoPrint dev environment.

Fixes #4362

### DIFF
--- a/src/octoprint/util/version.py
+++ b/src/octoprint/util/version.py
@@ -167,8 +167,8 @@ def get_comparable_version(version_string, cut=None, **kwargs):
     if cut is not None and (cut < 0 or not isinstance(cut, int)):
         raise ValueError("level must be a positive integer")
 
-    version = pkg_resources.parse_version(version_string)
     version_string = normalize_version(version_string)
+    version = pkg_resources.parse_version(version_string)
 
     if cut is not None:
         if isinstance(version, tuple):
--- a/src/octoprint/util/version.py
+++ b/src/octoprint/util/version.py
@@ -209,7 +209,7 @@ def is_prerelease(version_string):
 
 def normalize_version(version):
     if "-" in version:
-        version = version[: version.find("-")]
+        version = version.replace('-', '')
 
     # Debian has the python version set to 2.7.15+ which is not PEP440 compliant (bug 914072)
     if version.endswith("+"):
--- a/src/octoprint/util/version.py
+++ b/src/octoprint/util/version.py
@@ -210,7 +210,7 @@ def is_prerelease(version_string):
 def normalize_version(version):
     if "-rc" in version:
         version = version.replace("-rc", "rc")
-    
+
     if "-" in version:
         version = version[: version.find("-")]
 
--- a/src/octoprint/util/version.py
+++ b/src/octoprint/util/version.py
@@ -209,7 +209,8 @@ def is_prerelease(version_string):
 
 def normalize_version(version):
     if "-" in version:
-        version = version.replace('-', '')
+        index = version.rfind("-")
+        version = version[:index:] + version[(index+1):]
 
     # Debian has the python version set to 2.7.15+ which is not PEP440 compliant (bug 914072)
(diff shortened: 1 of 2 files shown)


src/octoprint/util/version.py
```diff
@@ -208,6 +208,9 @@ def is_prerelease(version_string):
 
 
 def normalize_version(version):
+    if "-rc" in version:
+        version = version.replace("-rc", "rc")
+
```
cp2004, line 213: This only covers the very specific case of having `-rc`.

By PEP440 version spec, https://www.python.org/dev/peps/pep-0440/#pre-release-separators, there could be `-a`, `-b`, `-c`, `-alpha`, `-beta`, `-pre`, `-preview`, `-rc` and possibly more I've missed off.

Probably better to either leave the `-` in (it should be parsed properly by `pkg_resources` or whatever's doing it) or always remove it.

cp2004 (changes requested):

src/octoprint/util/version.py
GonzoDMX: I agree this is very naive attempt at fixing the issue, initially I had proposed a more general solution along those lines however, leaving the `-` will fail the unit test and simply removing the `-` for all cases will fail the unit test. I changed to commit to see if this would pass.
I believe leaving it in or simply removing it for all cases should work fine but it will fail a unit test... so maybe that's what needs to change?

GonzoDMX: So more details on the unitest, it tests the string "1.6.0-dev123" and expects "1.6.0". Problem is I believe `-dev` is valid in PEP440.
So the best solution is probably to correct the test and leave in the `-`

GonzoDMX:
I've updated the PR, upon further assessment of the problem I believe removing the `-` char is redundant. The version is passed immediately after to pkg_resources.parse_version() which does this for us already.
I've updated the unit test for the function normalize_versions() to remove the case testing the `-` character.

cp2004 (approved):
I think that's the right way to go 👍

PR #4390, opened 2022-01-31
Add support for latin_1 encoding

synman:
- added serial setting "encodingScheme" and defaulted it to "ascii" (current value)
- updated settings API to support encodingSheme setting
- replaced hardcoded references to "ascii" encoding in comm.py to use new encodingScheme setting
- updated AUTHORS.md
- updated config_yaml.rst

  * [ X] Your changes are not possible to do through a plugin and relevant
  * [N/A ] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
  * [ X] Your PR targets OctoPrint's devel branch if it's a completely
  * [ X] Your PR was opened from a custom branch on your repository
  * [ X] Your PR only contains relevant changes: no unrelated files,
    if necessary!
  * [X ] Your changes follow the existing coding style
  * [ N/A] If your changes include style sheets: You have modified the
    lessc)
  * [ X] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [X ] You have run the existing unit tests against your changes and
    nothing broke
  * [ X] You have added yourself to the AUTHORS.md file :)

Parameterize the serial encoding method so that "extended" ascii characters can be sent over the serial link.  

I made the changes on my local, deployed them to a test machine, and tested with encodingScheme being unset (defaults to current hardcoded value of "ascii") and set to "latin_1".   I limited the selection to just these two encoding schemes.

I maintain a plugin for GRBL laser engravers and CNC machines.  GRBL has some quirky commands that live in the 128-255 character space.

Shared info about it on discord #dev-core

N/A

#### Further notes
None that I can think of


src/octoprint/server/api/settings.py
```diff
@@ -964,6 +965,11 @@ def _saveSettings(data):
                 ["serial", "ignoreEmptyPorts"], data["serial"]["ignoreEmptyPorts"]
             )
 
+        if "encodingScheme" in data["serial"]:
+            value = data["serial"]["encodingScheme"]
+            if value in ("ascii", "latin_1"):
```
foosel, line 970: tbh, I would actually go so far here to just check if it's a valid encoding identifier in general, and if so use it. Don't see why we should limit it explicitly to ascii and *latin1* of all things. If this is made configurable, I'd suggest to make it *really* configurable. If someone wants to run this against a device that talks `cp1252` for some reason, who am I to judge?

src/octoprint/settings.py
```diff
@@ -168,6 +168,7 @@ def settings(init=False, basedir=None, configfile=None, overlays=None):
         "resendRatioThreshold": 10,
         "resendRatioStart": 100,
         "ignoreEmptyPorts": False,
+        "encodingScheme": "ascii",
```
foosel, line 171: I think just `encoding` would be better.

foosel (changes requested):
Have two small adjustments to request.

src/octoprint/settings.py
synman: You got it on both counts.  I didn't have sufficient context as to how this would be received and took the more restrictive path.

I'll amend with both requested changes this evening (US ET time).

Thank you for helping further my non-typical use case.

synman:
PR updated as requested -- had a couple free minutes to address

foosel (approved):

PR #4402, opened 2022-02-05
✨  Add `--no-color` option to disable colored output on console

### LINKED ISSUE #4396: Standard output is unreadable on console
### The problem

I didn't do a git pull for a few months and somehow the console output became colorized so much that it's basically unreadable. This was okay before, at least in October 2021.

_No response_

### Version of OctoPrint

1.7.3

lI

Linux, Fedora 35

Firefox

- [X] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)

[octoprint-systeminfo-20220131185951.zip](https://github.com/OctoPrint/OctoPrint/files/7974982/octoprint-systeminfo-20220131185951.zip)
octocolor1

cp2004:

Add `--no-color` option to disable colored output on console.

Closes #4396

Also removed the setting of INFO logs to white, instead it will be left as default. If there is a light-colored background (such as white) then it becomes unreadable. Can then just use whatever user has configured for their default and not worry about contrast issues.

```
> octoprint --help
Usage: octoprint [OPTIONS] COMMAND [ARGS]...

Options:
  -b, --basedir PATH  Specify the basedir to use for configs, uploads,
                      timelapses etc.
  -c, --config PATH   Specify the config file to use.
  --overlay PATH      Specify additional config overlays to use.
  -v, --verbose       Increase logging verbosity.
  --safe              Enable safe mode; disables all third party plugins.
  --no-color          Disable colored console output
  --version           Show the version and exit.
  --help              Show this message and exit.
```

Manually by starting with and without the flag.

### DIFF
--- a/src/octoprint/cli/server.py
+++ b/src/octoprint/cli/server.py
@@ -31,6 +31,7 @@ def run_server(
     ignore_blacklist,
     octoprint_daemon=None,
     overlays=None,
+    disable_color=False,
 ):
     """Initializes the environment and starts up the server."""
 
@@ -120,6 +121,7 @@ def run_server(
             ignore_blacklist=ignore_blacklist,
             after_safe_mode=log_startup,
             after_environment_detector=log_register_rollover,
+            disable_color=disable_color,
         )
         (
             settings,
@@ -307,6 +309,7 @@ def serve_command(ctx, **kwargs):
     safe_mode = "flag" if get_value("safe_mode") else None
     ignore_blacklist = get_value("ignore_blacklist")
     overlays = get_value("overlays")
+    no_color = bool(get_value("no_color"))
 
     if v4 and not host:
         host = "0.0.0.0"
@@ -324,6 +327,7 @@ def serve_command(ctx, **kwargs):
         safe_mode,
         ignore_blacklist,
         overlays=overlays,
+        disable_color=no_color,
     )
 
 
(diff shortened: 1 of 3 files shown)


src/octoprint/cli/server.py
```diff
@@ -307,6 +309,7 @@ def get_value(key):
     safe_mode = "flag" if get_value("safe_mode") else None
     ignore_blacklist = get_value("ignore_blacklist")
     overlays = get_value("overlays")
+    no_color = bool(get_value("no_color"))
```
zaitcev, line 312: Why are we changing the style of boolean? The safe mode uses None-not-None, and we have Python bool for no-color. Seems perhaps unnecessarily inconsistent, if I may bikeshed a little.

zaitcev (commented):
Looks okay to me, although weirdly different from existing code in a tiny detail.

src/octoprint/cli/server.py
foosel: `safe_mode` is a special case, as it doesn't set a boolean value but rather a source of the safe mode activation. If it is anything but `None`, OctoPrint will run in safe mode, but it will log the reason for that based on the value of this flag.

However, the explicit cast to `bool` should indeed not be needed here, as this should already be a boolean thanks to being marked with `is_flag=True` in the click def.

PR #4403, opened 2022-02-05
🔨  Add `octoprint dev css:build` command

cp2004:

Has the necessary paths pre-configured and can build all the less files in one go, if you so wish.

Manually

Several times asked on discord how the less files are built, people are expecting some kind of easy command like this, rather than typing out the paths manually/individually. So I made one.

#### Details

```
> octoprint dev css:build --help
Usage: octoprint dev css:build [OPTIONS]

Options:
  -f, --file TEXT  Specify files to build, for a list use --list
  --all            Build all less files
  --list           List all available files
  --help           Show this message and exit.
```

If the `--all` option is specified, then all the files are built.

`-f` or `--file` can be used multiple times, for example:
```
octoprint dev css:build -f core -f login -f plugin_backup
```

List of available files/identifiers can be found using:
```
> octoprint dev css:build --all --list
Available files to build:
- core
- login
- recovery
- plugin_announcements
- plugin_appkeys_core
- plugin_appkeys_authdialog
- plugin_backup
- plugin_gcodeviewer
- plugin_logging
- plugin_pluginmanager
- plugin_softwareupdate
```

It then iterates through the files, building them!

#### Additional Information

I added the `-x` option, because all the files are compressed at the moment. However, the `-x` is deprecated, so a replacement may need to be found.

Tested with `less` 4.1.2 - there's a separate incompatibility blocking the core build with less 4.

### DIFF
--- a/src/octoprint/cli/dev.py
+++ b/src/octoprint/cli/dev.py
@@ -255,11 +255,11 @@ class OctoPrintDevelCommands(click.MultiCommand):
             "-f",
             "files",
             multiple=True,
-            help="Specify files to build, for a list use --list",
+            help="Specify files to build, for a list of options use --list",
         )
         @click.option("--all", "all_files", is_flag=True, help="Build all less files")
         @click.option(
-            "--list", "list_files", is_flag=True, help="List all available files"
+            "--list", "list_files", is_flag=True, help="List all available files and exit"
         )
         def command(files, all_files, list_files):
             import os.path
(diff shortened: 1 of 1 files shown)


src/octoprint/cli/dev.py
```diff
@@ -249,6 +248,122 @@ def command(name):
...
+                "login": {
+                    "source": "static/less/login.less",
+                    "output": "static/css/login.css",
+                },
+                "recovery": {
+                    "source": "static/less/recovery.less",
+                    "output": "static/css/recovery.css",
+                },
+                "plugin_announcements": {
+                    "source": "plugins/announcements/static/less/announcments.less",
```
foosel, line 282:
```suggestion
                    "source": "plugins/announcements/static/less/announcements.less",
```

foosel (changes requested):
One typo causing the announcements compile to not work as expected.

Also, I'm wondering if maybe the discovery of the bundled plugin files shouldn't be made automatic. As in, iterate over `src/octoprint/plugins`, check if there's a `static/less` and `static/css` inside and if so derive options as needed. As things are now I bet any new bundled plugins will be forgotten to be added here as well.

foosel (approved):

Issue #4368, opened 2022-01-13
[Request]Make configurable the g-code sent when CANCEL button is pressed and a SD print is on going


mriscoc:
### Is your feature request related to a problem? Please describe.

When the printer (Ender3v2) is printing from the SD card and CANCEL button is pressed in the octopi console, the Octoprint is issuing a M25 command:
```
Changing monitoring state from "Printing from SD" to "Cancelling"
Send: M108
Send: M25
```
That starts the pause sequence in the printer, if Advanced Pause is active, the printer goes to the parking position and wait for the user to continue the print, it is not possible to cancel the pause.

There is a better command for current Marlin 2.x printers: M524 https://marlinfw.org/docs/gcode/M524.html this command allows cancel the print and return the control to the user.

### Describe the solution you'd like

The g-code for cancel a SD print could be configurable, put a M524 in the "After print job is cancelled" g-code script do not prevent the M25 to be send.

Octoprint could check if the printer firmware is Marlin 2.x and make the change automatically

jneilliii:
It does seem that this is hard coded [here](https://github.com/OctoPrint/OctoPrint/blob/53eb9bf979246db739c4326fda1f186b91840554/src/octoprint/util/comm.py#L1755) (maintenance branch). It would make sense for this to be configurable.

Dominik-1980:
I can confirm if I print from SD on my Anycubic Vyper, I can't pause or cancel the print in octoprint.
I stucks in "canceling" forever and the only way is to turn the printer off and on again.
The M524 command Is what we need on Marlin 2.x
If I send it in the terminal the printer cancels the print and shows me the message in the touchscreen.

mriscoc:
For more recent printers it could be helpful that the default value was M524, and configurable to M25 if it is needed.

foosel:
@jneilliii I see that you made a commit referencing this issue in your fork - do you plan on sending a PR for this, are there issues I can help ironing out so we can get this included in 1.8.0?

jneilliii:
My issue was in the commit note. Not sure why it's not initially binding. I'll start a PR so we can discuss.

jneilliii:
@mriscoc the changes have been merged with the above commit and will be available as part of version 1.8.0.

mriscoc:
Excellent, thank you.

cp2004:
1.8.0 has been released 🎉

PR #4408, opened 2022-02-08
adds a setting for sd cancel command

jneilliii:
not binding to settingsViewModel as expected, needs adjustment to API?

    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

adds a setting for sd cancel command

installed locally in dev environment and verified new setting in settings

#4368 

image

#### Further notes
~~setting is not binding on initial load for some reason, and not sure why~~

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -575,6 +575,7 @@ class MachineCom(object):
         self._sdAlwaysAvailable = settings().getBoolean(["serial", "sdAlwaysAvailable"])
         self._sdRelativePath = settings().getBoolean(["serial", "sdRelativePath"])
         self._sdLowerCase = settings().getBoolean(["serial", "sdLowerCase"])
+        self._sdCancelCommand = settings().get(["serial", "sdCancelCommand"])
         self._blockWhileDwelling = settings().getBoolean(["serial", "blockWhileDwelling"])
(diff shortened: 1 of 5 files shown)


jneilliii:
> For the settings to bind, it needs to be added to the API output for settings. Unlike with plugins where it is everything minus exclusions, core works with explicitly forming the settings API response.

That's kind of what I thought and noted in my commit message, jut haven't got around to adjusting that part yet.

jneilliii:
Fixed the binding issue, thanks for the pointer Charlie, should be ready to review now. Need to verify it actually sends the command, as the virtual printer is too fast to catch a pause with and I never print from SD, but the logic should be right.

src/octoprint/util/comm.py
```diff
@@ -1752,7 +1752,7 @@ def _on_M400_sent():
             if self.isSdFileSelected():
                 if not external_sd:
                     self.sendCommand(
-                        "M25",
+                        settings().get(["serial", "sdCancelCommand"]),
```
foosel, line 1755: This should be read into a variable once at instantiation of the comm instance. See how `self._hello_command` is done.

foosel (changes requested):

src/octoprint/util/comm.py
jneilliii: requested adjustments have been made

foosel (approved):

Issue #2243, opened 2017-11-25
GCode Viewer: confused when parts are printed once at a time


bzed:
#### What were you doing?

I'm printing some parts with the 'one at a time' setting of cura, so as long as there is enough space to move the head, it will print part after part to reduce movements.

#### What did you expect to happen?

I've expected the gcode viewer to show me how each part is being printer and in which order.

#### What happened instead?

Instead it showed a mix of all parts. Tracking whats being printed at the moment seemed to be correct, but moving to the next layer shows a mix of printed parts and parts which are not printed at the moment. It should - imho - show whats being printed next.

#### Did the same happen when running OctoPrint in safe mode?
Can't check right now.

#### Branch & Commit or Version of OctoPrint

1.3.5

#### Operating System running OctoPrint

Debian Stretch

#### Printer model & used firmware incl. version

cr-10 s4 marlin 1.1.6

#### Browser and Version of Browser, Operating System running Browser

google chrome 63.0.3239.9-1, debian unstable

#### Link to octoprint.log
http://paste.debian.net/997554/

#### Link to contents of terminal tab or serial.log
not needed.

#### Link to contents of Javascript console in the browser
http://paste.debian.net/997555/

#### Screenshot(s)/video(s) showing the problem:

I'll follow up with screenshots and a gcoe example as soon as my own server is up and running again. Jus twant to document it before I forget about it.

I have read the FAQ.

GitIssueBot:
Hi @bzed,

It looks like there is some **information missing** from your bug report that will be needed in order to solve the problem. Read the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) which will provide you with a template to fill out here so that your bug report is ready to be investigated (I promise I'll go away then too!).

If you did not intend to report a bug but wanted to **request a feature or brain storm** about some kind of development, please take special note of the title format to use as described in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md).

**Please do not abuse the bug tracker as a support forum** - if you have a question or otherwise need some kind of help or support refer to the [Mailinglist](https://groups.google.com/group/octoprint) or the [G+ Community](https://plus.google.com/communities/102771308349328485741) instead of here.

[...]

PR #4430, opened 2022-02-18
GCodeViewer: Replace AVLTree with double binary search

JoveToo:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR attempts to reduce memory usage of the gcodeviewer plugin. It does this by replacing the particularly inefficient implementation of an AVLTree with a double binary search.

For a 110.9MB Gcode file, I measured the following memory usages:

|  Code | Memory Usage |
|----------|---------------------:|
| AVLTree based | 4240.72 MB |
| Double Binary Search | 1439.81 MB |

This is a 65% memory reduction.

As an alternative, one could fix the AVLTree implementation but I think this is not worth it. I am not a JS expert but as far as I can tell, the implementation is a tree of objects. These objects do not just contain the required data for the tree but each node contains its own copies of all the object member functions. Rather wasteful?

I do not think it worth it to use the AVLTree. With a double binary search you do one binary search in all the layers and 1 binary search inside the relevant layer. Such searches on these small datasets are more than efficient enough. I do not feel any difference in speed. An added advantage is that the AVLTree does not need to be created and loading the gcode is actually faster.

Even should you fix the AVLTree, it will still use extra memory while this double binary search does not.

These is one open issue: if you look for a certain completion percentage, it will return the closest smaller one (instead of the closest larger one). I am not sure this is correct but this needs to be determined with someone better knowledge of this code base.

TODO: remove AVLTree.js from the source tree.

I loaded the same gcode files on both implementations.
I can do some extra tests with more gcode files to gather more statistics if required.


foosel:
> * It would be good to understand what I can expect here: is an update sent every gcode command? Every X commands? Every 0.01%?

Push updates from the backend are capped at every 0.5s max. If the client is too slow to process this, or if simply nothing new is happening, the interval can grow longer.

JoveToo:
> Push updates from the backend are capped at every 0.5s max. If the client is too slow to process this, or if simply nothing new is happening, the interval can grow longer.

Ok, thanks,  I'll just search in the next 128 gcode commands. This would be 7 loops.

JoveToo:
I have implemented the caching. It will speed up the look up but this means reducing the lookup from perhaps 30 loop iterations to 5 or so. I've made the lookahead value easy to change in the source. Since in most prints I've tried, it usually jumps ahead 1 or 2 gcode commands, I've set the value to 64.

An important change is that layers are no longer sorted by z-height. For this code to work, percentage must increase when layers and cmds increase. This was not necessarily the case after sorting in z-height. (For example, the purge line put down by my start gcode is 0.28 in height while my first layer was 0.24). This does mean that the first layer displayed upon loading the model will almost always be the purge lines. Less pretty. We could probably add code to recognize the purge lines and skip them...

foosel:
As I see commits coming in here - are you sure this should not be marked as WIP for now? 😅 I'd like to review it once you are done tinkering with it, not several times in the middle 😉

JoveToo:
This should be done now.

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
```diff
@@ -77,34 +80,116 @@ GCODE.gCodeReader = (function () {
...
+        if (cacheLastLayer !== undefined) {
+            if (
+                rendererModel[cacheLastLayer][0].percentage <= key &&
+                (!rendererModel[cacheLastLayer + 1] ||
+                    rendererModel[cacheLastLayer + 1][0].percentage > key)
+            ) {
+                bestLayer = cacheLastLayer;
+
+                var upper =
+                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
```
foosel, line 140: Am I missing something here or is this really just a confusing way to say

```suggestion
                var upper =
                    Math.min(cacheLastCmd + cacheLookAhead, rendererModel[bestLayer].length - 1);
```

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
```diff
@@ -77,34 +80,116 @@ GCODE.gCodeReader = (function () {
...
+        // just return last position to keep the last
+        // position on screen.
+        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};
+
+        var bestLayer = undefined;
+        var bestCmd = undefined;
+
+        // check if we are within cacheLookAhead distance of our last position
+        if (cacheLastLayer !== undefined) {
+            if (
```
foosel, line 132: This needs a comment explaining the meaning of the expression. Something like

```js
// command likely in the cached layer
```

I also can't help but think that there's way too much code repetition in this cached layer handling.

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
```diff
@@ -77,34 +80,116 @@ GCODE.gCodeReader = (function () {
...
+                var upper =
+                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
+                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
+                if (tmpresult < upper) bestCmd = tmpresult;
+            } else if (
+                rendererModel[cacheLastLayer + 1][0].percentage <= key &&
+                (!rendererModel[cacheLastLayer + 2] ||
+                    rendererModel[cacheLastLayer + 2][0].percentage > key)
+            ) {
+                bestLayer = cacheLastLayer + 1;
```
foosel, line 148:
```js
// command likely in the next layer
```

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
```diff
@@ -77,34 +80,116 @@ GCODE.gCodeReader = (function () {
...
+
+                var upper =
```
foosel, line 150: This needs a comment, I struggle to figure out the logic behind this and I don't even want to imagine how long that would take anyone to grasp in a few months time.

Best make sure this whole search block is commented in such a way that you don't have to understand the code to understand what happens and why.

foosel (changes requested):
All in all this looks great, but I'm a bit worried about a general lack of comments in some of the critical locations and the affect of that on long term maintenance, so if you could look at improving things on that front and also look at some of the other points that would be lovely!

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
JoveToo: Before I try to fix the code repetition by adding function calls: 

A binary search is O(log n), so the number of loop repetition of a full search is limited. We have like max 30000 layers and lets say 100000 commands in a layer. That means a full search is max 15 + 23 iterations of the binary search loop. 

How many loops equals 1 function call in cost? In C, it would be quite a few. (flushing registers to stack, fetching function code, restoring registers from cache after the call, ...).

We may just choose to drop the caching feature altogether and just do a full lookup each time. It would simplify the code considerably. I believe the current code is a speedup but I am less sure if I start adding function calls to that cache lookup code.

foosel: Probably a good idea because frankly, just off the top of my head I wouldn't be able to say how much the function calls here weigh vs the caching gains.

I'd lean towards "a lot" myself though. Note that I didn't necessarily say that this needed to be consolidated into a function, it just felt so utterly repetitive that I was wondering if there was *something* that could be done to make it more readable because frankly right now it takes a lot of mental energy to process.

JoveToo:
I have run some benchmarks. I've compared three version of the code:

Version 1: This version is the original version.
```javascript
    var searchInPercentageTree1 = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[middle][0].percentage <= key &&
                    (!rendererModel[middle + 1] ||
                        rendererModel[middle + 1][0].percentage > key)
                )
                    return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }

            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (rendererModel[layer][middle].percentage == key) return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            if (
                rendererModel[cacheLastLayer][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 1] ||
                    rendererModel[cacheLastLayer + 1][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer;

                var upper =
                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            } else if (
                rendererModel[cacheLastLayer + 1] &&
                rendererModel[cacheLastLayer + 1][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 2] ||
                    rendererModel[cacheLastLayer + 2][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer + 1;

                var upper =
                    (cacheLastCmd +
                        cacheLookAhead -
                        rendererModel[cacheLastLayer].length) %
                    rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined)
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        if (bestCmd === undefined)
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```
Version 2: This basically seperates out the isInLayer logic.
```javascript
   var searchInPercentageTree2 = function (key) {
        function isInLayer(layer, key)
        {
            // we are in this layer if the key is equal or larger
            // than the percentage of the first command and
            // smaller than the percentage of the first command
            // of the next layer, if it exists. if it does not
            // exists. if it does not exists, this is the last
            // layer and being large than the percentage of this
            // layer is enough.
            return (
               rendererModel[layer] &&
               rendererModel[layer][0].percentage <= key &&
               (!rendererModel[layer + 1] ||
                   rendererModel[layer + 1][0].percentage > key)
            ) 
        }

        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (isInLayer(middle, key)) return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }

            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (rendererModel[layer][middle].percentage == key) return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

       var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            // we only check two layers, if we have moved ahead more than 1 layer, we fall back to a forced full lookup.

            // are we in the same layer?
            if (isInLayer(cacheLastLayer, key)) {
                bestLayer = cacheLastLayer;

                // we need to determine upper bound because it might be in the next layer
                var upper = Math.min ((cacheLastCmd + cacheLookAhead), rendererModel[bestLayer].length - 1);

                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                // we only accept results that are below upper because if we are after the 
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            } 
            // or have we progressed to the next layer?
            else if (isInLayer(cacheLastLayer + 1, key)) {
                bestLayer = cacheLastLayer + 1;

                // determine upper bound, keeping in mind, we may have used some of cacheLookAhead in the previous layer and it still might be in the next layer.
                var upper = Math.min((cacheLastCmd + cacheLookAhead - rendererModel[cacheLastLayer].length), rendererModel[bestLayer].length - 1);

                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                // we only accept results that are below upper because if we are after the 
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined) {
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        }
        if (bestCmd === undefined) {
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );
        }

        // remember last position
        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```
And version 3: This doesn't do caching at all.
```javascript
   var searchInPercentageTree4 = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[middle][0].percentage <= key &&
                    (!rendererModel[middle + 1] ||
                        rendererModel[middle + 1][0].percentage > key)
                )
                    return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }

            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (rendererModel[layer][middle].percentage == key) return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        var bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```
I ran 2 tests, each on the three version of the code. The cached test, just runs the search once outside the loop then runs the same search repeatedly inside the loop. the random search generates a random number sequence and then runs this through each version. 
| Version | cached | random|
|-----------|-----------|-------------|
|1|12.469325153374234 |197.42484662576686|
|2|13.508179959100204|210.95245398773005|
|3|28.917689161554193|197.92433537832312 |

These are the times for 100000 loops...
So the function calls do matter. but only cause a 5% increase in the random lookups. and less than 10% in the cached lookups.
All in all, it probably is better to drop the caching alltogether. It doubles the speed but we are talking about 0.16 microseconds here, for quite some complexity.

JoveToo:
Btw, it is interesting that the random searches take so long while they should return the same as cached version 3. Either I am missing a bug in my code or this is entirely due to processor cache misses.

This is the main code that calls the functions above.
```javascript
   var searchInPercentageTree = function(key)
    {
         var i, starttime, stoptime;

         if (rendererModel === undefined) return undefined;

         testcounter++;

         // first 100 cached lookups.         
         searchInPercentageTree1(key);
         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree1(key);
         }
         stoptime = performance.now();
         cached1 += (stoptime - starttime);
         console.log ("Time for cached searchInPercentageTree1: " + (stoptime - starttime) + " " + cached1/testcounter);

         searchInPercentageTree2(key);
         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree2(key);
         }
         stoptime = performance.now();
         cached2 += (stoptime - starttime);
         console.log ("Time for cached searchInPercentageTree2: " + (stoptime - starttime) + " " + cached2/testcounter);

         searchInPercentageTree4(key);
         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree4(key);
         }
         stoptime = performance.now();
         cached4 += (stoptime - starttime);
         console.log ("Time for cached searchInPercentageTree4: " + (stoptime - starttime) + " " + cached4/testcounter);

         // now 100 random lookups;
         var randomNumbers = [];
         for (i=0; i < 100000; i++)
             randomNumbers.push(Math.random()*100);

         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree1(randomNumbers[i]);
         }
         stoptime = performance.now();
         random1 += (stoptime - starttime);

         console.log ("Time for random searchInPercentageTree1: " + (stoptime - starttime) + " " + random1/testcounter);

         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree2(randomNumbers[i]);
         }
         stoptime = performance.now();
         random2 += (stoptime - starttime);
         console.log ("Time for random searchInPercentageTree2: " + (stoptime - starttime) + " " + random2/testcounter);

         starttime = performance.now();
         for (i=0; i < 100000; i++)
         {
             searchInPercentageTree4(randomNumbers[i]);
         }
         stoptime = performance.now();
         random4 += (stoptime - starttime);
         console.log ("Time for random searchInPercentageTree4: " + (stoptime - starttime) + " " + random4/testcounter);

         console.log ("---------------------------------------------------------------");

         return searchInPercentageTree1(key);
    }
```

foosel:
Did some modifications of the three implementations by extracting the duplicated internal functions and making sure that non-exact percentages can match:

```js
    var isInLayer = function (layer, key) {
        // we are in this layer if the key is equal or larger
        // than the percentage of the first command and
        // smaller than the percentage of the first command
        // of the next layer, if it exists. if it does not
        // exists. if it does not exists, this is the last
        // layer and being large than the percentage of this
        // layer is enough.
        return (
            rendererModel[layer] &&
            rendererModel[layer][0].percentage <= key &&
            (layer == rendererModel.length - 1 ||
                rendererModel[layer + 1][0].percentage > key)
        );
    };

    var isCmd = function (layer, cmd, key) {
        return (
            rendererModel[layer] &&
            rendererModel[layer][cmd] &&
            (rendererModel[layer][cmd].percentage == key ||
                (rendererModel[layer][cmd].percentage <= key &&
                    (cmd == rendererModel[layer].length - 1 ||
                        rendererModel[layer][cmd + 1].percentage > key)))
        );
    };

    var searchInLayers = function (lower, upper, key) {
        while (lower < upper) {
            var middle = Math.floor((lower + upper) / 2);

            if (isInLayer(middle, key)) return middle;

            if (rendererModel[middle][0].percentage > key) {
                upper = middle - 1;
            } else {
                lower = middle + 1;
            }
        }
        return lower;
    };

    var searchInCmds = function (layer, lower, upper, key) {
        while (lower < upper) {
            var middle = Math.floor((lower + upper) / 2);

            if (isCmd(layer, middle, key)) return middle;

            if (rendererModel[layer][middle].percentage > key) {
                upper = middle - 1;
            } else {
                lower = middle + 1;
            }
        }
        return lower;
    };

    var searchInPercentageTreeCached = function (key) {
        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            if (
                rendererModel[cacheLastLayer][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 1] ||
                    rendererModel[cacheLastLayer + 1][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer;

                var upper =
                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            } else if (
                rendererModel[cacheLastLayer + 1] &&
                rendererModel[cacheLastLayer + 1][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 2] ||
                    rendererModel[cacheLastLayer + 2][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer + 1;

                var upper =
                    (cacheLastCmd +
                        cacheLookAhead -
                        rendererModel[cacheLastLayer].length) %
                    rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined)
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        if (bestCmd === undefined)
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeCachedFunc = function (key) {
        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            // we only check two layers, if we have moved ahead more than 1 layer, we fall back to a forced full lookup.

            // are we in the same layer?
            if (isInLayer(cacheLastLayer, key)) {
                bestLayer = cacheLastLayer;

                // we need to determine upper bound because it might be in the next layer
                var upper = Math.min(
                    cacheLastCmd + cacheLookAhead,
                    rendererModel[bestLayer].length - 1
                );

                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                // we only accept results that are below upper because if we are after the
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            }
            // or have we progressed to the next layer?
            else if (isInLayer(cacheLastLayer + 1, key)) {
                bestLayer = cacheLastLayer + 1;

                // determine upper bound, keeping in mind, we may have used some of cacheLookAhead in the previous layer and it still might be in the next layer.
                var upper = Math.min(
                    cacheLastCmd + cacheLookAhead - rendererModel[cacheLastLayer].length,
                    rendererModel[bestLayer].length - 1
                );

                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                // we only accept results that are below upper because if we are after the
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined) {
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        }
        if (bestCmd === undefined) {
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );
        }

        // remember last position
        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeUncached = function (key) {
        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        var bestCmd = searchInCmds(
            bestLayer,
            0,
            rendererModel[bestLayer].length - 1,
            key
        );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```

This is the outcome:

image

Now I'm wondering if calling functions in the parent scope is more expensive than local scoped ones.

foosel:
Four variants now, cached, cached with `is{InLayer|Cmd}`, uncached, uncached with `is{InLayer|Cmd}`:

```js
    var searchInPercentageTreeC = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[middle][0].percentage <= key &&
                    (middle == rendererModel.length - 1 ||
                        rendererModel[middle + 1][0].percentage > key)
                )
                    return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[layer][middle].percentage == key ||
                    (rendererModel[layer][middle].percentage <= key &&
                        (middle == rendererModel[layer].length - 1 ||
                            rendererModel[layer][middle + 1].percentage > key))
                )
                    return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            if (
                rendererModel[cacheLastLayer][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 1] ||
                    rendererModel[cacheLastLayer + 1][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer;

                var upper =
                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            } else if (
                rendererModel[cacheLastLayer + 1] &&
                rendererModel[cacheLastLayer + 1][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 2] ||
                    rendererModel[cacheLastLayer + 2][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer + 1;

                var upper =
                    (cacheLastCmd +
                        cacheLookAhead -
                        rendererModel[cacheLastLayer].length) %
                    rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined)
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        if (bestCmd === undefined)
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeCF = function (key) {
        function isInLayer(layer, key) {
            // we are in this layer if the key is equal or larger
            // than the percentage of the first command and
            // smaller than the percentage of the first command
            // of the next layer, if it exists. if it does not
            // exists, this is the last layer and being larger
            // than the start percentage of this layer is enough.
            return (
                rendererModel[layer] &&
                rendererModel[layer][0].percentage <= key &&
                (layer == rendererModel.length - 1 ||
                    rendererModel[layer + 1][0].percentage > key)
            );
        }

        function isCmd(layer, cmd, key) {
            return (
                (rendererModel[layer] &&
                    rendererModel[layer][cmd] &&
                    rendererModel[layer][cmd].percentage == key) ||
                (rendererModel[layer][cmd].percentage <= key &&
                    (cmd == rendererModel[layer].length - 1 ||
                        rendererModel[layer][cmd + 1].percentage > key))
            );
        }

        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (isInLayer(middle, key)) return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (isCmd(layer, middle, key)) return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            // we only check two layers, if we have moved ahead more than 1 layer, we fall back to a forced full lookup.

            // are we in the same layer?
            if (isInLayer(cacheLastLayer, key)) {
                bestLayer = cacheLastLayer;

                // we need to determine upper bound because it might be in the next layer
                var upper = Math.min(
                    cacheLastCmd + cacheLookAhead,
                    rendererModel[bestLayer].length - 1
                );

                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                // we only accept results that are below upper because if we are after the
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            }
            // or have we progressed to the next layer?
            else if (isInLayer(cacheLastLayer + 1, key)) {
                bestLayer = cacheLastLayer + 1;

                // determine upper bound, keeping in mind, we may have used some of cacheLookAhead in the previous layer and it still might be in the next layer.
                var upper = Math.min(
                    cacheLastCmd + cacheLookAhead - rendererModel[cacheLastLayer].length,
                    rendererModel[bestLayer].length - 1
                );

                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                // we only accept results that are below upper because if we are after the
                // calculated upper, it will always return upper.
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined) {
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        }
        if (bestCmd === undefined) {
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );
        }

        // remember last position
        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeU = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[middle][0].percentage <= key &&
                    (middle == rendererModel.length - 1 ||
                        rendererModel[middle + 1][0].percentage > key)
                )
                    return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[layer][middle].percentage == key ||
                    (rendererModel[layer][middle].percentage <= key &&
                        (middle == rendererModel[layer].length - 1 ||
                            rendererModel[layer][middle + 1].percentage > key))
                )
                    return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        var bestCmd = searchInCmds(
            bestLayer,
            0,
            rendererModel[bestLayer].length - 1,
            key
        );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeUF = function (key) {
        function isInLayer(layer, key) {
            // we are in this layer if the key is equal or larger
            // than the percentage of the first command and
            // smaller than the percentage of the first command
            // of the next layer, if it exists. if it does not
            // exists, this is the last layer and being larger
            // than the start percentage of this layer is enough.
            return (
                rendererModel[layer] &&
                rendererModel[layer][0].percentage <= key &&
                (layer == rendererModel.length - 1 ||
                    rendererModel[layer + 1][0].percentage > key)
            );
        }

        function isCmd(layer, cmd, key) {
            return (
                (rendererModel[layer] &&
                    rendererModel[layer][cmd] &&
                    rendererModel[layer][cmd].percentage == key) ||
                (rendererModel[layer][cmd].percentage <= key &&
                    (cmd == rendererModel[layer].length - 1 ||
                        rendererModel[layer][cmd + 1].percentage > key))
            );
        }

        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (isInLayer(middle, key)) return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (isCmd(layer, middle, key)) return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        var bestCmd = searchInCmds(
            bestLayer,
            0,
            rendererModel[bestLayer].length - 1,
            key
        );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```

Results:

image

Based on this I'm leaning towards either cached or uncached without `is{InLayer|Cmd}`. Uncached indeed might be best for the sake of a reduction in code complexity.

foosel:
Added a layer percentage lookup table (your idea from #4435, slightly refined):

```js
    var searchInPercentageTreeCL = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    layerPercentageLookup[middle][0] <= key &&
                    layerPercentageLookup[middle][1] > key
                )
                    return middle;

                if (rendererModel[middle][0].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[layer][middle].percentage == key ||
                    (rendererModel[layer][middle].percentage <= key &&
                        (middle == rendererModel[layer].length - 1 ||
                            rendererModel[layer][middle + 1].percentage > key))
                )
                    return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = undefined;
        var bestCmd = undefined;

        // check if we are within cacheLookAhead distance of our last position
        if (cacheLastLayer !== undefined) {
            if (
                rendererModel[cacheLastLayer][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 1] ||
                    rendererModel[cacheLastLayer + 1][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer;

                var upper =
                    (cacheLastCmd + cacheLookAhead) % rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, cacheLastCmd, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            } else if (
                rendererModel[cacheLastLayer + 1] &&
                rendererModel[cacheLastLayer + 1][0].percentage <= key &&
                (!rendererModel[cacheLastLayer + 2] ||
                    rendererModel[cacheLastLayer + 2][0].percentage > key)
            ) {
                bestLayer = cacheLastLayer + 1;

                var upper =
                    (cacheLastCmd +
                        cacheLookAhead -
                        rendererModel[cacheLastLayer].length) %
                    rendererModel[bestLayer].length;
                var tmpresult = searchInCmds(bestLayer, 0, upper, key);
                if (tmpresult < upper) bestCmd = tmpresult;
            }
        }

        // do a full search if the cache missed
        if (bestLayer === undefined)
            bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        if (bestCmd === undefined)
            bestCmd = searchInCmds(
                bestLayer,
                0,
                rendererModel[bestLayer].length - 1,
                key
            );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };

    var searchInPercentageTreeUL = function (key) {
        function searchInLayers(lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    layerPercentageLookup[middle][0] <= key &&
                    layerPercentageLookup[middle][1] > key
                )
                    return middle;

                if (layerPercentageLookup[middle][0] > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        function searchInCmds(layer, lower, upper, key) {
            while (lower < upper) {
                var middle = Math.floor((lower + upper) / 2);

                if (
                    rendererModel[layer][middle].percentage == key ||
                    (rendererModel[layer][middle].percentage <= key &&
                        (middle == rendererModel[layer].length - 1 ||
                            rendererModel[layer][middle + 1].percentage > key))
                )
                    return middle;

                if (rendererModel[layer][middle].percentage > key) {
                    upper = middle - 1;
                } else {
                    lower = middle + 1;
                }
            }
            return lower;
        }

        if (rendererModel === undefined) return undefined;

        // this happens when the print is stopped.
        // just return last position to keep the last
        // position on screen.
        if (key == null) return {layer: cacheLastLayer, cmd: cacheLastCmd};

        var bestLayer = searchInLayers(1, rendererModel.length - 1, key);
        var bestCmd = searchInCmds(
            bestLayer,
            0,
            rendererModel[bestLayer].length - 1,
            key
        );

        cacheLastLayer = bestLayer;
        cacheLastCmd = bestCmd;

        return {layer: bestLayer, cmd: bestCmd};
    };
```

```js
    var rebuildLayerPercentageLookup = function (m) {
        if (!m) return;

        var result = [];
        for (var i = 0; i < m.length - 1; i++) {
            if (!m[i]) {
                result[i] = [-1, -1];
                continue;
            }
            var start = m[i].length ? m[i][0].percentage : -1;

            var end = -1;
            for (var j = i + 1; j < m.length; j++) {
                if (m[j].length) {
                    end = m[j][0].percentage;
                    break;
                }
            }

            result[i] = [start, end];
        }
        result[m.length - 1] = [m[m.length - 1][0].percentage, 100];

        layerPercentageLookup = result;
    };
```

Added to the process like this:

``` js
        passDataToRenderer: function () {
            var m = model;
            //if (gCodeOptions["sortLayers"]) m = sortLayers(m);
            if (gCodeOptions["purgeEmptyLayers"]) m = purgeLayers(m);

            rendererModel = m;
            rebuildLayerPercentageLookup(m);

            GCODE.renderer.doRender(m, 0);
            return m;
        },
```

Now I get this:

image

I think lookup + uncached might be a good compromise here. Sadly I so far was unable to get the actual memory consumption out of Firefox. Chrome shows 1.7GB JS Memory for the tab after loading the 106MB file. I need to find some way to better profile this. Had some tremendous issues with Chrome though, *really* slow with the file. Need to cross check if that also happens with 1.7.3.

JoveToo:
Thank you for taking the time to look into this so thoroughly :+1: 

I got my Firefox numbers by taking a memory snapshot.  It takes a while but the information is useful. It gives you a good indication how much memory is used and what it is used for. This is how I initially found the problem with the AVLTree implementation. (Gigabytes of function instantiations).

I will make an implementation with this lookup and uncached.  The reason #4435 contains the lookup list is that I otherwise have to decompress every layer to be able to read the percentage of the first command.. I like your slightly altered list and I will use that, it is safer.

Btw, since the `searchInCmds` logic is using `Math.floor` to calculate `middle`, the only way `middle == rendererModel[layer].length - 1` can happen is when `lower == upper == rendererModel[layer].length - 1` and then we have already exited the loop. The extra statement that compares with the percentage of the next command can indeed save loop iterations!

foosel:
Just spotted a major issue with the binary search vs how layers are being parsed. The 106MB GCODE file I was testing with showed some weird layer order: 0.6, then 0.2, then 0.4. Excerpt with which to reproduce attached:

[test.gcode.txt](https://github.com/OctoPrint/OctoPrint/files/8140665/test.gcode.txt)

The problem is that there's an initial z lift right after the purge line which goes to 0.6, after which the head moves to 0.2 and starts printing the actual model. The way this ends up in the layer structure is with the first layer at 0.6, the first two commands in there, then 0.2 gets filled as *second* layer, then 0.4 as *third* layer, then data lands in 0.6 as *first* layer again, leading up to percentage of 99.9. So We have a first layer that goes from almost 0 to 99, a second also from almost 0 to 60 or something... stuff is completely out of order. Wasn't an issue before, I'd expect some big problems finding the whole 0.6 layer now due to this as it's basically become invisible. 

So that's a bug in creating the data structure that has to be solved before this can be merged. Stuff is tricky.

foosel:
So I managed to fix that, and also fixed #2243 in the process, as I just tested successfully 😬 I think that's now ready to get merged. RC phase might unearth some more minor bugs, and possibly force us to roll things back (I threw a ton of my test files against it, but as always there's probably going to be some corner cases I didn't yet consider), but at least for now this looks and feels ready.

PR #4433, opened 2022-02-19
⚡  Don't load FontAwesome twice

cp2004:

In the preload templates, an extra `?v=version` was added on the end of the URL. In the CSS file however, this was not used so the resource was fetched again, unnecessarily. The version number is embedded in the path, so it will be refetched if that changes.

image

This has actually been logging to the warnings all the time, but I never actually looked at why:
image

Additionally, both the `woff2` and `woff` files are not required by everyone - the woff2 files are preferred, with woff only used for those that don't support it. And [woff2 has a 96% compatibility](https://caniuse.com/woff2), everything except IE supports it, and it will be loaded automatically if required by the css file.

Manually in the browser

With this change, the number of FontAwesome related requests drops to a nice little 3:
image

I was testing out FontAwesome 6, so I was poking in these kinds of areas... 👀 

#### Further Notes

The final unnecessary loading is of FontAwesome 3, but that is for backwards compatibility with plugins. Should be axed at some point IMO.


foosel:
Good find! Also reminded me that I've been meaning to take a look at those warnings myself for quite some time 😬

PR #4437, opened 2022-02-23
Support two-digit flavors for G00 G01 G02 G03

vector76:
This adds support for these four commands: G00 G01 G02 and G03, treating them the same as G0, G1, G2, and G3.

Many CAM processors generate two-digit movement commands.  These are legal movement commands but the viewer is currently blind to them.  In my opinion, this is a bug.  (Evidently slicers never generate these, or this would have already been an issue.)

This does not fully solve rendering of non-extruding toolpaths, but it enables a workaround in combination with M101 added to the preamble.  Without recognition of two-digit movement commands, a more involved preprocessor is necessary.

A reviewer should be able to validate it by inspection, but I also made the modification locally while I was checking that M101 would produce a rendering.  During testing is also when I discovered that G02 and G03 are necessary, and G00 and G01 are not sufficient.  Arcs are much more common in CAM-generated paths, compared to sliced polyhedra.

A more complete CNC viewer plugin could overlap in functionality, but IMO failing to recognize G00 and G01 is _also_ a bug, and this PR targets the bug specifically and brings none of the other CNC considerations.

### DIFF
--- a/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
+++ b/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
@@ -380,7 +380,7 @@ var doParse = function () {
 
         var log = false;
 
-        if (/^(?:G0|G1|G2|G3)(\.\d+)?\s/i.test(line)) {
+        if (/^(?:G0|G1|G2|G3|G00|G01|G02|G03)(\.\d+)?\s/i.test(line)) {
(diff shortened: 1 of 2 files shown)


foosel:
Thanks! I took the liberty to add it to the gcode analysis in the backend as well.

vector76:
Awesome.  Thank you!

PR #4438, opened 2022-02-24
Add rust compiler to the development environment installation list.

JoveToo:
Rust is necessary for libcst of codemods in the pre-commit.


JoveToo:
This doesn't solve the problem.

PR #4449, opened 2022-03-09
GCodeViewer Only clear state on event when idle

JoveToo:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

With the current code, when downloading and splitting the file, the GCodeViewer status is cleared on every update event. This is not only unnecessary, it prevents the GCodeViewer progress bar to show the state "downloading" and "splitting lines".

The state should only be cleared when the GCodeViewer is in "idle" state.

Load large gcode file.

### DIFF
--- a/src/octoprint/plugins/gcodeviewer/static/js/gcodeviewer.js
+++ b/src/octoprint/plugins/gcodeviewer/static/js/gcodeviewer.js
@@ -614,7 +614,7 @@ $(function () {
                 }
                 self.errorCount = 0;
             } else {
-                self.clear();
+                if (self.status === "idle") self.clear();
                 if (
                     data.job.file.path &&
                     data.job.file.origin !== "sdcard" &&


cp2004:
>  it prevents the GCodeViewer progress bar to show the state "downloading" and "splitting lines".

I've seen both these states on files I've tried the gcode viewer with, so it can't completely prevent it?

JoveToo:
> I've seen both these states on files I've tried the gcode viewer with, so it can't completely prevent it?

It prevents it here on Firefox and maintenance branch. The states appear for like 0.5 seconds until a new event arrives.

cp2004:
From a quick look through the rest of the code - I can see how the 'Downloading' state could be interrupted (the filename is not stored until download complete), not yet the splitting one. Doesn't mean it doesn't happen, just means I haven't worked it out from reading it, and don't have time to finish right now.

JoveToo:
I worked it out from reading the code that this was wrong, at least for the Downloading state. I haven't tried figuring it out for the splitting case, either. 

In any case, with this change, both states are now visible on every file load.

JoveToo:
I am not entirely sure how promise completion handlers are called but this particular completion handler is slow. It sets "Splitting" status then calls gCodeReader.loadFile, which will do the splitting. Only after that, the loadedFilepath/loadedfileDate variables are set. So if an event comes between setting "splitting" and before the splitting ends, it will get cleared. *if* these are truely async events.

PR #4467, opened 2022-03-18
Tiny documentation typo, timeouts -> timeout

vector76:
Sorry for the annoyance.  I saw this small error in the documentation on docs.octoprint.org and this looked like the source.

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -654,7 +654,7 @@ class Settings(object):
         serial:
             port: "/dev/ttyACM0"
             baudrate: 250000
-            timeouts:
+            timeout:
                 communication: 20.0
                 temperature: 5.0
                 sdStatus: 1.0


vector76:
Oops, wrong branch.

cp2004:
This isn't the documentation you've edited, it's the actual source code of the settings, which should definitely not be changed.

The docs are here https://github.com/OctoPrint/OctoPrint/blob/master/docs/configuration/config_yaml.rst, the change should be going the other way correcting the typo in the docs.

Since it just docs for existing versions, if you want to do it again you can make it against master, and it will be allowed through.

vector76:
I had already opened a corresponding PR against a different branch https://github.com/vector76/OctoPrint/pull/1 since this seems to be the wrong branch.

The document in question is this: https://docs.octoprint.org/en/master/modules/settings.html#octoprint.settings.Settings where the example contains 'timeouts' and then is later referred to as 'timeout' which would not work.

It appears that the source for that documentation is here: https://github.com/OctoPrint/OctoPrint/blob/master/docs/modules/settings.rst and that file contains nothing.  I had to do some digging and the best I could figure out is that it is auto-generated from the source code.

I thought I was editing a doc string in the source code that is crunched to produce the documentation, and has zero effect other than producing documentation.

If this is incorrect, I'm happy to be educated.  If so, I would suggest commenting on the other PR.

cp2004:
Sorry, I didn't initially spot where this docstring ended up, I was thinking the typo was in OctoPrint's actual config docs not in the example, my mistake there.

> I had already opened a corresponding PR against a different branch https://github.com/vector76/OctoPrint/pull/1 since this seems to be the wrong branch.

It looks like you've created that PR against your own repository, not the OctoPrint one, so we can't do anything with it at the moment.

The docstring can be edited to change that example, but note it is not representative of OctoPrint's *actual* settings keys or structure, it is just an example of how the settings module could be used with example settings. If that one line is edited, the rest of the docstring also needs editing to match it as `timeouts` is used across the example.

Since it is editing the docstring which is inside the source code, maintenance is the right branch.

vector76:
hahaha PR against myself!  Clearly I am not a pro at this.

I understand, it's just an example that happens to match the actual config variables (at the moment).  Elsewhere in the example `timeout` is used, not `timeouts`, so the proposed change brings it into agreement with the rest of the example.  The internal inconsistency within the example is the only problem I'm trying to fix.

Somewhat my fault for not explaining in the original description of the PR.

cp2004:
> Elsewhere in the example timeout is used, not timeouts, so the proposed change brings it into agreement with the rest of the example. The internal inconsistency within the example is the only problem I'm trying to fix.

I really badly failed at reading about 4 times trying to look at this PR's content to see what it was doing. It's the end of the week so I'm going to use that as an excuse for now.

vector76:
No worries.  I take much of the blame for not having a clearer explanation.  This one is much better: https://github.com/OctoPrint/OctoPrint/pull/4468

PR #4468, opened 2022-03-18
Tiny documentation typo, timeouts -> timeout

vector76:
Within the internal module documentation for settings
 https://docs.octoprint.org/en/master/modules/settings.html#octoprint.settings.Settings
there is an inconsistency between the example definition and how it is referred to:
image

This documentation appears to be automatically generated from an autodoc string in the source code.

This PR changes "timeouts" to "timeout" in the autodoc string to bring them into agreement.  This also happens to match the actual configuration variable but that is somewhat beside the point.

### DIFF
--- a/src/octoprint/settings.py
+++ b/src/octoprint/settings.py
@@ -654,7 +654,7 @@ class Settings(object):
         serial:
             port: "/dev/ttyACM0"
             baudrate: 250000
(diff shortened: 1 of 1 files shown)


cp2004 (approved):
Apologies for the delay in merging this.

PR #4483, opened 2022-04-07
Don't block unused keys when keyboard control is active

j7126:
  * [x] Your PR targets OctoPrint's maintenance branch
    if necessary!
    have added unit tests
    nothing broke

Currently when they keyboard controls are active by hovering over the webcam, all key events will be captured and prevented. This means that if the user wants to use some other keyboard shortcut (for example ctrl+r or f5 to refresh the page) while they happen to have keyboard controls active, they cannot do so.

This PR allows any other keys that are not used by the keyboard controls to still do their normal action. 
I have also added a case for the tab key to prevent it from un-focusing the webcam and un-intentionally deactivating keyboard controls. 

This can be tested this by trying other keyboard shortcuts (such as to refresh the page) while the keyboard controls are active, and they should now work.

Additionally all of the keyboard controls still function correctly.

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -728,9 +728,12 @@ $(function () {
                     // z home
                     button = $("#control-zhome");
                     break;
-                default:
+                case 9: // prevent tab key from removing focus from webcam
                     event.preventDefault();
                     return false;
+                default:
+                    // don't prevent other keys
+                    return true;
             }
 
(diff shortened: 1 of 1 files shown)


foosel:
FYI, I've seen this, I have it in the backlog, I just have my hands full even more than usual this month, so it'll be a while until I come around to reviewing this.

cp2004 (approved):
👍 Thanks for the PR.

Not entirely sure what was wrong with the failing checks, re-ran them and it worked fine.

PR #4484, opened 2022-04-07
⏫ Upgrade to FontAwesome 6

cp2004:

Updates Fontawesome from V5 to V6.

FA6 claims to be backwards compatible with V4 (with the included V4 shims), so this is a safe upgrade to do. It is (apparently) completely compatible with V5 with no extra changes necessary.

**However** there have been some changes, but I would consider them minor. For example, over the lifetime of V5 a couple of the brand icons requested to be removed, so they will no longer work. I don't think this is an issue, as it is very specific and the brands asked to be removed anyway. The version of V5 OctoPrint was using was quite up to date anyway.

I also added in a compatibility section for the icons in the sidebar headers. So far they were fixed to using `fa fa-<icon>`, the V4 style. Now, if the icon specified starts with `fa` OctoPrint will not add the prefix automatically, so you could use `fas`, `far`, or `fab` if desired, to use a V5+ icon - only the V4 ones are shimmed when used with `fa`.

I changed the printer notifications icon before realising I hadn't included the V4 shims. It doesn't actually need changing and works as-is, but now it is fixed it was the last icon using V4 styles that I know of.

*Finally*, I removed the plus icon from the plugin manager - apparently it was broken previously, and I liked the look of the button without it, so thought I would keep it out. It randomly started working again when FA6 was used.


cp2004:
One thing I didn't mention - the supported browsers have now changed, because they dropped woff support and only have woff2 and ttf formats.

This primarily impacts internet explorer, dropping support for that.

 https://fontawesome.com/docs/web/dig-deeper/browser-support

Caniuse data:

foosel:
IE has been on the "do not use" list of OctoPrint since 2013, I think if people still cling to that its their problem, not ours.

Issue #4326, opened 2021-11-29
[Request] Ability to ignore Printer Notifications based on regex


dburr:
### Is your feature request related to a problem? Please describe.

I use the [Detailed Progress](https://plugins.octoprint.org/plugins/detailedprogress/) plugin to update my printer's display with the estimated time to completion, etc. It does this by periodically injecting `M117` commands to the printer. Unfortunately since `M117` commands are echoed back as notification actions, this results in the "Printer Notifications" list getting filled with useless/trivial notifications such as this.
![Screenshot from 2021-11-28 15-58-37](https://user-images.githubusercontent.com/1237762/143791639-23e5d963-af98-4b85-92eb-87679234cd8f.png)

It would be even worse if I had popups enabled, as then I would have these trivial messages constantly popping up. I have actually wanted to enable popups so that I can be alerted of truly important/noteworthy things (filament runout, thermal runaway, etc.) but enabling popups with the situation like this would be infuriating.

### Describe the solution you'd like

It would be nice to be able to conditionally ignore notifications from the printer based on a regex. So, for example, to filter out the useless messages in my screenshot above, I would use regexes like:

`^\d+%`
`^Layer \d+\/\d+`
`^Done In .*`
`^(Mon|Tue|Wed|Thu|Fri|Sat|Sun).*`

### Describe alternatives you've considered

Obviously I could disable the Detailed Progress plugin, but then I would miss out on that useful information being shown on the printer (which I do check periodically)

foosel:
> Unfortunately since `M117` commands are echoed back as notification actions

TBH, that sounds more like the firmware is abusing the printer notification feature. It's meant for notifications that are NOT triggered by the host itself but would otherwise require the user to stand in front of the printer to even have a knowledge of. Sadly this feature has seen a TON of abuse like this ever since I added it to OctoPrint, to a point where I wonder whether it was a good idea to add it at all, with firmware happily spamming its users through it.

We can look into a filter here, but **please** also report this to your firmware issue tracker so that may look into fixing this constant spam on their end as well.

jneilliii:
Marlin has been doing this for a while now. I think it may have been utilizing this to support the newer touch screen interfaces.

foosel:
[...]

PR #4513, opened 2022-05-18
Adding ability for users to filter notifications bassed on regular ex…

DShenkle:
…pressions

  * [*] Your changes are not possible to do through a plugin and relevant
  * [*] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
  * [*] Your PR targets OctoPrint's devel branch if it's a completely
  * [*] Your PR was opened from a custom branch on your repository
  * [*] Your PR only contains relevant changes: no unrelated files,
    if necessary!
  * [*] Your changes follow the existing coding style
  * [*] If your changes include style sheets: You have modified the
    lessc)
  * [*] You have tested your changes (please state how!) - ideally you
    have added unit tests

Used virtual printer commands to generate notifications such as !!DEBUG:send //action:notification 27%. Entered regular expressions into new settings field and confirmed that if a regular expression such as "^\d+%" is entered the notification is ignored.

  * [*] You have run the existing unit tests against your changes and
    nothing broke
  * [*] You have added yourself to the AUTHORS.md file :)

This pull request adds the ability for users to filter out unwanted printer notifications.

This was tested by using the virtual printer to generate various notifications such as 

!!DEBUG:send //action:notification 27%

and then used the settings system to enter regular expressions for filtering out these notifications in this case 

^\d+%

#4326

image

### DIFF
--- a/src/octoprint/plugins/action_command_notification/__init__.py
+++ b/src/octoprint/plugins/action_command_notification/__init__.py
@@ -1,6 +1,7 @@
 __license__ = "GNU Affero General Public License http://www.gnu.org/licenses/agpl.html"
 __copyright__ = "Copyright (C) 2020 The OctoPrint Project - Released under terms of the AGPLv3 License"
 
+import re
 import time
 
 import flask
@@ -60,7 +61,7 @@ class ActionCommandNotificationPlugin(
     # ~ SettingsPlugin
 
     def get_settings_defaults(self):
-        return {"enable": True, "enable_popups": False}
+        return {"enable": True, "enable_popups": False, "filters": ""}
 
     # ~ SimpleApiPlugin
 
@@ -120,6 +121,12 @@ class ActionCommandNotificationPlugin(
             return
 
         message = parameter.strip()
+
+        filters = self._settings.get(["filters"]).split("\n")
+        for filter in filters:
+            if re.fullmatch(filter, message):
+                return
+
         self._notifications.append((time.time(), message))
         self._plugin_manager.send_plugin_message(self._identifier, {"message": message})
 
--- a/src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
+++ b/src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
@@ -16,4 +16,11 @@
             </label>
         </div>
     </div>
+    <div class="control-group">
+        <label class="control-label" for="settings-serialAdditionalPorts">{{ _('Notification Filters') }}</label>
+        <div class="controls">
+            <textarea rows="4" class="block" id="settings-serialAdditionalPorts" data-bind="value: settings.plugins.action_command_notification.filters"></textarea>
+            <span class="help-inline">{{ _("Use this to define filters to hide unwanted notifications. One regular expression per line. Notifications that match a regular expression will not be shown.") }}</span>
+        </div>
+    </div>
 </form>


src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -120,6 +121,12 @@ def action_command_handler(self, comm, line, action, *args, **kwargs):
             return
 
         message = parameter.strip()
+
+        filters = self._settings.get(["filters"]).split("\n")
```
cp2004, line 125: For performance, would it be better to pre-fetch the filters and set it as an attribute on the class? Like `self.filters` or something simple. It would save looking them up in the settings for every line that is received. Probably only a small difference, but the less work that is done on every line then the less likely stuttering issues due to slow processing will be?

foosel: To take it further (since the `re.fullmatch` below will have to do the work of compiling the regex for every single line too, which is a big red flag): fetch once on start (and after settings save), compile into regex (`re.compile`), us THAT on every received `action:notification` line.

src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
```diff
@@ -16,4 +16,11 @@
             </label>
         </div>
     </div>
+    <div class="control-group">
+        <label class="control-label" for="settings-serialAdditionalPorts">{{ _('Notification Filters') }}</label>
+        <div class="controls">
+            <textarea rows="4" class="block" id="settings-serialAdditionalPorts" data-bind="value: settings.plugins.action_command_notification.filters"></textarea>
+            <span class="help-inline">{{ _("Use this to define filters to hide unwanted notifications. One regular expression per line. Notifications that match a regular expression will not be shown.") }}</span>
```
foosel, line 23: Should probably get a link to the Python regex docs, as not everyone would know how to use this otherwise.

foosel (changes requested):

DShenkle:
Good feedback from everyone! 

I just pushed changes that should hopefully address these issues.

src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -56,11 +58,13 @@ def get_assets(self):
     def on_event(self, event, payload):
         if event == Events.DISCONNECTED:
             self._clear_notifications()
+        elif event == Events.SETTINGS_UPDATED or event == Events.STARTUP:
+            self._configure_filters()
```
foosel, line 62: I wouldn't do this via event handling (runs concurrently to the rest of the platform), but rather ensure it really is done before further actions by hooking into `initialized` and `on_settings_save`.

src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -120,6 +124,11 @@ def action_command_handler(self, comm, line, action, *args, **kwargs):
             return
 
         message = parameter.strip()
+
+        for filter in self._filters:
+            if filter.fullmatch(message):
```
foosel, line 129: Hm... I'm pretty sure that in other places we use `match` behaviour (so, partial match). I think we should do that here as well for reasons of consistency with other filter settings in OctoPrint. The user can require a full match by using `^` and `$` markers in their patterns.

src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -130,6 +139,13 @@ def _clear_notifications(self):
         self._plugin_manager.send_plugin_message(self._identifier, {})
         self._logger.info("Notifications cleared")
 
+    def _configure_filters(self):
+        filters = self._settings.get(["filters"]).split("\n")
+
+        self._filters.clear()
+        for filter in filters:
+            self._filters.append(re.compile(filter))
```
foosel, line 147: This gives you a brief period where the filters are completely unset. If we didn't have to think about error handling here, I'd suggest to go for a more atomic (and also IMHO more pythonic) approach like this:

```suggestion
        self._filters = [re.compile(filter) for filter in filters]
```

However, a broken regex might be configured by the user and throw an exception here during compile, that needs to be handled gracefully. 

In any case, as suggested above, we should only end up with *one* regex in the backend.

src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
```diff
@@ -16,4 +16,14 @@
             </label>
         </div>
     </div>
+    <div class="control-group">
+        <label class="control-label" for="settings-serialAdditionalPorts">{{ _('Notification Filters') }}</label>
```
foosel, line 20: `for` is wrong.

src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
```diff
@@ -16,4 +16,14 @@
...
+        <div class="controls">
+            <textarea rows="4" class="block" id="settings-serialAdditionalPorts" data-bind="value: settings.plugins.action_command_notification.filters"></textarea>
```
foosel, line 22: `id` is wrong.

foosel (changes requested):
Do we really need multiple filters? Why don't we rather make the user use regexes to their full potential and create *one* regex with optional `|` chaining? That makes all our processing here on the backend much easier, as we only have *one* regex we need to compile *once* and also only match *once* per incoming notification line.

Example:

`sample notification|other sample notification|foo` will match `sample notification`, `other sample notification` but not `the printer is on fire`

foosel:
@DShenkle are you still looking into this?

foosel:
No response for a quarter of the year now, so I guess not :/

Issue #4040, opened 2021-03-05
[Request] Record last_fan_speed like last_position and last_temperature


jantman:
### Is your feature request related to a problem? Please describe.

I'm trying to work on a print-pausing feature using gcode scripts. We have access to the last temperatures (``pause_temperature`` in the Jinja context) and last positions (``pause_position``), but not the fan speed when paused.

### Describe the solution you'd like

I'd like a ``pause_fan_speed`` variable available in the Jinja context, like the temperature and position, so that I can also pause and resume the fan.

### Describe alternatives you've considered

For the time being, I'm just not changing the fan speed when paused. But this isn't optimal.

For context: I've had a number of short power outages recently that ruined long-running prints. I've now put my OctoPi and printer on a 1000VA UPS, but that runs down *very* quickly. So, my goal is to react to the UPS being on-battery for a certain length of time by pausing the print. That's "simple", and I have it working via the API, along with stopping all heating. But the fan still runs, and I'd like to be able to eek out every last bit of runtime that I can.

jneilliii:
Couldn't this be implemented with a single file plugin like in the example [here](https://github.com/OctoPrint/Plugin-Examples/blob/master/gcode_script_variables.py)? It would have to add a gcode phasing hook as well to keep track of the last fan speed sent to the printer to store internally to provide to the script variable return I'm guessing. The largest problem here is that this doesn't take into account any manual adjustments made on the printer panel.

jneilliii:
So I tool a stab at it and couldn't figure out how the variables work with teh scripts callback, but I was able to get the fam speed to prefix teh beforePrintStart script to achieve the same thing. You can move from prefix to postfix if you want it to set fan speed at the end of the script rather than the end of the script.

```
# coding=utf-8
from __future__ import absolute_import

import octoprint.plugin
import re


class FanSpeedScriptVariable(octoprint.plugin.StartupPlugin):
	def __init__(self):
		self.last_fan_speed = 0
		self.fan_speed_match = re.compile("^M106.+?S(\d+)")

	def gcode_script_variables(self, comm, script_type, script_name, *args, **kwargs):
		if not script_type == "gcode" or not script_name == "beforePrintResumed":
			return None

		prefix = "M106 S{}".format(self.last_fan_speed)
		postfix = None
		return prefix, postfix

	def gcode_callback(self, comm_instance, phase, cmd, cmd_type, gcode, subcode=None, tags=None, *args, **kwargs):
[...]

PR #4531, opened 2022-05-28
Record last fan speed

FedericoNembrini:
Implementation of #4040 
This PR adds last_fanspeed, pause_fanspeed and cancel_fanspeed, so we can pause and resume the fan directly from GCode scripts.

#### Task
* [x] Add self.last_fanspeed, self.pause_fanspeed, self.cancel_fanspeed to MachineCom object
* [x] Inject into
https://github.com/OctoPrint/OctoPrint/blob/5f7679b11baafc0bb58861c462b9e4568041080b/src/octoprint/util/comm.py#L1268-L1361
* [x] Record on pause/cancel in
https://github.com/OctoPrint/OctoPrint/blob/5f7679b11baafc0bb58861c462b9e4568041080b/src/octoprint/util/comm.py#L2516-L2528
* [x] Add _gcode_M106_sent to parse and set fan speed when sent to printer
* [x] Update docs

Tested with virtual printer, adding in pause and cancel GCode scripts:
`Fan speed when paused: {{ pause_fanspeed.fanspeed }}`
`Fan speed when canceled: {{ cancel_fanspeed.fanspeed }}`

#4040

#### Further notes
In the settings, within Behaviour, there are the settings 'Log potision on pouse' and 'Log position on cancel'. Would it be appropriate to indicate that these settings are the same as those for the fan speed?

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -429,6 +429,21 @@ class TemperatureRecord:
             return actual, target
 
 
+class FanspeedRecord:
+    def __init__(self):
+        self._fanspeed = None
+
+    def copy_from(self, other):
+        self._fanspeed = other.fanspeed
+
+    def set_speed(self, speed):
+        self._fanspeed = speed
+
+    @property
+    def fanspeed(self):
+        return self._fanspeed
+
+
 class MachineCom:
     STATE_NONE = 0
     STATE_OPEN_SERIAL = 1
@@ -759,6 +774,10 @@ class MachineCom:
         self.last_position = PositionRecord()
         self.pause_position = PositionRecord()
         self.cancel_position = PositionRecord()
+
+        self.last_fanspeed = FanspeedRecord()
+        self.pause_fanspeed = FanspeedRecord()
+        self.cancel_fanspeed = FanspeedRecord()
 
         self._record_pause_data = False
         self._record_cancel_data = False
@@ -1299,6 +1318,7 @@ class MachineCom:
                 "printer_profile": self._printerProfileManager.get_current_or_default(),
                 "last_position": self.last_position,
                 "last_temperature": self.last_temperature.as_script_dict(),
+                "last_fanspeed": self.last_fanspeed,
             }
         )
 
@@ -1307,6 +1327,7 @@ class MachineCom:
                 {
                     "pause_position": self.pause_position,
                     "pause_temperature": self.pause_temperature.as_script_dict(),
+                    "pause_fanspeed": self.pause_fanspeed,
                 }
             )
         elif scriptName == "afterPrintCancelled":
@@ -1314,6 +1335,7 @@ class MachineCom:
                 {
                     "cancel_position": self.cancel_position,
                     "cancel_temperature": self.cancel_temperature.as_script_dict(),
+                    "cancel_fanspeed": self.cancel_fanspeed,
                 }
             )
 
@@ -2616,6 +2638,7 @@ class MachineCom:
                             self._record_pause_data = False
                             self.pause_position.copy_from(self.last_position)
                             self.pause_temperature.copy_from(self.last_temperature)
+                            self.pause_fanspeed.copy_from(self.last_fanspeed)
                             self._pause_preparation_done()
 
                         if self._record_cancel_data:
@@ -2623,6 +2646,7 @@ class MachineCom:
                             self._record_cancel_data = False
                             self.cancel_position.copy_from(self.last_position)
                             self.cancel_temperature.copy_from(self.last_temperature)
+                            self.cancel_fanspeed.copy_from(self.last_fanspeed)
                             self._cancel_preparation_done()
 
                         self._callback.on_comm_position_update(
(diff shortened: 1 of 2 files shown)


src/octoprint/util/comm.py
```diff
@@ -429,6 +429,21 @@ def _to_new_tuple(cls, current, actual, target):
...
+        self._fanspeed = other.fanspeed
+
+    def set_speed(self, speed):
+        self._fanspeed = speed
+
+    @property
+    def fanspeed(self):
+        return self._fanspeed
+
+
```
foosel, line 446: No need for that :) For the temperature records, we do this since we have to keep track of actual & targets of multiple heaters. Here we only track one single integer value, so we might as well use that instead of a our new type.

src/octoprint/util/comm.py
```diff
@@ -760,6 +775,10 @@ def __init__(
         self.pause_position = PositionRecord()
         self.cancel_position = PositionRecord()
 
+        self.last_fanspeed = FanspeedRecord()
+        self.pause_fanspeed = FanspeedRecord()
+        self.cancel_fanspeed = FanspeedRecord()
```
foosel, line 780:
```suggestion
        self.last_fanspeed = None
        self.pause_fanspeed = None
        self.cancel_fanspeed = None
```

src/octoprint/util/comm.py
```diff
@@ -2616,13 +2638,15 @@ def convert_line(line):
                             self._record_pause_data = False
                             self.pause_position.copy_from(self.last_position)
                             self.pause_temperature.copy_from(self.last_temperature)
+                            self.pause_fanspeed.copy_from(self.last_fanspeed)
```
foosel, line 2641:
```suggestion
                            self.pause_fanspeed = self.last_fanspeed
```

src/octoprint/util/comm.py
```diff
@@ -2616,13 +2638,15 @@ def convert_line(line):
...
                             self._pause_preparation_done()
 
                         if self._record_cancel_data:
                             reason = "cancel"
                             self._record_cancel_data = False
                             self.cancel_position.copy_from(self.last_position)
                             self.cancel_temperature.copy_from(self.last_temperature)
+                            self.cancel_fanspeed.copy_from(self.last_fanspeed)
```
foosel, line 2649:
```suggestion
                            self.cancel_fanspeed = self.last_fanspeed
```

src/octoprint/util/comm.py
```diff
@@ -5371,6 +5395,23 @@ def _gcode_G4_sent(
...
+    def _gcode_M106_sent(
+        self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs
+    ):
+        s_match = regexes_parameters["intS"].search(cmd)
+
+        _fanspeed = 0
+        if s_match:
+            _fanspeed = int(s_match.group("value"))
+
+        self.last_fanspeed.set_speed(_fanspeed)
```
foosel, line 5407:
```suggestion
        self.last_fanspeed = _fanspeed
```

src/octoprint/util/comm.py
```diff
@@ -5371,6 +5395,23 @@ def _gcode_G4_sent(
...
+
+    def _gcode_M107_sent(
+        self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs
+    ):
+        _fanspeed = 0
+        self.last_fanspeed.set_speed(_fanspeed)
```
foosel, line 5413:
```suggestion
        self.last_fanspeed = 0
```

src/octoprint/util/comm.py
```diff
@@ -5371,6 +5395,23 @@ def _gcode_G4_sent(
         self._timeout = self._get_new_communication_timeout() + _timeout
         self._dwelling_until = time.monotonic() + _timeout
 
+    def _gcode_M106_sent(
+        self, cmd, cmd_type=None, gcode=None, subcode=None, *args, **kwargs
+    ):
+        s_match = regexes_parameters["intS"].search(cmd)
+
+        _fanspeed = 0
```
foosel, line 5403: Just `fanspeed` please, it's not a protected instance field but just a function local variable.

src/octoprint/util/comm.py
```diff
@@ -1299,6 +1318,7 @@ def _getGcodeScript(self, scriptName, replacements=None):
                 "printer_profile": self._printerProfileManager.get_current_or_default(),
                 "last_position": self.last_position,
                 "last_temperature": self.last_temperature.as_script_dict(),
+                "last_fanspeed": self.last_fanspeed,
```
foosel, line 1321: This would try to dump a record object into the placeholder, not the contained value. Same below. But since we should switch to an int anyhow (see above), this then will work.

foosel (commented):
Thank you for this! Got a couple points, can you please look into those? :)

foosel (changes requested):
Accidentally hit "comment" instead of "request changes" on the above review 🤦‍♀️

foosel (approved):

foosel:
Thank you! 🚀

PR #4532, opened 2022-05-28
Fix misc. typos

luzpaz:
Found via `codespell -q 3 -S *.po`


foosel:
Thank you! But can you please refile that against the `maintenance` branch? The `devel` branch is reserved for big new (possibly breaking) functionality, which typos thankfully don't fall under :)

luzpaz:
Closed in favor of #4533

Issue #4520, opened 2022-05-23
[Request] Show feedback if Github Access token is configured in UI


bruvv:
### Is your feature request related to a problem? Please describe.

https://github.com/OctoPrint/OctoPrint/issues/4519

### Describe the solution you'd like

Now when filling in the GitHub toke in the GUI there is no UI feedback. As a front end developer I would like to see some feedback that it did anything. Now it does not change whenever there is a key in `config.yaml` 

I would suggest to change the label of the input field:
![image](https://user-images.githubusercontent.com/3063928/169842305-d621b663-e015-49fd-b48d-fe73180ea5b5.png)
When you set it to "disabled" also the cursor will change to a red marker showing you you cannot edit it anymore.

```
<input type="text" class="input-block-level" data-bind="value: config_githubToken" placeholder="Token entered but hidden for privacy reasons" disabled="">
```

cp2004:
Seems like a better solution, agreed. I don't think disabling the field should be done however, since that stops the token being replaced if it expires or a new one is generated. Would you like to create a PR?

bruvv:
That is very true! I'll see what I can do.

FedericoNembrini:
I am working on it.

Here are my considerations:
As far as I can see, the credentials are not retrieved from the configuration file because of this line of code
`return {"never": [["credentials"]]}`
Without retrieving the credentials, it is impossible to tell whether a token has been entered or not. The solution I'm thinking of is to remove the rule so that I can check for credentials and set a boolean flag, and then remove the credentials so that they are not exposed via the API, maintaining the current logic.

Also, is it not possible to remove an inserted token? I realised this because I can't find the config.yaml file, so I can't remove it.

cp2004:
It may be possible to use something like [`on_settings_load`](https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.on_settings_load) to replace the `credentials` section with boolean values rather than the actual credentials.

> Also, is it not possible to remove an inserted token? I realised this because I can't find the config.yaml file, so I can't remove it.

The only way would be through config.yaml at the moment - https://docs.octoprint.org/en/master/configuration/config_yaml.html explains where it is stored on most platforms. This may be in a hidden folder. That could be another PR with a 'delete' button 🙂

FedericoNembrini:
Yes, I am currently doing tests using this:
```
[...]

PR #4540, opened 2022-06-11
Add a display indication when the Github access token is set

FedericoNembrini:
Implementation of #4520 
This PR adds a computed flag to indicate to the view whether or not to show the input of the Github authentication token with a specific placeholder (if not present/if already present but hidden)

Start OctoPrint, display token placeholder when not present. Add the token to view the updated placeholder.

#4520 

image
image

#### Further notes
I am not entirely sure if this implementation is the most correct one. Please let me know if you have any suggestions.

Also, as mentioned in #4520 , it is currently not possible to remove the token entered via the interface. If I can in the next few days I will create a request and start working on it.

### DIFF
--- a/src/octoprint/plugins/softwareupdate/__init__.py
+++ b/src/octoprint/plugins/softwareupdate/__init__.py
@@ -704,7 +704,6 @@ class SoftwareUpdatePlugin(
             "updatelog_cutoff": 30 * 24 * 60,  # 30 days
             "credentials": {
                 "github": None,
-                "github_set": False,
                 "bitbucket_user": None,
                 "bitbucket_password": None,
             },
@@ -714,14 +713,11 @@ class SoftwareUpdatePlugin(
         # ensure we don't persist check configs we receive on the API
         data = dict(octoprint.plugin.SettingsPlugin.on_settings_load(self))
 
-        # set credentials data flags and remove credentials fields
-        if "credentials" in data:
-            if "github" in data["credentials"]:
-                data["credentials"]["github_set"] = bool(data["credentials"]["github"])
-            for key in data["credentials"]:
-                if key == "github_set":
-                    continue
-                data["credentials"][key] = None
+        # set credentials flag
+        credentials = self._settings.get(["credentials"])
+        data["credentials"] = {}
+        for key, value in credentials.items():
+            data["credentials"][f"{key}_set"] = bool(value)
 
         if "checks" in data:
             del data["checks"]
@@ -779,7 +775,7 @@ class SoftwareUpdatePlugin(
         return data
 
     def get_settings_restricted_paths(self):
-        return {}
+        return {"never": [["credentials"]]}
 
     def on_settings_save(self, data):
         # ~~ plugin settings
--- a/src/octoprint/plugins/softwareupdate/__init__.py
+++ b/src/octoprint/plugins/softwareupdate/__init__.py
@@ -704,6 +704,7 @@ class SoftwareUpdatePlugin(
             "updatelog_cutoff": 30 * 24 * 60,  # 30 days
             "credentials": {
                 "github": None,
+                "github_set": False,
                 "bitbucket_user": None,
                 "bitbucket_password": None,
             },
@@ -712,6 +713,19 @@ class SoftwareUpdatePlugin(
     def on_settings_load(self):
         # ensure we don't persist check configs we receive on the API
         data = dict(octoprint.plugin.SettingsPlugin.on_settings_load(self))
+
+        # set credentials data flags and remove credentials fields
+        if "credentials" in data:
+            if "github" in data["credentials"]:
+                data["credentials"]["github_set"] = (
+                    data["credentials"]["github"] is not None
+                    and data["credentials"]["github"] != ""
+                )
+            for key in data["credentials"]:
+                if key == "github_set":
+                    continue
+                data["credentials"][key] = None
+
         if "checks" in data:
             del data["checks"]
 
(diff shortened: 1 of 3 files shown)


cp2004 (dismissed):

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -704,6 +704,7 @@ def get_settings_defaults(self):
             "updatelog_cutoff": 30 * 24 * 60,  # 30 days
             "credentials": {
                 "github": None,
+                "github_set": False,
```
foosel, line 707: This should not be part of the settings defaults as it is inherently tied to the value of `credentials.github`, but rather be set on the API response via `on_settings_load`. That would then also mean the change to `get_settings_restricted_paths` can be rolled back again, and would simplify the logic in `on_settings_load`.

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -712,6 +713,16 @@ def get_settings_defaults(self):
...
+
+        # set credentials data flags and remove credentials fields
+        if "credentials" in data:
+            if "github" in data["credentials"]:
+                data["credentials"]["github_set"] = bool(data["credentials"]["github"])
+            for key in data["credentials"]:
+                if key == "github_set":
+                    continue
+                data["credentials"][key] = None
+
```
foosel, line 725: Instead of all of this which basically duplicates logic that is already handled by `get_settings_restricted_paths`, all just because the flag is put into the `credentials` field on the default settings (see above), how about this:

```suggestion

        # set credentials flag
        data["credentials"] = {"github_set": bool(self._settings.get(["credentials", "github"])) }
```

Or, even better (because it will automatically adapt to more credentials added in the future):

```suggestion

        # set credentials flag
        credentials = self._settings.get(["credentials"])
        data["credentials"] = {}
        for key, value in credentials.items():
            data["credentials"][f"{key}_set"] = value is not None
```

src/octoprint/plugins/softwareupdate/__init__.py
```diff
@@ -768,7 +779,7 @@ def on_settings_load(self):
         return data
 
     def get_settings_restricted_paths(self):
-        return {"never": [["credentials"]]}
+        return {}
```
foosel, line 782: Revert this

foosel (changes requested):
I'd like to propose a minor logic change that would also get rid of the boolean flag stored in the config that is fully dependent on something already stored in the config.

src/octoprint/plugins/softwareupdate/__init__.py
FedericoNembrini: Thanks for the explanation.
I thought that from the plugin it was only possible to access the settings allowed by `get_settings_restricted_paths`

FedericoNembrini:
I was testing the proposed changes but noticed a strange behaviour.
If after entering a github token, I save the general settings (even without changing any settings) the softwareupdate credentials are overwritten with:
```
softwareupdate:
    _config_version: 9
    credentials:
      bitbucket_password_set: false
      bitbucket_user_set: false
```

From what I have been able to see so far it is because when it save the settings a check is made and only those that have been changed are saved. In this case it detects that only these two have been changed and overwrites them on top of the rest.

foosel:
Ah, good point. First of all, `on_settings_save` also needs to filter out these newly added keys.

Something like this (untested):

``` python
def on_settings_save(self, data):
    data = {(key, value) for key, value in data.items() if not key.endswith("_set")}
    if data:
        super().on_settings_save(data)
```

The rest of your description sounds like there might be a bug however in how these incoming settings are being merged back into the settings structure, potentially a side effect of the settings refactoring that happened in 1.8.0... That needs some investigation.

FedericoNembrini:
Filtering the data with:
```
if "credentials" in data:
    data["credentials"] = dict((key, value) for key, value in data["credentials"].items() if not key.endswith("_set"))
```

The problem persists, however, because if the dictionary 'credentials' is present, but empty, the credentials are still emptied.

A possible solution would be to remove credentials if there are no values within it:
```
if "credentials" in data:
    data["credentials"] = dict((key, value) for key, value in data["credentials"].items() if not key.endswith("_set"))
    if len(data["credentials"]) == 0:
        del data["credentials"]
```

However, at this point, if necessary, it would no longer be possible to remove all credentials by calling the save function and passing it the empty credentials, but they would all have to be specified, without any value.

Otherwise, the actual saving function must be changed to take into account the individual keys of the dictionary credentials:
https://github.com/OctoPrint/OctoPrint/blob/22d42d48641d9c1a9ca459487b35b3dfb5037d42/src/octoprint/plugins/softwareupdate/__init__.py#L772-L788

foosel:
With [18b0565](https://github.com/OctoPrint/OctoPrint/commit/18b0565ea636902d692677a4969f6e47bb997fca) that should be a thing of the past.

cp2004:
That shouldn't be necessary, as it is iterating through the default values, which wouldn't include the `_set` tokens as I understand it

foosel:
> That shouldn't be necessary, as it is iterating through the default values, which wouldn't include the `_set` tokens as I understand it

Exactly. The change in `maintenance` got rid of this necessity.

FedericoNembrini:
Ah, you are right. I misunderstood the last comment. :/

foosel:
Please always request a new review when you are done making your modifications.

foosel (approved):
Looking good now! 👍

PR #4559, opened 2022-06-27
GCodeViewer: Async download

JoveToo:
    Brainstorming ticket
    if necessary!
    lessc)
    have added unit tests
    nothing broke

This PR moves the download of the GCode file in the GCodeViewer to the Worker and makes it asynchronous. The GCode file is never entirely in memory.

This now allows downloading a 500MB GCode file in Chrome (with compression enabled).

Downloaded and parsed several files.

#### Further notes

This is not complete, it does not implement the skipUntil feature yet.


JoveToo:
The pre-commit hooks fail on eslint, because it does not handle async functions correctly. 

I do not know how to fix that.

I have made the PR anyway so this can be evaluated in the mean time.

foosel:
> The pre-commit hooks fail on eslint, because it does not handle async functions correctly.
> 
> I do not know how to fix that.

This is intentional as eslint is set up to enforce ES5 compatibility (many people out there are still using old tablets with ancient browsers that don't support ES6). So no `async` keyword support and no arrow functions either. See also a9ea150b927db7d0a275ece9103218eab3e9bfa2. With [nodejs-bin](https://pypi.org/project/nodejs-bin/) it might be possible to give an initial `babel` run a chance here, but that is something that should be looked at outside of the context of this PR.

JoveToo:
@foosel I don't think I can implement this without those keywords.
Do you have statistics of your octoprint clients?  ES6 is already 7 years old...

foosel:
Good point actually, the last time this came up we did not yet have data on this, so around the time I added the linter to prevent accidents after a complaint I also added the client environment to the tracking data:

https://data.octoprint.org/export/client_environment_stats_30d.json

JoveToo:
Alright, that are some useful statistics.
|Type|number|percentage|
|:----|----:|----:|
|Total Instances:  |   175974 ||
|Support:             | 174438 | 99.13% |
|No Support:        |     399  | 0.23% |
|Unknown support:  |     1137 | 0.65% |

This is the script that calculated the statistics. If you can identify which version of WebKit support ES6 you can get 424 extra instances classified.
[convertstatistics.js.txt](https://github.com/OctoPrint/OctoPrint/files/9004658/convertstatistics.js.txt)

foosel:
With that data, a case could be made for dropping ES5 support. However, that is not a decision I cannot make overnight (and while exhausted AF). I've got a vacation coming up, I'll look into this again on my return at the end of July. I've extracted it as a request into #4562. In there I also mention this:

> If we go for this, what we definitely should make available/link to is some page for OctoPrint's users to figure out if their browser supports ES6 (and possibly other stuff in the future). caniuse.com has the data, and https://caniuse.com/es6 holds the info on ES6 specifically, but I couldn't yet locate a "you are here" kinda feature. Any hints in that regard would be highly appreciated.

Any input on that point would be greatly appreciated.

foosel:
I've just pushed a commit to `maintenance` that stops the enforcement of ES5. ES6 language features such as arrow functions and async/await should thus no longer be an issue to get through the pre-commit, and are hereby also allowed.

JoveToo:
This does not include the skipUntil feature because that is difficult.  It needs a decision on if and how this needs to be implemented. Statistics on who actually uses this feature would be useful.

Since we are parsing as the data comes in, we need to skip the data until the "SkipUntil" string. If this string is not found, it will not parse any data. This would be a change in functionality compared to the previous functionality when the setting was ignored if the string was not found.

The only way to implement the old functionality would be to cache all data while searching for the string, throwing it away if it is found or parse it all when it isn't. In the case the string isn't found, we are back to caching the whole gcode file in memory. Considering avoiding this is exactly why this PR exists, it seems contra-productive to me.

JoveToo:
Hmm, I thought of a possible solution. 

If the generator function would work as normal but the string check would be part of parsing a gcode line, we could purge any data that has been parsed so far, basically resetting the parsing. This would include a command sent to the reader to purge any layers it has received this far. 

The generator would also check for the string and return it as normal but reset the `currentDownloadLength` and adjust the `totalDownloadLength` variables. 

It will need some serious documention, however.

foosel:
> Statistics on who actually uses this feature would be useful.

Sadly I have absolutely zero data on that, sorry. 

Am I right in understanding that the problem is that we do not know whether `skipUntilThis` is even contained in the file or not? If so, how about some server side pre-check in that regard? One request that utilizes something like [this](https://github.com/OctoPrint/OctoPrint-FileCheck/blob/c24114008b7445fdbf668ae6a36dd2a109c6ebc5/octoprint_file_check/__init__.py#L78-L118) (which I'd be happy to move into OctoPrint's core utils) to figure this out, then the viewer can make an informed decision on how to proceed, without any overhead. If for some reason the request or check cannot be made, assume no skipuntil value was set and just visualize everything.

JoveToo:
Yes, that is the issue.  If we know for sure it is in the file, we can throw away the data until we see the string. A server side check would simplify the gcode viewer a lot.

JoveToo:
Btw, based on #4594, now we have to implement this feature in the worker, we could modify the current code to only ignore head movements and extrusion commands but parse all the other stuff normally. This would mean a change in how the feature works but it would prevent such issues.

Btw, how is the percentage handled on server side? If the gcode viewer cuts off a part of the file, the completion percentage calculation is diffferent from the one on the server and things should get out of sync?

foosel:
> we could modify the current code to only ignore head movements and extrusion commands but parse all the other stuff normally

Probably a good idea, I say go for it.

> Btw, how is the percentage handled on server side? If the gcode viewer cuts off a part of the file, the completion percentage calculation is diffferent from the one on the server and things should get out of sync?

The viewer calculates the percentages based on position in file vs file size, *before* the skipUntil string ever gets even remotely looked at. So they do sync up, the viewer side just skips the first few percent.

JoveToo:
> Probably a good idea, I say go for it.

Will do.

> The viewer calculates the percentages based on position in file vs file size, _before_ the skipUntil string ever gets even remotely looked at. So they do sync up, the viewer side just skips the first few percent.

I don't think this is accurate. The first thing the gcodeviewer does is shorten the file, before it is parsed and the percentages are calculated. See https://github.com/OctoPrint/OctoPrint/blob/maintenance/src/octoprint/plugins/gcodeviewer/static/js/gcodeviewer.js#L515 which calls `self.showGCodeViewer` which does the skipUntil split and passes the shorted file to the gCodeReader which calls `prepareGCode` which does the percentage calculation. So unless Octoprint itself also calculates the percentages on a shorter file, they get out of sync.

src/octoprint/plugins/gcodeviewer/__init__.py
```diff
@@ -70,6 +73,19 @@ def on_settings_migrate(self, target, current):
                     self._settings.set_int(["sizeThreshold"], config["sizeThreshold"])
                 self._settings.global_remove(["gcodeViewer"])
 
+    def on_api_get(self, request):
```
foosel, line 76: The way this is implemented heavily deviates from other file related endpoints. Initially I thought it would be easiest to just have the get point only operate on the currently selected file, making the `SimpleApiPlugin` the easiest approach to implementation since no parameter transfer would be needed, but if you want to go with a parameter based approach (which also has its valid reasons, to combat issues caused by race conditions should the user go nuts with selecting new files), this should rather be a `BlueprintPlugin` with a `GET` method and the file path encoded into its path.

That has the upside that we can also just return 404 if we can't find the file, instead of claiming "nope, clean".

Something like this (assuming the mixin `octoprint.plugin.BlueprintPlugin` mixin is also included above):

``` python
@octoprint.plugin.BlueprintPlugin.route("/skipuntilcheck/<string:origin>/<path:filename>", methods=["GET"])
def check_skip_until_presence(self, origin, path):
    try:
        path = self._file_manager.path_on_disk(origin, path)
    except NotImplementedError:
        # storage doesn't support path on disk
        flask.abort(404)

    if not os.path.exists(path):
        # path doesn't exist
        flask.abort(404)

    skipUntilThis = self._settings.get(["skipUntilThis"])
    if not skipUntilThis:
        # no skipUntilThis, no need to search, shortcut
        return flask.jsonify(present=False)

    return flask.jsonify(present=search_through_file(path, skipUntilThis))

def is_blueprint_csrf_protected(self):
    return True
```

src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
```diff
@@ -376,10 +447,23 @@ var doParse = function () {
...
 
+    // skipUntil preparations
+    var skipUntilFound = false;
+
     var i, j, args;
 
+    // if skipUntil is set, get skipUntilPresent
+    skipUntilPresent = false;
+    if (skipUntil !== undefined && skipUntil !== "") {
+        result = await fetch("/api/plugin/gcodeviewer?path=" + localPath);
```
foosel, line 458: With the above change, this would then be

```suggestion
        result = await fetch("/plugin/gcodeviewer/skipuntilcheck/local/" + localPath);
```

src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
```diff
@@ -376,10 +447,23 @@ var doParse = function () {
...
+        if (result.ok) {
+            response = await result.json();
+            skipUntilPresent = response.result == "true";
```
foosel, line 461:
```suggestion
            skipUntilPresent = response.present;
```

src/octoprint/plugins/gcodeviewer/static/js/viewer/reader.js
```diff
@@ -210,35 +210,18 @@ GCODE.gCodeReader = (function () {
...
                 (gCodeOptions["compressionSizeThreshold"] > 0 &&
                     gCodeOptions["compressionSizeThreshold"] <= totalSize);
 
             GCODE.ui.worker.postMessage({
-                cmd: "parseGCode",
+                cmd: "downloadAndParseGCode",
                 msg: {
-                    gcode: gcode,
+                    path: reader.path,
+                    localPath: reader.localPath,
```
foosel, line 223: `path` vs `localPath` confused me a bit during review further down. How about `url` and `path` instead?

foosel (changes requested):
Finally gotten around to another review, sorry for the delay. 1.8.x has been nuts the past weeks. 

Anyhow, I have some points, mostly related to a refactoring I'd like to see for better consistency, see below.

src/octoprint/plugins/gcodeviewer/__init__.py
JoveToo: I see you return True on the csrf protected method. Do I need to do something in the worker for that?

foosel: No, not since you are only using a `GET` endpoint. If at some point you'd want to `POST`, `PUT`, `PATCH`, `DELETE` stuff, the `fetch` calls would need some love to add some headers. The `return True` here is merely so that an explicit decision has been made for this plugin, because otherwise OctoPrint will implicitly make it (`False`) and log two warnings, one for the implicit decision, one for the disabled CSRF protection. See the heads-ups in the release notes of 1.8.3.

foosel (approved):

rfinnie:
Sorry if I'm missing something, but I've been unable to load any gcode in the viewer with this commit applied.  The in-UI status is stuck on "Parsing... (100%)" and the browser console displays:

> 09:20:53.786 Error calling fromCurrentData on view model GcodeViewModel : loadFile@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:798:523
GcodeViewModel/self.loadFile@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:727:866
GcodeViewModel/self._processData@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:731:740
GcodeViewModel/self.fromCurrentData@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:728:448
callViewModelIf@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1726:75
callViewModelsIf/<@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1720:153
Pn@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:30:530
m/ur/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:51:66
callViewModelsIf@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1720:95
callViewModels@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1719:72
DataUpdater/self._onCurrentData/<@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1600:266
DataUpdater/self._ifInitialized@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1614:1002
DataUpdater/self._onCurrentData@http://127.0.0.1:5000/static/webassets/packed_core.js?cc63d943:1600:226
OctoPrintSocketClient.prototype.propagateMessage/<@http://127.0.0.1:5000/static/webassets/packed_client.js?39d41c31:24:116
Pn@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:30:530
m/ur/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:51:66
OctoPrintSocketClient.prototype.propagateMessage@http://127.0.0.1:5000/static/webassets/packed_client.js?39d41c31:24:77
onMessage/<@http://127.0.0.1:5000/static/webassets/packed_client.js?39d41c31:31:316
m/Zt/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:47:23
_t@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:35:392
m/Vt/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:46:369
m/ur/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:51:73
onMessage@http://127.0.0.1:5000/static/webassets/packed_client.js?39d41c31:31:278
[5]</r.prototype.dispatchEvent@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:1585:2864
[14]</</w.prototype._transportMessage/<@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:1585:13481
[14]</</w.prototype._transportMessage@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:1585:13432
[3]</i.prototype.emit@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:1585:1745
[38]</c/this.ws.onmessage@http://127.0.0.1:5000/static/webassets/packed_libs.js?6bee1dee:1585:39734
packed_core.js:1727:37

Though I haven't yet been able to figure out what the error it's reporting actually is.  Tried in Firefox and Chrome, in multiple environments (systeminfo bundle is from a minimal dev environment in upstream/maintenance, but also tried switching my "normal" home environment to maintenance).  Multiple gcode files tried, and I've bisected it directly to this commit.

[octoprint-systeminfo-20221027092154.zip](https://github.com/OctoPrint/OctoPrint/files/9881361/octoprint-systeminfo-20221027092154.zip)

cp2004:
I can confirm I wasn't able to load a gcode file - chromium based browsers give much more helpful error messages (at least for OctoPrint):

```
helpers.js:1471 
        
       Error calling onTabChange on view model GcodeViewModel : ReferenceError: totalSize is not defined
    at Object.loadFile (http://localhost:5000/plugin/gcodeviewer/static/js/viewer/reader.js:217:65)
    at GcodeViewModel.self.loadFile (http://localhost:5000/plugin/gcodeviewer/static/js/gcodeviewer.js:524:35)
    at GcodeViewModel.self.onTabChange (http://localhost:5000/plugin/gcodeviewer/static/js/gcodeviewer.js:875:22)
    at callViewModelIf (http://localhost:5000/static/js/app/helpers.js:1455:35)
    at http://localhost:5000/static/js/app/helpers.js:1382:13
    at Pn (http://localhost:5000/static/js/lib/lodash.min.js:9:530)
    at Function.<anonymous> (http://localhost:5000/static/js/lib/lodash.min.js:30:66)
    at callViewModelsIf (http://localhost:5000/static/js/app/helpers.js:1380:7)
    at callViewModels (http://localhost:5000/static/js/app/helpers.js:1374:5)
    at OctoPrint.coreui.exports.onTabChange (http://localhost:5000/static/js/app/main.js:153:13)
```

JoveToo:
This problem is fixed in #4673

PR #4560, opened 2022-06-28
🐛 Make settings threadsafe

### LINKED ISSUE #4551: Settings Object Is Not Thread Safe - And Thus Sets/Saves Can Lose Data
### The problem

I have had a few users reporting issues where OctoPrint will reboot into safe mode sometimes when OctoEverywhere is installed. I have been debugging it for a while, and I think I found the cause.

When OctoPrint boots into safe mode, it's because of a "failed startup". The flag that controls this is the `incompleteStartup` flag. The flag is set when the server first boots and is cleared just after `on_after_startup` is called on the plugins.

What was happening in my case was two issues. I have logic that starts a **thread** to make the main connection to my servers after `on_after_startup` is called. That thread does some work, some of which writes settings into the plugin settings object. The first problem was that I was using the `force=True` argument for all of my set and save calls, which I shouldn't have been.

But the second problem I debugged was that it seems the settings object set/save functions don't thread safe. So if it just so happened that OE was setting and saving the config the same time the system logic was saving the removal of the `incompleteStartup`, sometimes the `incompleteStartup` removal would get overwritten.

With about 15 plugins loaded on a rasp pi W2 I was about to get the race above to repo about 1 out of 10 times OctoPrint started. 

I looked into the Settings object a little, and while I don't understand it entirely, I do think the `self._dirty` flag could get into bad states since it's not protected by a lock at all. But that would be alright, the next setting write/save would correct it. I think the bigger problem is the save/set functions don't seem to be thread-safe between each other, which I think results in the data loss.

I think it would be easy to repo this more often if you wrote a test plugin that spins in a loop after `on_after_startup` that users set/save with force=True.

While I'm going to remove the force=ture flags from my logic and try to make sure every setting possible is set and written before on_after_startup returns, I think there's still some chance this could happen to OE or any plugin in the future since the settings object doesn't seem to be entirely thread-safe.

No, it did not persist (it's an issue with one of your installed plugins, don't report here)

3rd party plugins aren't enabled

### Version of OctoPrint

1.8.2

OctoPi

Raspi W2 (and lots of other devices from my community)

_No response_

- [X] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
- [X] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)
- [X] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

_No response_

foosel:
  * [ ] ~~If your changes are large or otherwise disruptive: You have
    Brainstorming ticket~~
    if necessary!
  * [ ] ~~If your changes include style sheets: You have modified the
    lessc)~~
    have added unit tests
    nothing broke

setter, save and dirty flag access are now behind
a reentrant lock.

Fixes #4551

Run the server, check if saving still works. Aggressive race condition testing is still needed.

#4551 

#### Further notes

Would love to get some more sets of eyes on these changes, not sure if I missed something.

### DIFF
--- a/src/octoprint/settings/__init__.py
+++ b/src/octoprint/settings/__init__.py
@@ -25,6 +25,7 @@ import logging
 import os
 import re
 import sys
+import threading
 import time
 from collections import ChainMap
 from collections.abc import KeysView
@@ -486,6 +487,8 @@ class Settings:
         self._last_effective_hash = None
         self._mtime = None
 
+        self._lock = threading.RLock()
+
         self._get_preprocessors = {"controls": self._process_custom_controls}
         self._set_preprocessors = {}
 
@@ -716,9 +719,10 @@ class Settings:
         self._last_effective_hash = None
 
     def _mark_dirty(self):
-        self._dirty = True
-        self._dirty_time = time.time()
-        self._forget_hashes()
+        with self._lock:
+            self._dirty = True
+            self._dirty_time = time.time()
+            self._forget_hashes()
 
     @property
     def effective(self):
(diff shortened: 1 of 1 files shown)


src/octoprint/settings/__init__.py
```diff
@@ -1769,8 +1774,9 @@ def remove(self, path, config=None, error_on_path=False, defaults=None):
         chain = self._map.with_config_defaults(config=config, defaults=defaults)
 
         try:
-            chain.del_by_path(path)
```
QuinnDamerell, line 1772: I'm not familiar with this logic or how it works at all, but would `self._map.with_config_defaults` also need to be in the lock? Or would that be safe to execute from multiple threads?

QuinnDamerell (commented):
Looks good to me, pretty much exactly how I would have done it. I just have one comment about the safety of a function I'm not familiar with.

src/octoprint/settings/__init__.py
foosel: That's a very good question that I have currently no answer on and that needs some digging. Probably a thing for after my vacation.

foosel: Have given that some more thought and looking at now, and I don't think we need that. Building the new chain to work on is pure recombining of a view on existing data structures, without any modification taking place as far as I can see. So that should be inherently thread safe already and thus doesn't need to be pulled into the lock. The goal here is to make sure we don't write an inconsistent state to disk due to concurrent modification from two threads, and that should be ensured by the existing changes.

QuinnDamerell:
Great work! This will be a great fix to get into the system!

Issue #4525, opened 2022-05-25
Fix double message when trying to skip a mandatory wizard during


foosel:
### Problem

  - run setup wizard (or trigger some other mandatory wizard steps to show up)
  - try to skip a mandatory steps by clicking "Next" without filling it in
  - Result: You get a warning that the step is mandatory *twice*

### Solution 

Find the reason for that and fix it.

max246:
I have investigated a little bit on this issue, apparently on plugin bootstrapWizard is binding the click event twice in someone to the button NEXT and PREVIOUS.

I am not familar with this plugin but seems like the method "onTabShow" and "onTabChange" is getting called twice as soon as you click the button.

foosel:
1.9.0 has been released.

PR #4561, opened 2022-06-28
Fix/4525

max246:
  * [] ~If your changes are large or otherwise disruptive: You have
    Brainstorming ticket~
    if necessary!
    lessc)
    have added unit tests
    nothing broke

Fix the double trigger of the innerTabShown by removing one listener action.
This is a fix for this ticket #4525

I have tested on the local enviroment and it can be tested by running the wizard

Ticket #4525

#### Further notes

Might need to keep an eye regarding the previous note left about bootstrap and the reason why there were two actions

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -166,6 +166,7 @@ date of first contribution):
   * [Nicu Surdu](https://github.com/surdu)
   * [Zack Lewis](https://github.com/lima3w)
   * [Billy Richardson](https://github.com/richardsondev)
+  * [Christian Bianchini](https://github.com/max246)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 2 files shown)


max246:
> I have no idea yet _why_ this fixes the issue, as I cannot see the removed event firing in the broken state, but it _does_ indeed fix the issue sweat_smile Merging. But also going to give this mystery maybe half an hour to figure out the cause, because magic like this always worries me no_mouth
I guess is something about the show event, jquery might not be clever enough to add only one listener and it fires twice. I dont know enough jquery, it has been a dark world all the time :) but at least is not something it the octoprint code that gets loaded twice!

foosel:
Most likely caused by jquery's namespace feature...

> An event name can be qualified by event namespaces that simplify removing or triggering the event. For example, "click.myPlugin.simple" defines both the myPlugin and simple namespaces for this particular click event. A click event handler attached via that string could be removed with .off("click.myPlugin") or .off("click.simple") without disturbing other click handlers attached to the elements. Namespaces are similar to CSS classes in that they are not hierarchical; only one name needs to match. Namespaces should contain upper/lowercase letters and digits only.

So `show` would trigger both `show` and `show.bs.tab`, explaining the double calls. Nice find!

Issue #4493, opened 2022-04-22
The first M20 done on connection is missing the L flag.


nophead:
### The problem

With the latest version of Marlin with `Cap:EXTENDED_M20:1` the first display of the file list after connection does not show long filenames but refreshing the list does. The first M20 command that is sent immediately after the `M115` does not have the `L` parameter but subsequent ones do. 

It looks like the M20 command is built and added to the send queue before the results from the M115 have been parsed and the EXTENDED_M20 capability set.

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

OctoPrint 1.7.3 Python 3.7.3

### Operating system running OctoPrint

OctoPi 0.18.0

### Printer model & used firmware incl. version

Mendel90

### Browser and version of browser, operating system running browser

Chrome on Win7 64.

### Checklist of files to include below

- [X] Systeminfo Bundle (always include!)
- [ ] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
- [ ] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)
- [ ] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

### Additional information & file uploads

[octoprint-systeminfo-20220422153035.zip](https://github.com/OctoPrint/OctoPrint/files/8541814/octoprint-systeminfo-20220422153035.zip)

```
Recv: Cap:SERIAL_XON_XOFF:0
Recv: Cap:BINARY_FILE_TRANSFER:0
Recv: Cap:EEPROM:1
Recv: Cap:VOLUMETRIC:1
Recv: Cap:AUTOREPORT_POS:0
Recv: Cap:AUTOREPORT_TEMP:1
Recv: Cap:PROGRESS:0
Recv: Cap:PRINT_JOB:1
Recv: Cap:AUTOLEVEL:0
Recv: Cap:RUNOUT:0
Recv: Cap:Z_PROBE:0
Recv: Cap:LEVELING_DATA:0
Recv: Cap:BUILD_PERCENT:0
Recv: Cap:SOFTWARE_POWER:0
Recv: Cap:TOGGLE_LIGHTS:0
Recv: Cap:CASE_LIGHT_BRIGHTNESS:0
Recv: Cap:EMERGENCY_PARSER:1
Recv: Cap:HOST_ACTION_COMMANDS:0

---------------------------------------------------------------------------
Your printer's firmware supports host action commands, but they are
[...]

PR #4577, opened 2022-07-14
Fix/4493 - Defer M20 commands until initial capability report is done

kForth:
## What does this PR do and why is it necessary?

This is a proposed fix for issue #4493.

This PR tracks the start and end of the first capability report and defers any `M20` commands until it's done.

The first `M20` command is queued before the response from `M115` is processed, meaning long filenames are not shown in the file list until it's refreshed.

## How was it tested? How can it be tested by the reviewer?

- Connect to any printer and observe the terminal output
- Verify no `M20` command is sent before the capability report is done
- Verify an `M20` command is sent shortly after the capability report finishes
  - with the `L` parameter if `EXTENDED_M20` is supported

## What are the relevant tickets if any?

This is a fix for issue #4493.

## Screenshots (if appropriate)

Not really screenshots but close enough.

<details>
  <summary>Terminal Output Before</summary>

  ```
    Connected to: VIRTUAL(read_timeout=10.0,write_timeout=10.0,options={'_config_version': 1, 'enabled': True}), starting monitor
    Send: N0 M110 N0*125
    Recv: start
    Send: N0 M110 N0*125
    Recv: Marlin: Virtual Marlin!
    Recv: \x80
    Recv: SD card ok
    Recv: ok
    Send: N0 M110 N0*125
    Changing monitoring state from "Connecting" to "Operational"
    Recv: ok
    Send: N0 M110 N0*125
    Recv: ok
    Send: N1 M115*39
    Recv: ok
    Send: N2 M20*19
    Recv: FIRMWARE_NAME:Virtual Marlin 1.0 PROTOCOL_VERSION:1.0
    Recv: Cap:AUTOREPORT_TEMP:1
    Recv: Cap:AUTOREPORT_SD_STATUS:1
    Recv: Cap:AUTOREPORT_POS:0
    Recv: Cap:EMERGENCY_PARSER:1
    Recv: Cap:EXTENDED_M20:1
    Recv: ok
    Send: M155 S2
    Recv: Begin file list
    Recv: End file list
    Recv: ok
    [...]
    Recv: ok
    [...]
    Recv: ok
```
</details>

<details>
  <summary>Terminal Output After</summary>

  ```
    Connected to: VIRTUAL(read_timeout=10.0,write_timeout=10.0,options={'_config_version': 1, 'enabled': True}), starting monitor
    Send: N0 M110 N0*125
    Recv: start
    Send: N0 M110 N0*125
    Recv: Marlin: Virtual Marlin!
    Recv: \x80
    Recv: SD card ok
    Recv: ok
    Send: N0 M110 N0*125
    Changing monitoring state from "Connecting" to "Operational"
    Recv: ok
    Recv: ok
    Send: N0 M110 N0*125
    Recv: ok
    Send: N1 M115*39
    Recv: FIRMWARE_NAME:Virtual Marlin 1.0 PROTOCOL_VERSION:1.0
    Recv: Cap:AUTOREPORT_TEMP:1
    Recv: Cap:AUTOREPORT_SD_STATUS:1
    Recv: Cap:AUTOREPORT_POS:0
    Recv: Cap:EMERGENCY_PARSER:1
    Recv: Cap:EXTENDED_M20:1
    Recv: ok
    Send: M155 S2
    Recv: ok
    [...]
    Recv: ok
    Send: M20 L
    Recv: Begin file list
    Recv: End file list
    Recv: ok
  ```
</details>

## Further notes

I'm unsure how likely/possible it is for the 'cap:' lines to be interrupted, but it could cause the M20 to be sent before the `EXTENDED_M20` capability is reported.

An alternative solution is to refresh the list as soon as the `EXTENDED_M20` capability is reported (like below), but that will double report the file list.

```python
elif (
          capability == self.CAPABILITY_EXTENDED_M20 and enabled
      ):
          self._logger.info(
              "Firmware states that it supports long filenames"
          )
         self.refreshSdFiles()
```

Thoughts are appreciated.


kForth:
I've realized another potential bug in this approach: the server will block all calls to `refreshSdFiles` if the printer firmware never reports any 'cap:' lines.

This may not be a big issue since the reprap wiki shows [good support](https://reprap.org/wiki/G-code#M115:_Get_Firmware_Version_and_Capabilities) the `M115` command, and all [firmwares that support](https://reprap.org/wiki/G-code#M20:_List_SD_card) `M20` also support `M115`.


An alternative approach is to force the first command to be `M20 L`, probably activated by a setting in the printer profile.

printer profile setting

A few caveats with this approach, if the option is set to True:
 - First refresh will be `M20 L` even if printer responds with `Cap:EXTENDED_M20:0`
 - Every refresh will be `M20 L` if the printer never reports a `Cap:EXTENDED_M20` line

There is no change to the existing behaviour if this option is set to False.
 
You can see my implementation of this on my [fix/4493-assume-extended-m20-option](https://github.com/kForth/OctoPrint/tree/fix/4493-assume-extended-m20-option) branch .
 
Thoughts are appreciated.

cp2004:
I was about to test if no capabilities would result in an issue. If there's one thing that's certain, it's that you are always uncertain on what might happen with printer communication - random things break *all the time*. Creality 2x temp reporting for example, that's quite an unexpected break.

I like the idea of sending an M20 L command as soon as the capability report comes in - it's fairly robust. The issue then becomes when to send a regular M20, if the capability is not received.

With the assumption method you have there, I think the option would be more at home in the serial connection settings than the printer profile. Yes, I agree that it could change on a printer-by-printer basis, but none of the other serial settings are part of a profile - unless they are all migrated then I think it should stick with the rest of them. The only thing I am worried about is some firmware variants complaining about an 'Unknown parameter' - I have tested my printer, it doesn't. One I would like to see is Prusa, since that has had some of the strangest protocol behaviour before (aside from Creality).

To summarise the options - all have one downside that I can see.

* Original PR: Wait until capabilities are finished reporting to refresh files
   * May never be sent if no cap report received
* Listing files when report received
  * Duplicates the listing command
* Assume LFN support, send M20 L unless configured/detected as **not supported**
  * Will some printers not respond to this?

I don't like this kind of dilemma...

jneilliii:
> * Assume LFN support, send M20 L unless configured/detected as **not supported**
>   * Will some printers not respond to this?

I was thinking this would be the way to go, but that's under the assumption that the unknown parameter would just be ignored. Gina may have more insight when she gets back from vacation in regard to that. Pretty sure Marlin is pretty forgiving with stuff like that, but not sure on other firmwares.

jneilliii:
Have reached out on Discord to some PrusaFirmware users to see what happens when `M20 L` is sent to their printer.

jneilliii:
adding responses from PrusaFirmware.

[M20.log](https://github.com/OctoPrint/OctoPrint/files/9128534/M20.log)

kForth:
Thanks for the feedback.

Thinking more about this, I may prefer the 'Assume LFN Support' option because it doesn't modify the original command order or timing.

A fourth option is:
 - Add a setting to override the command used in `refreshSdFiles`.
   - See my implementation on my [fix/4493-sd-refresh-command-setting](https://github.com/kForth/OctoPrint/tree/fix/4493-sd-refresh-command-setting) branch

Then the user could override it to always be `M20 L` or whatever command their firmware may prefer.

The only potential issue I see right now is how to handle adding the `L` parameter for commands other than M20.
The capability is `EXTENDED_M20` so I would not add it to other commands, but I'm not sure the best solution.
 
Screen Shot 2022-07-17 at 10 50 06 PM

I've also updated my original implementation so that there's a setting to enable the feature, default is disabled.
May want to add a warning about potentially breaking the sd card file list if the printer doesn't respond to `M115` correctly.

'Wait to load SD card file list' setting

jneilliii:
I like both these as optional settings, deferring file list loading and what command to use to list sd files. After thinking about it though, the later would technically resolve both allowing the user to manually override the command being sent initially prior to the capability report being read/loaded.

jneilliii:
maybe default the setting to M20 and on capabilities report read programmatically update the setting automatically? Basically make it part of the firmware discovery process, which I think is used for other settings IIRC.

src/octoprint/util/comm.py
```diff
@@ -2003,6 +2008,16 @@ def refreshSdFiles(self, tags=None):
         if not self.isOperational() or self.isBusy():
             return
 
+        if (
+            settings().get(["serial", "waitToLoadSdFileList"])
+            and self._first_cap_report_pending
+        ):
+            self._refresh_sd_files_after_cap_report = True
+            self._logger.info(
```
jneilliii, line 2016: probably would want to be .debug not .info.

src/octoprint/util/comm.py
```diff
@@ -2581,6 +2596,27 @@ def convert_line(line):
...
+
+                    # capability report done
+                    elif self._first_cap_report_started:
+                        self._first_cap_report_pending = False
+                        self._first_cap_report_started = False
+
+                        # refresh sd files now if it was deferred while waiting for report
+                        if self._refresh_sd_files_after_cap_report:
+                            self._refresh_sd_files_after_cap_report = False
+                            self._logger.info("Performing deferred sd file refresh")
```
jneilliii, line 2617: same, probably want as .debug instead of .info

src/octoprint/util/comm.py
```diff
@@ -2730,6 +2766,10 @@ def convert_line(line):
                                 self._logger.info(
                                     "Firmware states that it supports emergency GCODEs to be sent without waiting for an acknowledgement first"
                                 )
+                            elif capability == self.CAPABILITY_EXTENDED_M20 and enabled:
+                                self._logger.info(
```
jneilliii, line 2770: ditto

foosel (changes requested):
The capability report detection logic should really be implemented analog to the firmware information detection logic. So, add another boolean flag `self._firmware_capabilities_received` next to the declaration of `self._firmware_info_received`. A capability report is ongoing if `self._firmware_capabilities` has at least one entry (one `cap:` line was read and parsed into it). So adding some check like this in the monitor loop:

``` python
if self._firmware_capabilities and not self._firmware_capabilities_received and not lower_line.startswith("cap:"):
    self._firmware_capabilities_received = True

    # fire off an event?

    if self._defer_sd_list:
        # SD listing was delayed, do it now
        self.refreshSdFiles()
```

should track nicely whether we have received the full capability report. `self._defer_sd_list` should be set to `settings().getBoolean(["serial", "waitToLoadSdFileList"])` *once* in the constructor, similar to most of the other config parameters.

src/octoprint/util/comm.py
kForth: I've left this as INFO level to match the above capability reports, could just remove the whole `elif` section since it's only for logging purposes.

kForth:
Thanks for the input, I've refactored my code to better match the firmware_info implementation and added an event hook.

I also created a quick test plugin for the hook: https://github.com/kForth/OctoPrint-AfterReportTest

Let me know what you think.

foosel (changes requested):
The logic looks sound now, thank you! Only issue I have now is the name of the new hook 😅

`after_report` is a bit to generic. `capabilities` is already taken for the individual capability lines, but I think `octoprint.comm.protocol.firmware.capability_report` would be fairly self descriptive without any ambiguity as to *what* report we are talking about.

In hindsight it would admittedly have been better to parse the full report first, then call plugins, instead of introducing this individual `capabilities` hook in the first place, but well, we are stuck with that now.

*edit* Also please make sure you have pre-commit installed on your end, your changes currently don't pass its checks due to a typo as seen in the CI results.

foosel:
@kForth please request a new review once you deem this PR ready for one

foosel:
FYI, seen the review request, it's on my TODO list, but I'm currently fully focused on 1.8.3.

foosel (approved):

foosel:
Thank you for the PR, and thank you for your patience while I was busy with 1.8.x :)

PR #4580, opened 2022-07-14
Fix - Invert start and end angles for G02 commands in gcode interpreter

kForth:

This PR fixes a bug in the gcode interpreter that incorrectly calculated the min/max of `G02` commands.
`G2` commands have their start and end angles swapped so they may be treated as `G3` arcs, the angles do not get swapped for `G02` commands.

- Load one of the attached gcode files: [arc_test_files.zip](https://github.com/OctoPrint/OctoPrint/files/9113517/arc_test_files.zip)
- Observe the following model sizes:

| Filename                   | MinX Before | MinX After |
|:----------------------|-------------:|------------:|
| `test_arc_G2.gcode`   |         9.014  |         9.014 |
| `test_arc_G02.gcode` |         -7.089 |         9.014 |
| `test_arc_G3.gcode`   |         -7.089 |       -7.089 |
| `test_arc_G03.gcode` |         -7.089 |       -7.089 |

Not a ticket, but this is a bugfix for PR #4437.

| How G02 arcs are currently analyzed | How G02 arcs should be analyzed |
|:--------------------------------------:|:-----------------------------------:|
| bad_g02_arc | good_g02_arc |

### DIFF
--- a/src/octoprint/util/gcodeInterpreter.py
+++ b/src/octoprint/util/gcodeInterpreter.py
@@ -547,7 +547,7 @@ class gcode:
                 startAngle = math.atan2(oldPos.y - centerArc.y, oldPos.x - centerArc.x)
                 endAngle = math.atan2(pos.y - centerArc.y, pos.x - centerArc.x)
 
-                if gcode == "G2":
+                if gcode in ("G2", "G02"):
                     startAngle, endAngle = endAngle, startAngle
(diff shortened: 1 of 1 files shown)


cp2004 (approved):
👍 Thank you for the PR, I can confirm the issue & fix works as described.

Issue #4582, opened 2022-07-18
G02 or G03 are not interpreted as arcs in the gcodeviewer


cp2004:
### The problem

G2 and G3 commands are correctly rendered as arcs, however G02 and G03 commands are not and just render as direct straight lines. Discovered while testing #4580 

### Contribution help

*This issue may be suitable for a first-time contributor to fix & test*

Check the parsing of movement commands to see why G02/G03 commands are ignored

https://github.com/OctoPrint/OctoPrint/blob/a2e9aefe9c8a51120931e9b14bb95ebf9ea6c7f5/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js#L405-L487

### Additional information & file uploads

[octoprint-systeminfo-20220718155606.zip](https://github.com/OctoPrint/OctoPrint/files/9132967/octoprint-systeminfo-20220718155606.zip)

kForth:
This seems to be a result of the following lines:

https://github.com/OctoPrint/OctoPrint/blob/a2e9aefe9c8a51120931e9b14bb95ebf9ea6c7f5/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js#L482-L485

https://github.com/OctoPrint/OctoPrint/blob/a2e9aefe9c8a51120931e9b14bb95ebf9ea6c7f5/src/octoprint/plugins/gcodeviewer/static/js/viewer/renderer.js#L1074

The `.charAt(1)` in both checks will return `0` for `G02` and `G03` commands, meaning `direction` doesn't get set here and will be `undefined` when it's time to render.

I've created PR #4587 to fix this.

foosel:
1.9.0 has been released.

PR #4587, opened 2022-07-19
Fix/4582 - Fix gcodeviewer rendering G02 and G03 commands as lines

kForth:

This PR is a proposed fix to issue #4582.

The gcodeviewer currently renders G02 and G03 commands as straight lines instead of arcs.

- Upload the test file: [arc_test_G02_G03.gcode.zip](https://github.com/OctoPrint/OctoPrint/files/9137000/arc_test_G02_G03.gcode.zip)
- Observe the gcodeviewer and verify the file renders correctly

This is related to issue #4582.

| How `G02` and `G03` are currently rendered | How `G02` and `G03` should be rendered |
|:-:|:-:|
| Rendered Before | Rendered After |

#### Further notes

You could probably use a simpler regex expression but I decided to keep the style from the test above so that it's more clear.

### DIFF
--- a/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
+++ b/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
@@ -480,8 +480,10 @@ var doParse = function () {
                         center_j = Number(args[j].slice(1));
                         break;
                     case "g":
-                        if (/^(?:G2|G02)(\.\d+)?\s/i.test(line)) direction = -1;
-                        else if (/^(?:G3|G03)(\.\d+)?\s/i.test(line)) direction = 1;
+                        if (args[j].charAt(1) === "2" || args[j].charAt(2) === "2")
+                            direction = -1;
+                        else if (args[j].charAt(1) === "3" || args[j].charAt(2) === "3")
+                            direction = 1;
                         break;
                 }
             }
--- a/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
+++ b/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
@@ -480,8 +480,8 @@ var doParse = function () {
                         center_j = Number(args[j].slice(1));
                         break;
                     case "g":
-                        if (args[j].charAt(1).toLowerCase() === "2") direction = -1;
-                        if (args[j].charAt(1).toLowerCase() === "3") direction = 1;
+                        if (/^(?:G2|G02)(\.\d+)?\s/i.test(line)) direction = -1;
+                        else if (/^(?:G3|G03)(\.\d+)?\s/i.test(line)) direction = 1;
                         break;
                 }
             }


cp2004:
I wonder which is faster, using regexes or just a straight equality check

```js
if (args[j].charAt(1) === "2" || args[j].subString(1, 2) === "02") direction = -1;
if (args[j].charAt(1) === "3" || args[j].subString(1, 2) === "03") direction = 1;
```
(not sure really what the point of lowercasing as previously done was anyway, it is a number...)

We've worked quite a lot to improve the performance of loading files in the gcode viewer recently, such as in #4430 and #4432, and also #4435, so it is on my mind to consider it. If you have a couple of large (like 100mb+) files, I have simply timed the length of time it takes to render before and after changes.

If there is no noticeable impact, then don't worry about it. I suspect it will not be noticeable.

kForth:
Good thought about performance.

I created a jsFiddle to test the raw performance of a few implementations: https://jsfiddle.net/4kj0t7x3/

Below is the amount of time each implementation took for 1 billion checks:

| Implementation | code  | `G2` time (ms) | `G02` time (ms) |
|:-----------------|:------|----------------:|----------------:|
| Original CharAt | `args[j].charAt(1) === "2"`                                                         |     596 |     519 |
| Regex Line         | `/^(?:G2\|G02)(\.\d+)?\s/i.test(line)`                                            | 20481 | 20419 |
| Regex Arg          | `/^G2\|G02/i.test(args[j])`                                                          | 20762 | 21053 |
| Substring           | `args[j].charAt(1) === "2" \|\| args[j].substring(1, 3) === "02")` |     717 | 21168 |
| Substring 2        | `args[j].charAt(1) === "2" \|\| args[j].substring(1) === "02"`      |     721 | 22442 |
| Extra CharAt 1   | `args[j].charAt(1) === "2" \|\| (args[j].charAt(1) === "0" && args[j].charAt(2) === "2")` |   5030 |  5948 |
| Extra CharAt 2   | `args[j].charAt(1) === "2" \|\| args[j].charAt(2) === "2"`             |     710 |    699 |
| Last CharAt*      | `args[j].charAt(args[j].length - 1) === "2"`                                 |     594 |    601 |
| First and Last CharAt* | `args[j].charAt(1) === "2" \|\| args[j].charAt(args[j].length - 1) === "2"` |    697 |    699 |
| EndsWith**        | `args[j].endsWith("2")`                                                               | 17309 | 17391 |
| Slice*                  | `args[j].slice(-1) === "2"`                                                          | 19631 | 20018 |

\* doesn't work with gcode subcommands

\** requires ES6 support

I also used Firefox's performance profiler to test some of the implementations on a ~145MB file.

Below is the amount of time each implementation spent in the `gcodeviewer` `worker.js` `doParse` function:

| Implementation | No G2/G3 Moves (ms) | All G2/G3 Moves (ms) | All G02/G03 Moves (ms) |
|:-----------------|------------------------:|-----------------------:|--------------------------:|
| Original             |                            2975 |                          2957 |                               3012 |
| Regex Line         |                            3378 |                          3147 |                               3001 |
| Substring           |                            2922 |                          2914 |                               3083 |
| Extra CharAt 2   |                            3078 |                          2986 |                               2942 |

There doesn't seem to be too much of a real-world difference but the regex method is definitely slower.

It looks like `Extra CharAt 2` is the quickest in raw performance.

Thoughts?

cp2004:
The 'Extra Char 2' option doesn't sacrifice code readability in the name of performance in my opinion. That's the main thing to avoid, is that it becomes harder to understand but 'optimzed'! That would be a good solution for me. The results are close enough that I am OK with any of them, I'm just keeping an eye on performance when processing large data volumes.

Great numbers and tests on it, I found that really easy to understand.

src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
```diff
@@ -480,8 +480,10 @@ var doParse = function () {
                         center_j = Number(args[j].slice(1));
                         break;
                     case "g":
-                        if (args[j].charAt(1).toLowerCase() === "2") direction = -1;
-                        if (args[j].charAt(1).toLowerCase() === "3") direction = 1;
+                        if (args[j].charAt(1) === "2" || args[j].charAt(2) === "2")
+                            direction = -1;
+                        else if (args[j].charAt(1) === "3" || args[j].charAt(2) === "3")
+                            direction = 1;
```
foosel, line 486: That needs a further restriction to `args[j].charAt(1) === "0"` or something like `G12` could match, leading to unexpected results.

And instead of two `charAt` comparisons I'd honestly rather go with

```suggestion
                        if (args[j].charAt(1) === "2" || args[j].substring(1, 2) === "02")
                            direction = -1;
                        else if (args[j].charAt(1) === "3" || args[j].substring(1, 2) === "03")
                            direction = 1;
```

foosel (changes requested):
The current change has unexpected side effects if I'm not mistaken, see below.

src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
cp2004: That should be fine, as by the time it gets to here it must already be G0/G(0)1/G(0)2/G(0)3, in line 405

foosel: Ah, I overlooked that context. Thanks for the heads-up, and I retract my statement in that case ^^

foosel (approved):

Issue #3879, opened 2020-12-11
[Request] Better error message for failed uploads


Stretch96:
Currently if an upload fails, it only prompts to ensure that the file ends in a valid extension (eg. `.gcode`)

I've only recently noticed this error message, because I have OctoPrint behind Nginx

It'd be nice to also prompt the user to check any Nginx (or any other Proxy) settings, to ensure there's not an upload size limit

(This was the issue I ran into, as Nginx has quite a small limit by default)

cp2004:
So it showed you an error message that was incorrect? Just clarifying as I was unsure what exactly happened here.

Prompting the user to check proxy settings sounds annoying to me, to many people who may have configured their proxy correctly - and all those using OctoPi, some of which probably don't even know haproxy exists and would panic at the sight of it.

We do also have some [reverse proxy configuration examples](https://community.octoprint.org/t/reverse-proxy-configuration-examples/1107?u=charlie_powell) that you might want to take a look at.

Stretch96:
I think it's that the warning message, about having the wrong extension, is probably the least likely cause of a failed upload

Maybe the message could include the HTTP response code, which would hold more clues as to why an upload failed

PR #4588, opened 2022-07-19
Improve - Better error handling for uploading and removing files and folders

### LINKED ISSUE #655: [Request] Better error handling when disk is full during file uploads
a) The "upload failed" message in a "disk full" event is the same as when uploading an invalid file: 
Upload failed
Could not upload the file. Make sure that it is a GCODE file and has the extension ".gcode" or ".gco" or that it is an STL file with the extension ".stl" and slicing support is enabled and configured.
More: 
undefined

b) Optional: There is no preemptive warning on low disk space. 

Version: 1.2.0-dev-300-g048ad78 (devel branch)

I love cookies, but i'm not sure what kind of bug report / feature request this would ideally be. So i guess it's [Misc] for now.

kForth:

This PR improves the error handling for file/folder upload and removal.
This PR also prevents uploading a file if there's not enough free disk space (can still overwrite).

This PR solves 2 main issues:
- If a file or folder is currently inaccessible and OctoPrint fails to remove it, it will do it silently.
- If a file or folder upload fails OctoPrint shows a notification telling the user to check the file type, regardless of the actual issue.

To test the file size limit:
 - Limit disk space
   - Set the uploads folder path to a small USB flash drive
  or
   - Run `octoprint serve` with `--basedir` set to a small USB flash drive.
 - Try to upload a file that's too large thru the web ui

To cause a failed file upload:
 - Set an already uploaded file to read-only
 - Try to overwrite the file thru the web ui

To cause a failed file removal:
 - Set an already uploaded file to read-only
 - Try to remove file thru the web ui

To reproduce the original issue that caused file removal to fail (only on windows apparently):
- Upload a gcode file to octoprint and quickly click 'Remove' as soon as it's uploaded.
  - It helps if the file is large so analyzing takes longer, here is a sample file: [TestFile.gcode.zip](https://github.com/OctoPrint/OctoPrint/files/9143696/TestFile.gcode.zip)
- Observe that nothing changes in the web ui
  - OctoPrint log/terminal should show errors

My original issue was that file removal failed because the file was locked by the gcode interpreter process, but there are other reasons a file could be inaccessible to OctoPrint at any given moment.

This PR resolves #655 and #3879.

| Overwrite Confirmation | No Overwrite Confirmation |
|:-:|:-:|
| Screen Shot 2022-07-21 at 11 42 22 PM | size-limit-no-overwrite-confirm |

| Failed Upload | Failed Removal |
|:-:|:-:|
| upload-failed-notification | removal-fail-notification |

#### Further notes

Thoughts are appreciated.

Please suggest improvements for the various error message texts.

### DIFF
--- a/src/octoprint/filemanager/storage.py
+++ b/src/octoprint/filemanager/storage.py
@@ -740,7 +740,7 @@ class LocalFileStorage(StorageInterface):
         ):
             raise StorageError(
                 f"{destination_name} does already exist in {destination_path}",
-                code=StorageError.INVALID_DESTINATION,
+                code=StorageError.ALREADY_EXISTS,
             )
 
         source_meta = self._get_metadata_entry(source_path, source_name)
--- a/src/octoprint/server/api/files.py
+++ b/src/octoprint/server/api/files.py
@@ -661,11 +661,8 @@ def uploadGcodeFile(target):
                 allow_overwrite=True,
                 display=canonFilename,
             )
-        except octoprint.filemanager.storage.StorageError as e:
-            if e.code == octoprint.filemanager.storage.StorageError.INVALID_FILE:
-                abort(415, description="Could not upload file, invalid type")
-            else:
-                abort(500, description="Could not upload file")
+        except (OSError, StorageError) as e:
+            _abortWithException(e)
         else:
             filename = fileProcessingFinished(
                 added_file,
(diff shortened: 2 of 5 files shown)


kForth:
I've discovered that the file inaccessibility issue I was having happens on Windows 10 but not on MacOS.
On MacOS, OctoPrint is able to remove the file immediately even if the analyzer is running, I assume due to the way file locking is handled differently.

kForth:
Converting this to Draft because I realized I can expand this slightly to address a number of open issues.

foosel (approved):

Issue #3341, opened 2019-11-19
REQUEST: CR-10 v2 Filament Detector Support


KenLucke:
The current version of OctoPrint (using v1.3.12) apparently does not support the built-in filament breakage/end detector that the CR-10 v2 comes with natively.  When printing from the SD card, the machine will pause and alert you when the filament runs out or breaks.  When printing from OctoPrint, it does not.  Either that, or I'm experiencing an Eye-Dee-Ten-Tee error and cannot find the function.

Can someone please add the support for this to the current version, or add a plugin that covers it?

GitIssueBot:
Hi @KenLucke,

It looks like there is some **information missing** from your bug report that will be needed in order to solve the problem. Read the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md) which will provide you with a template to fill out here so that your bug report is ready to be investigated (I promise I'll go away then too!).

If you did not intend to report a bug but wanted to **request a feature or brain storm** about some kind of development, please take special note of the title format to use as described in the [Contribution Guidelines](https://github.com/foosel/OctoPrint/blob/master/CONTRIBUTING.md).

**Please do not abuse the bug tracker as a support forum** - that can be found at [discourse.octoprint.org](https://discourse.octoprint.org). Go there for any kind of issues with network connectivity, webcam functionality, printer detection or any other kind of such support requests or general questions.

Also **make sure you are at the right place** - this is the bug tracker of the official version of OctoPrint, not the Raspberry Pi image OctoPi nor any unbundled third party OctoPrint plugins or unofficial versions. Make sure too that you have **read through the [Frequently Asked Questions](http://faq.octoprint.org)** and searched the [**existing tickets**](https://github.com/foosel/OctoPrint/search?q=&ref=cmdform&type=Issues) for your problem - try multiple search terms please.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next two weeks (until 2019-12-03 08:20 UTC) I'll close this ticket so it doesn't clutter the bug tracker. This is nothing personal, so please just be considerate and help the maintainers solve this problem quickly by following the guidelines linked above. Remember, the less time the devs have to spend running after information on tickets, the more time they have to actually solve problems and add awesome new features. Thank you!

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being, so don't expect any replies from me :) Your ticket is read by humans too, I'm just not one of them.

[...]

PR #4590, opened 2022-07-20
Fire action hooks on ALL //action messages, not just unhandled ones

arrdem:

This PR reworks how user-defined handlers for slash actions are fired to enable userland hooks to spy on and extend the existing support for `//action:paused` and friends.

Installed and tested on my local OctoPrint instance. Since this patch will cause action handlers to be fired more often there are potential performance concerns here, but as action handlers are already expected to implement their own filtering and dispatch I suggest this is not a breaking API change.

When built with runout support, MarlinFW will emit the following action over the serial console:

```
//action:paused filament_runout T0
```

the interpretation of this action as a pause was initiated by runout while using tool 0. At present, OctoPrint correctly interprets this action as a pause. But it does NOT interpret this action as a source for the `Events.FILAMENT_CHANGE` event. This means that it isn't possible to hang other automation or listeners off of `Events.FILAMENT_CHANGE` event correctly because this pause is not interpreted as a filament change.

As originally implemented user-defined handlers run in the terminal `else:` block of a long `if`/`elif` chain. This means they will be fired ONLY IF no handler ran, which precludes extending OctoPrint's behavior in this particular case without patching `comm.py` one way or another.

This patch simply makes executing user-defined handlers unconditional. All handlers will always be fired on any action. This is not a change of contract or semantics as handlers already expect to receive any action and must decide whether they care.

It would appear that the capability I've built with this and a plugin was previously requested in #3341, but that was never actioned one way or another.

#### Further notes

This patch enables me to use https://github.com/arrdem/octoprint-runout-event-plugin + the Matrix event handler plugin to treat runouts as an explicit event I can be notified of, rather than notifying on pauses.

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -2401,25 +2401,25 @@ class MachineCom:
                                 level=logging.INFO,
                             )
                             self.refreshSdFiles()
-                        else:
-                            for name, hook in self._printer_action_hooks.items():
-                                try:
-                                    hook(
-                                        self,
-                                        line,
-                                        action_command,
-                                        name=action_name,
-                                        params=action_params,
-                                    )
-                                except Exception:
-                                    self._logger.exception(
-                                        "Error while calling hook from plugin "
-                                        "{} with action command {}".format(
-                                            name, action_command
-                                        ),
-                                        extra={"plugin": name},
-                                    )
-                                    continue
+
+                        for name, hook in self._printer_action_hooks.items():
+                            try:
+                                hook(
+                                    self,
+                                    line,
+                                    action_command,
+                                    name=action_name,
+                                    params=action_params,
+                                )
+                            except Exception:
+                                self._logger.exception(
+                                    "Error while calling hook from plugin "
+                                    "{} with action command {}".format(
+                                        name, action_command
+                                    ),
+                                    extra={"plugin": name},
+                                )
+                                continue
 
                     if self._state not in (
                         self.STATE_CONNECTING,


cp2004:
> What does this PR do and why is it necessary?
This PR reworks how user-defined handlers for slash actions are fired to enable userland hooks to spy on and extend the existing support for //action:paused and friends.

This makes sense to me. From a performance perspective, action commands are not expected to be used in a time-critical application in the middle of a print, so don't worry about that. If you have a slight delay pausing/cancelling/starting the print etc. then that's OK in the grand scheme of things.

If you wanted to, you can contribute the functionality of your plugin (triggering an event on `filament_runout`) directly to OctoPrint as well - it is a very small change that doesn't seem like it needs to be in a plugin in my opinion.

cp2004:
Also, it doesn't need to be targeted to `devel`, `maintenance` is fine so this change can be shipped with 1.9.0.

arrdem:
> If you wanted to, you can contribute the functionality of your plugin (triggering an event on `filament_runout`) directly to OctoPrint as well - it is a very small change that doesn't seem like it needs to be in a plugin in my opinion.

Sounds good. I'll get this cherry-picked and targeting `maintenance` later today.

Let's do just this change and I'll circle back on contributing `filament_runout` support. I will need to spend a bit of time with my tool and marlinfw once the current job finishes to document the `//action:paused filament_change` vs `//action:filament_change` behavior I thought I observed yesterday (the difference I think being between `M412 H0` or `M412 H1`) and looking at how the pause machinery generates events.

At present that plugin implementation causes a `Events.FILAMENT_RUNOUT` followed within the same second by an `Events.PAUSED` due to some interactions between the exact handler script in Marlin and how the `comm.py` state machine is implemented. I don't wanna just bundle that change in here without 1) making sure it does what I think it does 2) agreement from one of y'all about whether `Events.FILAMENT_RUNOUT` implies `Events.PAUSED` or not.

Issue #3620, opened 2020-06-29
Please handle filament runout other than "paused->timeout"


AndKe:
This is the current experience of filament runout:

```
Send: N26 G1 X248.200 Y56.325 E23.27269*84
Recv: ok
Send: N27 G1 X250.274 Y56.722 E23.46472*87
Recv:  T:210.20 /210.00 B:59.98 /60.00 @:28 B@:0
Recv: echo:busy: processing
Recv: //action:paused filament_runout 0
Recv:  T:210.27 /210.00 B:60.04 /60.00 @:27 B@:0
Recv: echo:busy: processing
Recv: ok
Send: N28 G1 X252.056 Y57.855 E23.65675*82
Recv: //action:paused
Printer signalled that it paused, switching state...
Changing monitoring state from "Printing" to "Pausing"
Recv:  T:210.35 /210.00 B:59.99 /60.00 @:26 B@:127
Recv: echo:busy: processing
Recv:  T:210.31 /210.00 B:60.08 /60.00 @:26 B@:127
Recv: echo:busy: processing
Recv:  T:210.31 /210.00 B:60.30 /60.00 @:26 B@:0
Recv: echo:busy: processing
Recv:  T:210.12 /210.00 B:60.28 /60.00 @:29 B@:0
..............................(and about a minute or so  later.........)
Recv: echo:busy: processing
Recv:  T:210.31 /210.00 B:60.19 /60.00 @:28 B@:0
Recv:  T:210.00 /210.00 B:60.34 /60.00 @:34 B@:0 W:0
Communication timeout while printing, trying to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
Changing monitoring state from "Pausing" to "Paused"
Send: M105
Communication timeout while printing, trying to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
Send: M105
No response from printer after 3 consecutive communication timeouts, considering it dead. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
Changing monitoring state from "Paused" to "Offline (Error: Too many consecutive timeouts, printer still connected and alive?)"
Connection closed, closing down monitor

```
The job is correctly paused (while the Marlin firmware would wait for filament change)  - but then the communication stops and the print is killed loooong before one can change the filament.

GitIssueBot:
Hi @AndKe,

[...]

PR #4591, opened 2022-07-21
Fire action hooks on ALL actions, not just unhandled ones

arrdem:
This PR supersedes #4590 with the same changeset targeting `maintenance` per @cp2004. A new PR was required in order to change the source branch. 

---

This PR reworks how user-defined handlers for slash actions are fired to enable userland hooks to spy on and extend the existing support for `//action:paused` and friends.

Installed and tested on my local OctoPrint instance. Since this patch will cause action handlers to be fired more often there are potential performance concerns here, but as action handlers are already expected to implement their own filtering and dispatch I suggest this is not a breaking API change.

When built with runout support, MarlinFW will emit the following action over the serial console:

```
//action:paused filament_runout T0
```

the interpretation of this action as a pause was initiated by runout while using tool 0. At present, OctoPrint correctly interprets this action as a pause. But it does NOT interpret this action as a source for the `Events.FILAMENT_CHANGE` event. This means that it isn't possible to hang other automation or listeners off of `Events.FILAMENT_CHANGE` event correctly because this pause is not interpreted as a filament change.

As originally implemented user-defined handlers run in the terminal `else:` block of a long `if`/`elif` chain. This means they will be fired ONLY IF no handler ran, which precludes extending OctoPrint's behavior in this particular case without patching `comm.py` one way or another.

This patch simply makes executing user-defined handlers unconditional. All handlers will always be fired on any action. This is not a change of contract or semantics as handlers already expect to receive any action and must decide whether they care.

It would appear that the capability I've built with this and a plugin was previously requested in #3341, but that was never actioned one way or another.

#### Further notes

This patch enables me to use https://github.com/arrdem/octoprint-runout-event-plugin + the Matrix event handler plugin to treat runouts as an explicit event I can be notified of, rather than notifying on pauses.


jneilliii:
additional context in #3341 and #3620

arrdem:
> Just curious @arrdem if you could achieve your end goal with the already existing [Action Commands](https://plugins.octoprint.org/plugins/actioncommands/) plugin or am I missing something?

Almost!

The "problem" I'm trying to solve is that my Marlin 2.0.9.3 + CR-10v3 + `#define ACTION_ON_FILAMENT_RUNOUT` + `#define ACTION_REASON_ON_FILAMENT_RUNOUT` build exhibits the following behavior when a runout occurs:

```
//action:paused filament_runout T0
//action:paused
```

Neither of these actions generate a `Event.FILAMENT_RUNOUT` in `comm.py` as currently implemented, nor am I able to hook into either one using `octoprint.comm.protocol.action` because both appear to `comm.py` to be `//action:paused` and hooks are not run. OctoPrint just pauses twice in a row.

All of this is in service of being able to build a three-liner plugin -

```
    def custom_action_handler(self, comm, line, command, *args, **kwargs):
        self._logger.info(f"Got // line: {line!r}, comm: {comm!r}, command: {command!r}")
        if command.find("filament_runout") != -1:
           self._event_bus.fire(Events.FILAMENT_CHANGE)
```

which will yes drive the existing Actions plugin or in my case the Matrix-Notifier plugin as they are presently implemented.

All of this is in response to trying the Matrix-Notifier plugin, seeing that it generated `Paused` notifications on runout events when I tested, going "that's weird" and falling down a rabbit hole here.

foosel (approved):

foosel:
Very sensible and minimal change that I have no issue with merging! Thank you!

PR #4597, opened 2022-07-30
Add aria labels and roles for accessibility

dawidpieper:
OctoPrint uses ```title``` attribute for describing UI elements, but this behavior is hardly accessible or even inaccessible for blind users using screenreaders.
This PR adds aria labels and roles for the most important parts of interface to enable screenreaders to properly label such elements.

Tested using NVDA screenreader on Windows, Voice Over on Mac and iOS, Commentary Screenreader on Android and Orca on Linux.

I haven't fixed official plugins yet, also I may have overlooked some elements. But the most important parts are labeled.

N/A

N/A

#### Further notes
N/A


cp2004:
Just FYI, there may be a delay in PR reviews as Gina is on vacation and there is now a small backlog. It seems like a good idea to me, I don't think there will be an issue if you also wanted to do the bundled plugins.

dawidpieper:
Thank you for the positive reception.
I added labels to the bundled plugins. Some plugins do not yet have aria descriptions of icons, but these are not critical needs anywhere and the problem is rather secondary, worth taking care of, but not necessarily a priority. Most importantly, OctoPrint is now basically accessible.

foosel (approved):
Long term I'll look if this maybe be achieved in a way that requires less code/text duplication, but merging this is as-is a no-brainer given the outcome! Thank you!

PR #4605, opened 2022-08-03
Recognize sd card initialization on Repetier fw

arekm:
Repetier firmware uses different text for sd card initialization:

Send: N10 M21*33
Recv: ok 10
Recv: Card successfully initialized.

Recognize that in octoprint, so sd card operations become available in
UI.

Tested with Repetier firmware 1.0.4.

    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

### DIFF
--- a/src/octoprint/util/comm.py
+++ b/src/octoprint/util/comm.py
@@ -2798,11 +2798,6 @@ class MachineCom:
                                 supportRepetierTargetTemp = True
                                 disable_external_heatup_detection = True
 
-                                sd_always_available = self._sdAlwaysAvailable
-                                self._sdAlwaysAvailable = True
-                                if not sd_always_available and not self._sdAvailable:
-                                    self.initSdCard()
-
                             elif "reprapfirmware" in firmware_name.lower():
                                 self._logger.info(
                                     "Detected RepRapFirmware, enabling relevant features for issue free communication"
@@ -2928,11 +2923,15 @@ class MachineCom:
                         "SD init fail" in line
                         or "volume.init failed" in line
                         or "openRoot failed" in line
+                        or "SD Card unmounted" in line
+                        or "SD card released" in line
                     ):
                         self._sdAvailable = False
                         self._sdFiles = []
                         self._callback.on_comm_sd_state_change(self._sdAvailable)
-                    elif "SD card ok" in line and not self._sdAvailable:
+                    elif (
+                        "SD card ok" in line or "Card successfully initialized" in line
+                    ) and not self._sdAvailable:
                         self._sdAvailable = True
                         self.refreshSdFiles()
                         self._callback.on_comm_sd_state_change(self._sdAvailable)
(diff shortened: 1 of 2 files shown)


foosel (changes requested):
As things are right now, if Repetier firmware is detected and automatic firmware detection is enabled, OctoPrint will always assume that an SD is available, so this PR would turn into a no-op. 

If we go for native support for Repetier SD card messaging and handling, that part of the autodetection needs to be removed (see lines 2787 and following in `comm.py`), and matching of error messages also needs to be added. As far as I can say looking at the firmware source, the relevant strings for init failure and such are [here](https://github.com/repetier/Repetier-Firmware/blob/7cb3741323ba45537eceb4d54e99b1cbdedf2739/src/ArduinoAVR/Repetier/SDCard.cpp#L93-L112) and for umounting [here](https://github.com/repetier/Repetier-Firmware/blob/7cb3741323ba45537eceb4d54e99b1cbdedf2739/src/ArduinoAVR/Repetier/SDCard.cpp#L146).

arekm:
In my case plugin does M21 command which ends up with "Recv: SD init fail". In result octoprint sees that and disables sd card. 

Then another M21 gets "Card successfully initialized" and that's not recognized by octoprint, so sd card support stays disabled.

But I can also do what you suggest and go fully for "native support".

foosel:
Ah, yeah, you are right, I didn't spot [this line](https://github.com/repetier/Repetier-Firmware/blob/7cb3741323ba45537eceb4d54e99b1cbdedf2739/src/ArduinoAVR/Repetier/SDCard.cpp#L91) where that is generated and which always gets generated for any of the error cases. So that should indeed be covered. In that case I think we only the `sdAlwaysAvailable` handling removed in the auto detection routine. Detecting an unmount would be nice-to-have though ([see generated message in this line](https://github.com/repetier/Repetier-Firmware/blob/7cb3741323ba45537eceb4d54e99b1cbdedf2739/src/ArduinoAVR/Repetier/SDCard.cpp#L146)) ;)

foosel:
Also, please make sure to install `pre-commit`, your changes currently fail some checks.

arekm:
Suggestions implemented.

(no idea why github check still says "1 failing"; pre-commit run --hook-stage manual --all-files passes here locally)

foosel:
That would be the E2E tests that are sadly turning more and more flaky on CI and are driving me slowly insane. Not your fault, big todo on my own task list :)

src/octoprint/util/comm.py
```diff
@@ -2928,11 +2923,15 @@ def convert_line(line):
                         "SD init fail" in line
                         or "volume.init failed" in line
                         or "openRoot failed" in line
+                        or "SD Card unmounted" in line
+                        or "SD card released" in line
```
foosel, line 2927: Where did you find that string? I'm looking at the Marlin source but so far haven't found anything similar in there (doesn't even seem to have any string for umounting sadly).

arekm: [here](https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/cardreader.cpp#L267)

foosel: Ah, Prusa Firmware, gotcha, thanks!

foosel (approved):

PR #4608, opened 2022-08-04
✏️ Fix several typos

luzpaz:
Found via `codespell -q 3 -S *.po`

### DIFF
--- a/docs/plugins/gettingstarted.rst
+++ b/docs/plugins/gettingstarted.rst
@@ -403,8 +403,8 @@ Remember that Wikipedia link we added to our little link in the navigation bar? 
 what if we want to allow our users to adjust that according to their wishes, e.g. to link to the German language node
 about "Hello World" programs instead?
 
-To allow your users to customized the behaviour of your plugin you'll need to implement the :class:`~octoprint.plugin.SettingsPlugin`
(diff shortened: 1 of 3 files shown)


luzpaz:
Updated description by editing the 1st post of the thread

cp2004 (changes requested):
Codespell is now setup in pre-commit to run against the repository. If the configuration can be improved, please update the configuration to contribute to the future of the project. 

I am not a fan of getting multiple PRs all 'fixing typos', yes I can see that the typos are there but it would be better to leave it to the automated tools instead of spending our time reviewing these PRs.

luzpaz:
Makes sense. Just note that the typos mentioned may not be found just yet since codespell hasn't had an update in a while. That means the dictionaries it uses aren't up to date.  Hopefully that changes. Feel free to do as you will, of course. Please advise

foosel:
I agree with @cp2004 here, this should be detected automatically, and if it isn't yet being detected automatically the tooling needs to be improved to detect it automatically (e.g. through a PR!). We've recently been swamped with typo fix PRs and as piecemeal as they are filed they frankly add more overhead than value given our limited review resources.

luzpaz:
fair enough

PR #4610, opened 2022-08-04
M20 T support (timestamps for files on sd card)

arekm:
M20 T support (timestamps for files on sd card).

    Use "M20 L T" (if CAPABILITY_EXTENDED_M20 is available) which gives
    us long file names but also file timestamps.

    Wire up that timestamp into "date" field which gives us "Uploaded: ...
    ago" for files on SD card. Sorting by upload date also works.

    Implementing this also fixes parsing of response to (for example
    manually sent) "M20 T" command. Octoprint assumed that third entry
    in response is always long file name while with M20 T it is timestamp.

### DIFF
--- a/src/octoprint/plugins/virtual_printer/virtual.py
+++ b/src/octoprint/plugins/virtual_printer/virtual.py
@@ -15,6 +15,7 @@ from serial import SerialTimeoutException
 
 from octoprint.plugin import plugin_manager
 from octoprint.util import RepeatedTimer, get_dos_filename, to_bytes, to_unicode
+from octoprint.util.files import unix_timestamp_to_m20_timestamp
 
 
 # noinspection PyBroadException
@@ -126,6 +127,7 @@ class VirtualPrinter:
 
         self._writingToSd = False
         self._writingToSdHandle = None
+        self._writingToSdFile = None
         self._newSdFilePos = None
 
         self._heatingUp = False
@@ -256,6 +258,7 @@ class VirtualPrinter:
                     pass
             self._writingToSd = False
             self._writingToSdHandle = None
+            self._writingToSdFile = None
             self._newSdFilePos = None
 
             self._heatingUp = False
@@ -1414,8 +1417,7 @@ class VirtualPrinter:
                 "path": entry.path,
                 "dosname": dosname,
                 "size": entry.stat().st_size,
-                # 0 for now
-                "timestamp": hex(0),
+                "timestamp": hex(unix_timestamp_to_m20_timestamp(entry.stat().st_mtime)),
             }
         return result
 
@@ -1793,6 +1795,7 @@ class VirtualPrinter:
         except Exception:
             self._send("error writing to file")
         self._writingToSdHandle = handle
+        self._writingToSdFile = file
         self._writingToSd = True
         self._selectedSdFile = file
         self._send("Writing to file: %s" % filename)
@@ -1806,6 +1809,12 @@ class VirtualPrinter:
             self._writingToSdHandle = None
         self._writingToSd = False
         self._selectedSdFile = None
+        # Most printers don't have RTC and set some ancient date
+        # by default. Emulate that using 2000-01-01 01:00:00
+        # (taken from prusa firmware behaviour)
+        st = os.stat(self._writingToSdFile)
+        os.utime(self._writingToSdFile, (st.st_atime, 946684800))
+        self._writingToSdFile = None
         self._send("Done saving file")
 
     def _sdPrintingWorker(self):
--- a/src/octoprint/plugins/virtual_printer/virtual.py
+++ b/src/octoprint/plugins/virtual_printer/virtual.py
@@ -1417,7 +1417,7 @@ class VirtualPrinter:
                 "path": entry.path,
                 "dosname": dosname,
                 "size": entry.stat().st_size,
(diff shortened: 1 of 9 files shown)


arekm:
Printers don't have real time clock and on M28 upload they set ancient date (like 2000-01-01 01:00:00 - prusa mk3s does that) which then shows up in octoprint as "Uploaded: 23 years ago".

Printer tells us such timestamp in M20 T, so it is being shown but maybe it would be better to ignore such dates and show nothing(?)

Virtual printer emulates this behaviour, too. If you want updated timestamp just touch file in ~/.octoprint/virtualSd/.

timestamps on vfat filesystems (used on most sd cards) are considered to be in "local time" (this is behaviour of windows and macos). Linux unfortunately considers these UTC, so if someone mounts sd card on Linux (with default options), puts gcode files there and then put sd card into printer there will be offset shown, but we can't do nothing about this. If someone does the same on windows or macos then printer and octoprint will show correct upload time.

foosel:
> Printer tells us such timestamp in M20 T, so it is being shown but maybe it would be better to ignore such dates and show nothing(?)

I've always followed the approach that what the firmware reports gets used. Otherwise we'll get "bug reports" about missing timestamps. This way of course we'll risk "bug reports" about invalid timestamps, but at least we can then just point people to this PR and explain that we just display what the firmware reports and that it is up to the firmware to make a decision here (e.g. NOT return a timestamp for files created directly by the firmware).

src/octoprint/util/files.py
```diff
@@ -270,3 +271,59 @@ def silent_remove(file):
...
+    d = datetime.date(year, month, day)
+
+    tm = timestamp & (2**16 - 1)
+    second = (tm & (1 << 5) - 1) * 2
+    minute = (tm >> 5) & ((1 << 6) - 1)
+    hour = (tm >> 11) & ((1 << 5) - 1)
+    t = datetime.time(hour, minute, second)
+
+    combined_dt = datetime.datetime.combine(d, t)
+    return combined_dt.timestamp()
```
foosel, line 329: Please add unit tests for these.

Use [parameterized pytest fixtures](https://docs.pytest.org/en/6.2.x/parametrize.html) (I'm slowly migrating over to these from `ddt`).

src/octoprint/util/comm.py
```diff
@@ -2438,30 +2439,54 @@ def convert_line(line):
...
+                size = None
+                timestamp = None
                 # if we are currently receiving an sd file list, each line is just a filename, so just read it and abort processing
                 if self._sdEnabled and self._sdFileList and "End file list" not in line:
                     preprocessed_line = line
-                    fileinfo = preprocessed_line.split(None, 2)
-                    if len(fileinfo) == 3:
-                        # name, size, long name
-                        filename, size, longname = fileinfo
+                    fileinfo = preprocessed_line.split(None, 3)
```
foosel, line 2448: All of this is starting to get a complexity that it really should be moved into a dedicated parser function that has a ton of unit test coverage. I can do that after I merge this, or alternatively you look into it as part of this PR. What's your preference?

src/octoprint/plugins/virtual_printer/virtual.py
```diff
@@ -14,7 +14,13 @@
...
 
 from octoprint.plugin import plugin_manager
-from octoprint.util import RepeatedTimer, get_dos_filename, to_bytes, to_unicode
+from octoprint.util import (
+    RepeatedTimer,
+    get_dos_filename,
+    to_bytes,
+    to_unicode,
+    unix_timestamp_to_m20_timestamp,
+)
```
foosel, line 23: `unix_timestamp_to_m20_timestamp` should be imported from `octoprint.util.files`, see other note.

foosel (changes requested):
Got some review annotations, mostly regarding the structure and test coverage.

foosel:
Please make sure to request a new review once you've implemented the requested changes.

foosel:
I sadly only just now noticed that you opened this PR against `devel`. Since this is not a backwards incompatible new feature but rather an improvement of an existing one, this should be opened against `maintenancen` instead. Also see the [contribution guidelines](https://github.com/OctoPrint/OctoPrint/blob/master/CONTRIBUTING.md#pull-requests) and [the dev docs](https://docs.octoprint.org/en/master/development/branches.html).

src/octoprint/util/comm.py
arekm: Please do. I'm occasional programmer and unfamiliar with pytest (hope the tests I added are ok).

arekm:
Should be now against maintenance branch (and I saw docs but wasn't sure what printers do if T is unrecognized. Fortunately they just ignore it).

arekm:
Moved parsing into separate function, added tests.

foosel:
Please make sure to request a new review once you're done working on this and it is ready for another review.

arekm:
I see "foosel was requested for review" and I don't see any new way to request review again in github UI.

At this moment I have no pending changes/improvements.

foosel:
I just kept seeing new and new pushes, so I wanted to make sure you really are done here before I give this another review. It's a bit tricky to hit a moving target ;)

foosel (approved):

foosel:
Looks good now, merging, thank you :)

PR #4613, opened 2022-08-08
Show multiple whitespace in filename.

arekm:
Long filename like <pre>more   than   one   space.gcode</pre> was shown in UI as
<pre>more than one space.gcode</pre> Fix that.


foosel:
> Fix that.

Why?

arekm:
Because these are two different filenames.

```
# ls -ld more*
-rwxr-xr-x 1 root root 0 Aug 15 15:05 'more   than   one   space.gcode'
-rwxr-xr-x 1 root root 0 Aug 15 15:05 'more than one space.gcode'
```

different-files

PR #4618, opened 2022-08-10
Restore doesn't restore plugins versions.

arekm:
It isn't obvious that restore doesn't restore plugin version, so document that.


docs/bundledplugins/backup.rst
```diff
@@ -208,3 +209,5 @@ its source repository under ``src/octoprint/plugins/backup``.
 .. [#2] Note that only those plugins that are available on `OctoPrint's official plugin repository <https://plugins.octoprint.org>`_
         can be automatically restored. If you have plugins installed that are not available on there you'll get their
         names and - if available - homepage URL displayed after restore in order to be able to manually reinstall them.
+.. [#3] Note that restore process will install latest possible version of plugins instead of version that was installed at backup
```
cp2004, line 212:
```suggestion
.. [#3] Note that the restore process will install the latest possible versions of the plugins instead of the version that was installed at backup
```

cp2004 (changes requested):
Some readability changes suggested in the comments.

arekm:
Thanks. Should be better now.

cp2004 (approved):

PR #4620, opened 2022-08-11
:lipstick: Add tooltip to disabled plugin manager install

### LINKED ISSUE #4611: [Request] Install plugin button is not obviously disabled

_No response_

i was trying to install a plugin for the first time and there is currently a print going

I did not notice the warning saying that plugin management will be disabled while printing right away and spent 5 or so minutes debugging, different browser, searching the internet, reading the HTML source only to finally see the warning and realize that the buttons are intentionally disabled. 

Can the buttons be made more obviously disabled? I suppose it's the theme's doing but the buttons look enabled to me 

Or, there should be a tooltip saying that it's disabled for x reason 

image

_No response_

### Additional context

_No response_

cp2004:
Closes #4611

The contrast between enabled/disabled is not great across the board for Bootstrap 2 when there is no enabled button beside it to compare with.

### DIFF
--- a/src/octoprint/plugins/pluginmanager/templates/pluginmanager_settings.jinja2
+++ b/src/octoprint/plugins/pluginmanager/templates/pluginmanager_settings.jinja2
@@ -266,7 +266,7 @@
                         <div class="row-fluid">
                             <div class="span12">
                                 <div class="span3 pull-right">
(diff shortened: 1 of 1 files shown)


foosel (commented):
The attribute needs to be wrapped in `gettext` to become translatable. I'm wondering though if we should maybe rather change the button text altogether?

cp2004:
I was wondering about changing the text, but at the same time I couldn't think of something better that was still short. I'll play around with some things this evening.

cp2004:
I couldn't think of any words that would not go over two lines that would solve this issue so far...

foosel:
"Disabled"? 🤔

foosel (approved):

PR #4625, opened 2022-08-20
Add __str__ method for StorageError exception.

arekm:
Add __str__ method to get message when exception is handled like:

"octoprint.filemanager.storage.StorageError: /tmp/test is not a
valid directory"


src/octoprint/filemanager/storage.py
```diff
@@ -453,6 +453,9 @@ def __init__(self, message, code=None, cause=None):
             code = StorageError.UNKNOWN
         self.code = code
 
+    def __str__(self):
+        return self.message
+
```
foosel, line 458: I think instead of reimplementing `__str__` it makes more sense to adjust the constructor to something like this:

``` python
def __init__(self, message, code=None, cause=None):
  Exception.__init__(self, message)
  # ...
```

That way the inheritance is correct (`BaseException` here is actually a bug) and we reuse the default `__str__` handler of `Exception`.

foosel (changes requested):

src/octoprint/filemanager/storage.py
arekm: Looked if self.message is used outside this class but can't find anything, so dropped it, too.

Issue #4300, opened 2021-11-04
[BUG] Multiple file uploads seemingly possibly on drag'n'drop, but not working properly


bolsoncerrado:
### The problem

I try to upload two (or more) files which already exist on the upload folder of octoprint.
For the first uploaded file, I'm asked if I want it overwritten, I choose Yes, overwrite.
The rest of the files I chose to upload are just ignored (not uploaded).

I was expecting the rest of the files I chose to be uploaded to be overwritten or asked what to do.

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

1.7.2 (and lowers too, was already there in 1.6.x)

### Operating system running OctoPrint

OctoPi 0.16

### Printer model & used firmware incl. version

Prusa MK3S

### Browser and version of browser, operating system running browser

Chrome 95 on MacOS BigSur

### Checklist of files to include below

- [X] Systeminfo Bundle (always include!)
- [ ] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
- [ ] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)
- [ ] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

### Additional information & file uploads

[octoprint-systeminfo-20211104080151.zip](https://github.com/OctoPrint/OctoPrint/files/7472868/octoprint-systeminfo-20211104080151.zip)

jneilliii:
Also reported in #4224, but never received additional information. Thank you for your complete bug report.

foosel:
So, multiple file uploads in one go are currently not supported. So either this is a bug report about the UI allowing this, or/and a feature request to support this.

cp2004:
Uploading multiple files via drag and drop is allowed and works just fine, but only if the files don't collide. If they collide, only the first file is uploaded, which is what this bug report is about.

I just tested this in 1.5.3 to see what the behaviour was before the collision detection stuff, and it was the same, but on collisions they were overwritten on all uploads, as collisions weren't handled previously.

Pressing the 'upload' button only allow you to select a single file, as expected.

foosel:
Aaah, ok, yeah the drag and drop is handled by the file upload widget and I guess that will send off several upload requests. Let me rephrase then, officially multiple uploads aren't supported, and that it currently works by drag and drop should be considered a side effect.

bolsoncerrado:
[...]

PR #4633, opened 2022-08-26
✨ Enable multiple file uploads

cp2004:

This PR has the necessary code changes to allow multiple file uploads to work in the UI with no strange side effects, and making sure that the 'file exists' dialogs work correctly. It fixes issue #4300 and aims to address the comments in #1868 as well.

Some of the stuff it does:

* Allows to select multiple files to upload
* Only refreshes the file list at the end of batch uploading 
* Queues the 'File already exists' dialogs
* Displays the filename in any errors that are shown, so if 1 file out of 5 fails for example, you can see what failed.
* All the files are highlighted yellow, but only scrolling to the position of the last file to be uploaded on stop.
  * This is the only outstanding TODO - I can't get it to scroll to the correct position currently. It will scroll but to completely the wrong place 🙈

I have a small folder of 8 files of varying file sizes that I have been using to test this. Some things that can be done:

* Upload one file (gcode, STL and name collision) (check everything still works as before 🙂)
* Select multiple files from the 'Upload' button
* Drag and drop multiple files
* Uploading multiple STL files (Probably requires a slicer plugin - I used the Slic3r with my PrusaSlicer install)
* Uploading a mix of gcode/STLs
* Uploading a mix of files that collide and files that can just be uploaded
* To demo the error handling, I simulated an error in the API by adding this to the code:
  ```python
  def uploadGcodeFile(target):
      from random import randint
      if randint(1, 5) % 2 == 0:
          abort(500, "simulated error")
  ```

See issues

I have dug quite deep into how the jquery file upload works, the documentation wiki is good
https://github.com/blueimp/jQuery-File-Upload/wiki/Options

#4300 and #1868

#### Screenshots & Videos

Coming soon.... 👀 

#### Further notes

I am not completely sure that the recursive way that file collision dialogs are handled is necessarily the 'best', but it seems to work well. When the dialog closes, the next one is called.

When writing new code, I used modern JavaScript features to make it less verbose. I will have to check on the browser compatibility, but it's passed the ESLint checks.

I'll carry on testing, I would like to test this on my Pi or with a simulated slow network connection to actually watch the progress bar works well. I think it does but it does move quite quickly when just uploading to `localhost` on my SSD!

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/files.js
+++ b/src/octoprint/static/js/app/viewmodels/files.js
@@ -1603,6 +1603,83 @@ $(function () {
             }
         };
 
+        self._uploadExistsQueue = []; // Files will be in this queue if their test fails and something needs to be done
+        self._uploadExistsOpen = false;
+
+        self._processUploadQueue = function () {
+            if (!self._uploadExistsQueue.length > 0 || self._uploadExistsOpen) return;
+
+            const hideAndSubmit = function (data) {
+                self.uploadExistsDialog.modal("hide");
+                self._uploadExistsOpen = false;
+                data.submit();
+                // Recursively move on to process the queue every time a dialog is closed
+                if (self._uploadExistsQueue.length > 0) {
+                    self._processUploadQueue();
+                }
+            };
+
+            // Collect an item from the queue that needs an overwrite dialog
+            const {data, response, path, fileSizeTooBig} =
+                self._uploadExistsQueue.shift();
+            const file = data.files[0];
+
+            const formData = {};
+            if (path !== "") {
+                formData.path = path;
+            }
+
+            // Build and show a dialog
+            self._uploadExistsOpen = true;
+
+            $("h3", self.uploadExistsDialog).text(
+                _.sprintf(gettext("File already exists: %(name)s"), {
+                    name: file.name
+                })
+            );
+            $("p, form", self.uploadExistsDialog).toggle(!fileSizeTooBig);
+            $("span", self.uploadExistsDialog).toggle(fileSizeTooBig);
+            $("input", self.uploadExistsDialog)
+                .val("")
+                .prop("placeholder", response.suggestion);
+            $("a.upload-rename", self.uploadExistsDialog)
+                .toggle(!fileSizeTooBig)
+                .prop("disabled", false)
+                .off("click")
+                .on("click", function () {
+                    var newName = $("input", self.uploadExistsDialog).val();
+                    if (newName === "") newName = response.suggestion;
+
+                    OctoPrint.files.exists("local", path, newName).done(function (r) {
+                        if (r.exists) {
+                            $(".control-group", self.uploadExistsDialog).addClass(
+                                "error"
+                            );
+                            $(".help-block", self.uploadExistsDialog).show();
+                        } else {
+                            $(".control-group", self.uploadExistsDialog).removeClass(
+                                "error"
+                            );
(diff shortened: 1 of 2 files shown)


jneilliii:
Replaces https://plugins.octoprint.org/plugins/MultipleUpload/, should we delist the plugin once merged and released?

cp2004:
I would say yes - tagging @eyal0 to notify, I think there would not be a need to maintain that plugin in the repository anymore.

cp2004:
Note that when testing it with some small files and some large files (on a Pi with an SD card), there is some odd refreshing of the file list - if you upload a 200kb file and a 60mb one on a slow connection for example, OctoPrint can finish analysing the small file, so the list is refreshed mid-upload. These requests would end up being made anyway, so the files just pop up in the list mid way through.

multiple file upload analysis

eyal0:
As an alternative to delisting it, we could also just mark the *maximum* version for it.  That might be appropriate because we might have users that want multiupload but haven't yet upgraded their printers to the latest OctoPrint.  Something like:

```
# Any additional requirements besides OctoPrint should be listed here
plugin_requires = ["OctoPrint<2.12345"]
```

Added to [setup.py](https://github.com/eyal0/OctoPrint-MultipleUpload/blob/master/setup.py).

jneilliii:
Good point @eyal0, can also do that with the plugin listing's md file in compatibility section.

cp2004:
Hmm, adding it to `plugin_requires` would downgrade the OctoPrint version installed to whatever satisfies the constraint if someone tried to install it, which might not be desired behaviour - rather than blocking installation. Putting the compatibility information into the plugin's listing on https://plugins.octoprint.org would mean it would show as 'incompatible' and not be able to be installed through selecting in the plugin manager, and would be a better solution in my opinion.

eyal0:
This file? https://github.com/eyal0/plugins.octoprint.org/blob/gh-pages/_plugins/MultipleUpload.md

What change do I need to make?  Or, one of you can push a PR...?

jneilliii:
I can assist with the modifications when the time comes. It's basically adding PEP440 values to the compatibility.octoprint section of the template.

cp2004:
> * This is the only outstanding TODO - I can't get it to scroll to the correct position currently. It will scroll but to completely the wrong place 🙈
>    * I'll try and sort this out next week.

I have pushed a fix that means that the correct code now runs. It scrolls mostly towards the correct file, although not perfectly in line each time. I am not sure if that is something I impacted however, so maybe I will check that. Otherwise I'm happy this is done.

foosel:
FYI, looking at this is on my TODO list, but I'm currently fully focused on 1.8.3.

src/octoprint/static/js/app/viewmodels/files.js
```diff
@@ -456,21 +456,21 @@ $(function () {
...
+            var focus = [];
             var switchToPath;
 
             if (_.isObject(params)) {
                 focus = params.focus || undefined;
                 switchToPath = params.switchToPath || undefined;
             } else if (arguments.length > 1) {
                 log.warn(
-                    "FilesViewModel.requestData called with old argument list. That is deprecated, please use parameter object instead."
+                    "FilesViewModel.fromResponse called with old argument list. That is deprecated, please use parameter object instead."
```
foosel, line 473: Damn you, copy-🍝! 🙈

foosel:
Now all that remains are merge conflicts 🙈 On it.

Issue #4657, opened 2022-10-02
[Request] Extensible chart markings / labels


rfinnie:
### Is your feature request related to a problem? Please describe.

The new Start/Stop/etc markings on the chart (from #4382) are nice, but they are hardcoded to a handful specific events. It would be awesome if plugins could take advantage of them as well.  I maintain [a plugin](https://plugins.octoprint.org/plugins/bedcooldown/) which begins the bed cooldown shortly before the end of the print, and it would be a good addition to show on the chart when that cooldown begins.

### Describe the solution you'd like

Conceptual invocation:

``` python
from octoprint.events import Events

class BedCooldown(octoprint.plugin.EventHandlerPlugin):
    def on_event(self, event, payload):
        self._logger.info("Cooling down bed")
        self._printer.commands("M140 S0")
        if hasattr(Events, "CHART_MARKING"):
            self._event_bus.fire(
                Events.CHART_MARKING,
                {
                    "label": "Cooldown",
                    "color": "#ffc300",
                    "description": "The bed has begun cooling down",
                },
            )
```

Label and color are self-evident.  Description could be used for hover text.  `hasattr()` for backwards compatibility.

jneilliii:
wonder if you could call it directly via injected properties somehow. this is the relevant function to call: [add_marking](https://github.com/OctoPrint/OctoPrint/blob/f430257d7072a83692fc2392c683ed8c97ae47b6/src/octoprint/printer/standard.py#L716-L722). something like `self._printer.add_marking("Cooldown")` not sure how the js side will handle it properly, if at all...

rfinnie:
Yeah, IMHO `_event_bus` is the "proper" place to put events like this, but extending `add_marking` is probably the practical way to do it.

BTW, as is, the JS *really* does not like arbitrary markings. Most of the UI just disappears as soon as a plugin calls `add_marking`. :laughing: 
![image](https://user-images.githubusercontent.com/704589/193473159-69347932-20b3-43bd-ab04-7e97853358d3.png)

cp2004:
I would agree, extending `add_marking` seems more logical to me. If it got some light documentation in the process as well of course.

rfinnie:
PR opened, I think future discussion can go in there: #4658

foosel:
1.9.0 has been released.

PR #4658, opened 2022-10-03
Add extensible chart markings support

rfinnie:
* [x] (_Not a large audience, but relevant to plugin development_)Your changes are not possible to do through a plugin and relevant
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke
  * [x] (_Already in there_)You have added yourself to the `AUTHORS.md` file :)

Add extensible chart markings support
    
- Replace add_marking with CHART_MARKED event
- Allow for arbitrary markings in the JS UI

Note that this removes `add_marking()`.  In #4657 I said it would be easiest to just extend it, but since plugin maintainers will need to have a check for backwards compatibility anyway, extending it would require them to catch `TypeError` for missing kwargs, which IMHO was too broad.  As it was impossible to use `add_marking()` for plugins in the past, nothing could have used it and it's safe to replace with an event.

~~The only thing I'm not totally happy with is moving style functionality (the colors) into functional code, but I think the additional functionality to plugin writers is worth it.~~

Edit: moved the colors to CSS, with a default color of grey, and removed all color mentions from the code path.  If plugin authors want, they can define custom CSS for `.temperature-mark-type-{type}`

- Verified identical visual functionality for all existing built-in markings (print, done, cancel, pause, resume).
- Simple plugin:

```python
from octoprint.events import Events
import octoprint.plugin
import octoprint.util


class TestMarkings(octoprint.plugin.StartupPlugin):
    def on_after_startup(self):
        if hasattr(Events, "CHART_MARKED"):
            timer = octoprint.util.RepeatedTimer(30, self.example_recurring)
            timer.start()

    def example_recurring(self):
        self._event_bus.fire(
            Events.CHART_MARKED,
            {
                "type": "testmarkings_example",
                "label": "Example",
            },
        )


__plugin_name__ = "TestMarkings"
__plugin_pythoncompat__ = ">=3,<4"


def __plugin_load__():
    global __plugin_implementation__
    __plugin_implementation__ = TestMarkings()
```

Per original feature request:

> I maintain [a plugin](https://plugins.octoprint.org/plugins/bedcooldown/) which begins the bed cooldown shortly before the end of the print, and it would be a good addition to show on the chart when that cooldown begins.

- #4657 (My feature request)
- #4382 (Original PR for chart markings)

image

#### Further notes

I'm not sure how "versionadded" is added for documentation updates; I just put "1.9.0" for the time being.

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/temperature.js
+++ b/src/octoprint/static/js/app/viewmodels/temperature.js
@@ -203,29 +203,6 @@ $(function () {
 
         self.markings = [];
 
-        self.marking_props = {
-            print: {
-                label: gettext("Start"),
-                color: "#218656"
-            },
-            pause: {
-                label: gettext("Pause"),
-                color: "#FDC02F"
-            },
-            resume: {
-                label: gettext("Resume"),
-                color: "#27CAEE"
-            },
-            cancel: {
-                label: gettext("Cancel"),
-                color: "#DA3749"
-            },
-            done: {
-                label: gettext("Done"),
-                color: "#1B72F9"
-            }
-        };
-
         self.showStateMarks = ko.observable(
             loadFromLocalStorage("temperatureGraph.showStateMarks", true)
         );
@@ -485,15 +462,15 @@ $(function () {
                     y: self.plot.getAxes().yaxis.max
                 });
 
+                var markLabel = mark.type;
+                if (mark.label) {
+                    markLabel = gettext(mark.label);
+                }
                 if (o.left > yAxisLabelWidth) {
                     var label = $("<div></div>");
-                    label.html(self.marking_props[mark.type].label);
-                    var css = {
-                        backgroundColor: self.marking_props[mark.type].color
-                    };
-
-                    label.css(css);
+                    label.html(markLabel);
                     label.addClass("temperature-mark-label");
+                    label.addClass("temperature-mark-type-" + mark.type);
 
                     graph.append(label);
 
@@ -520,7 +497,7 @@ $(function () {
                 }
 
                 return {
-                    color: self.marking_props[mark.type].color,
+                    color: label.css("background-color"),
                     lineWidth: lineWidth,
                     xaxis: {from: time, to: time}
                 };
@@ -528,6 +505,17 @@ $(function () {
 
             return marks;
         };
+
+        // Dummy translation requests for dynamic strings supplied by the backend
+        // noinspection BadExpressionStatementJS
+        [
+            // mark labels
+            gettext("Start"),
+            gettext("Done"),
+            gettext("Cancel"),
+            gettext("Pause"),
+            gettext("Resume")
+        ];
 
         self._initializePlot = function (force, plotInfo) {
             var graph = $("#temperature-graph");
--- a/docs/events/index.rst
+++ b/docs/events/index.rst
@@ -605,6 +605,17 @@ GcodeScript${ScriptName}Finished
 
    .. versionadded:: 1.6.0
 
+ChartMarked
+   A time-based marking has been made on the UI's temperature chart.
+
+   Payload:
+
+     * ``type``: the marking's ID. Built-in types are ``print``, ``done``, ``cancel``, ``pause``, and ``resume``. Plugins may set arbitrary types
+     * ``label``: the human-readable short label of the marking
+     * ``time``: the epoch time of marking
+
+   .. versionadded:: 1.9.0
+
 .. _sec-events-available_events-gcode_processing:
 
 GCODE processing
--- a/src/octoprint/events.py
+++ b/src/octoprint/events.py
@@ -81,6 +81,7 @@ class Events:
     PRINT_PAUSED = "PrintPaused"
     PRINT_RESUMED = "PrintResumed"
     ERROR = "Error"
+    CHART_MARKED = "ChartMarked"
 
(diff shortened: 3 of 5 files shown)


rfinnie:
Sorry, rebased from devel to maintenance.  I don't know why I was thinking of this as being completely new functionality :)

cp2004:
One issue with switching the adding of the labels from the JS viewmodel to the server-side events system, is that the `gettext` translations won't work anymore. Because there is no language context there, they will remain as their English defaults. Currently, they are translated by the frontend. `gettext` only functions correctly when called as part of a flask request/with flask context.

rfinnie:
Hmm, is that correct? I've switched over the language to German and the server-side `flask_babel.gettext()` appears to correctly translate the markings:

image

cp2004:
I'll have to double check then - perhaps it varies based on whether you set the server language vs. user language vs. requesting in the URL with `?i18n=` (may not be exactly that - I can't remember...!).

rfinnie:
Cool, thanks, it's entirely possible there's an edge case I haven't encountered yet (at least i18n-wise).

cp2004:
So I've just had a look at this for a test, noticed a couple of things:

* The colours of the lines don't seem to be working correctly. In your screenshot in https://github.com/OctoPrint/OctoPrint/pull/4658#issuecomment-1267289386 they are grey as well - the labels are working OK, but the lines aren't.
* My comment about the translations was mostly correct. Translating server side, outside the request context means that it only respects the default language of the server, not an individual user's language or with the request values `l10n=<language code>`. Now this is not the only place where this is the case in OctoPrint, there are certain things that are already not respecting the translations the same way (such as the permissions roles). I'll leave it up to @foosel to decide which is best in this case.

rfinnie:
Thanks for looking into this.

Re: language, agreed, we should defer to @foosel on it.  I think having the labels server-side is the right way to do it since the mechanism for core and plugins becomes the same, but I don't know how critical HTTP Language / query option overrides should be considered vs OctoPrint-wide language.

Re: lines, turns out that's because there's no way to apply a style to canvas stuff, and while the flag is an HTML element, turns out the line is part of the canvas.  I hadn't even noticed it until you pointed it out, and while it is a visual regression, IMHO it's acceptable given the functionality trade-off.

jneilliii:
>Re: lines, turns out that's because there's no way to apply a style to canvas stuff, and while the flag is an HTML element, turns out the line is part of the canvas. I hadn't even noticed it until you pointed it out, and while it is a visual regression, IMHO it's acceptable given the functionality trade-off.

How did the previous implementation do it? Seems it's able to color the lines fine, so maybe just needs a little more tweaking.

image

src/octoprint/static/js/app/viewmodels/temperature.js
```diff
@@ -520,7 +497,7 @@ $(function () {
                 }
 
                 return {
-                    color: self.marking_props[mark.type].color,
+                    color: "#aeb6bf",
```
jneilliii, line 500: I think this is where you need adjust the code to have the line the same color? You're forcing the gray.

rfinnie: tl;dr this ended with me going "oh right, you can extract CSS properties via JS!" :)  Fixed to match the corresponding CSS.

image

foosel:
> Re: language, agreed, we should defer to @foosel on it. I think having the labels server-side is the right way to do it since the mechanism for core and plugins becomes the same, but I don't know how critical HTTP Language / query option overrides should be considered vs OctoPrint-wide language.

I was absolutely sure I had already replied to this, but maybe the response just sat unsent in some tab and then went the way of the Dodo on a browser restart.

Anyhow... the way I would go about this is using the same approach I use for the printer state translations. Dummy translations:

https://github.com/OctoPrint/OctoPrint/blob/a68916f05a251052e139486df9932ecb246ae591/src/octoprint/static/js/app/main.js#L257-L281

So, have that somewhere on the frontend, send language agnostic (aka english) strings from the backend (better for an API anyhow) and then use gettext on the dynamic string on the frontend, with a working language context.

rfinnie:
> So, have that somewhere on the frontend, send language agnostic (aka english) strings from the backend (better for an API anyhow) and then use gettext on the dynamic string on the frontend, with a working language context.

Thanks @foosel, PR updated, and confirmed mark labels work with either server-wide language or `?l10n=de`.

foosel (approved):

foosel:
Looking good, merged, thank you very much!

foosel:
Currently "something else" in this case is "Gina will do it when she looks into merging and releasing stuff" 😅

cp2004:
As long as you either update the less, or update the less *and* css, it's alright. Just not the css on it's own.

rfinnie:
@foosel If you are interested in highlighting this functionality with the 1.9.0 release, I just wanted to let you know I've released OctoPrint-BedCooldown 0.6.0 today, which has support for it, and (IMHO) best practices:

- Conditional support (does not break pre-1.9.0 installations)
- Stylized (sets its own marking color)
- Language support (no actual translations, but the gettext framework is in place)

image

foosel:
@rfinnie That's awesome! Would it be cool to use this screenshot in the release announcement?

rfinnie:
@foosel Yep, no problem to include that screenshot.

PR #4679, opened 2022-11-03
:pencil2: virtual: Fix a log typo

cp2004:
:)

I'm going to assume there was no reason for this, so hopefully it is fine...

I just noticed the inconsistency in the logging plugin's settings


foosel:
If there was a reason, I can't remember it, so thanks 😄

Issue #4693, opened 2022-12-15
[Request] Please allow to rename jobs (gcode files) and "star" those to keep.


AndKe:
### Is your feature request related to a problem? Please describe.

![Screenshot from 2022-12-15 14-34-35](https://user-images.githubusercontent.com/1468842/207872652-6584d0e8-14cd-45ed-93dc-a2b95e69aa19.png)

### Describe the solution you'd like

When working with FreeCad and/or other tools, exported items may get different names.
Then - the objects might be duplicated in the slicer - so even if a job is automatically named: "Mount for xyz"  - it can be desired to save it as "3x Mount for xyz"   or something that describes the material and surface: "Mount for xyz - PLA on PEI"

Please allow renaming files. - good jobs may be worth saving.

Also: the files have properties like upload date, printed date and so on. It would be very nice to have a simple bool attribute "starred"  - then the file list could be sorted by "starred" (worth keeping jobs)   - with good names.

jneilliii:
Try pressing that little cut/scissor button and putting in a new name?

foosel:
As stated by @jneilliii, renaming (and moving) is already possible via the scissors icon. Which makes this a request to be able to star dear-to-heart print jobs. And that makes it a duplicate of #3433

PR #4694, opened 2022-12-16
:lipstick: File/Folder move UI improvements

cp2004:

Noticed a couple of small QoL things that I thought I would address with the file/folder moving & renaming.

1. Missing styling on the name box
  image

  *fixed*

1. Not obvious you can rename files until after you click the 'scissors' icon. The icon & tooltip don't imply this (see #4693) to the average user perhaps? 
   *Made it more obvious.*

1. No distinction between moving files or folders. The title says 'Move file or folder', the name just says 'filename' (but applies to folder names as well) & text at the bottom says 'file/folder will be moved to'. Just got on my nerves when I looked at it 😆 
   *Made it show the correct name*

1. Text at the bottom 'File/Folder will be moved to' is visible even when there is no action to be performed.
  *Made it disappear when nothing will happen*

1. Button says 'Move', while file moving & renaming are the same in the code, I thought it made more sense for this to be 'confirm'. Up for debate on that one I guess, just sounded better to me *personally*.
  *Made it confirm*

Locally on my development install by moving a folder or file.

Inspired by #4693 being opened


jneilliii:
confirm makes sense, but you could also make the text change from move to rename when the source folder is the same as destination...

foosel:
> confirm makes sense, but you could also make the text change from move to rename when the source folder is the same as destination...

I actually really like that idea. "Confirm" while valid is less explicit, and a dynamic button title would make things crystal clear here. @cp2004, want to look into that?

cp2004:
> @cp2004, want to look into that?

Will do as soon as I get a chance.

cp2004:
Done 🙂

PR #4700, opened 2022-12-21
helpers for action_command_notification built-in

synman:
<#### What does this PR do and why is it necessary?
adds helpers to the action_command_notification plugin so that other plugins can interact with it
-add notification
-get notifications
-clear notifications 

For my specific use case, the helper I really need is the add notification one since the plugin is currently locked down to only allow for receipt of `//action:notification` messages over the serial link.  Many machines do not use this notation and the prominent Notifications Sidebar is often times empty when it could be displaying useful information.

```
        # get a reference to the action_command_notification 
        # plugin's add_notification helper 
        helpers = self._plugin_manager.get_helpers("action_command_notification")
        if helpers and "add_notification" in helpers:
            self._add_notification = helpers["add_notification"]
            self._add_notification("I am a notification")
```
I also performed regression by simulating an `//action:notification` receive by hooking into `hook_gcode_received` and returning "//action:notification I am a test message".  This demonstrated that the core functionality of the plugin remains intact.

I have a video that shows my use case as I can understand on the surface it may seem odd why one may want to override this plugin:

https://www.youtube.com/watch?v=KIfJP2_FG3k

Prior PR that went up in flames:  https://github.com/OctoPrint/OctoPrint/pull/4697

See the video linked above

### DIFF
--- a/src/octoprint/plugins/action_command_notification/__init__.py
+++ b/src/octoprint/plugins/action_command_notification/__init__.py
@@ -55,7 +55,7 @@ class ActionCommandNotificationPlugin(
 
     def on_event(self, event, payload):
         if event == Events.DISCONNECTED:
-            self._clear_notifications()
+            self.clear_notifications()
 
     # ~ SettingsPlugin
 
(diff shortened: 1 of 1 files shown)


synman:
I've been thinking more about this, and potentially at the cost of another PR, I was thinking about how terminal filters are implemented and if it is feasible to add additional criteria via regex to the actual action_command_notification plugin.

Since this plugin already has a gcode_received hook, it would keep it all self contained.

The regex marches could be exposed via the plug-in settings panel and have a default entry that matches //action: notification.

synman:
I'm closing this one too.  I ended up solving my specific issue by use of some viewmodel wizardry.

PR #4704, opened 2023-01-03
⚡ Speed up file list by moving full_extension_tree call out of the hot path

the-ress:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

Moves `full_extension_tree` call outside the file listing loop so the extension tree is not computed for each file.

I tested the changes by listing files (`/api/files/local`) on an instance with virtual printer and also on an instance I use for my printer.

I was investigating why it takes so long to get a list of files, and this was one of the things I noticed that could be improved. I currently have ~500 files in my uploads folder and `full_extension_tree` would take almost 2 seconds on RPi Zero 2. (One of the extensions I use has extension tree hook, but doesn't do anything intensive there, just returns a value.)

While I could remove some of the gcode files (again), I figured I'd rather try to speed up the code.

### DIFF
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -180,19 +180,19 @@ def get_content_type_mapping_for_extension(extension, subtree=None):
     return None
 
 
-def valid_extension(extension, type=None):
+def valid_extension(extension, type=None, tree=None):
     if not type:
-        return extension in get_all_extensions()
+        return extension in get_all_extensions(subtree=tree)
     else:
-        extensions = get_extensions(type)
+        extensions = get_extensions(type, subtree=tree)
         if extensions:
             return extension in extensions
 
 
-def valid_file_type(filename, type=None):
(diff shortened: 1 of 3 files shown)


cp2004:
> While I could remove some of the gcode files (again), I figured I'd rather try to speed up the code.

I definitely prefer this solution 🙂

the-ress:
> I definitely prefer this solution 🙂

I sometimes want to re-print an older gcode and I hate having to ssh over to the RPi and look for the file in `folder-where-I-moved-old-gcodes` (as opposed to just using octoprint UI).

foosel:
Really great find! Also makes me think we could be even more aggressive and cache this based on the list of registered slicer plugins and extension tree hooks. As long as that list doesn't change, the tree never needs to be recomputed. We could also just declare SlicerPlugins and the hook as restart needing, in which case we can just compute this thing once on startup and never again.

Does this sound reasonable or am I missing something here?

Issue #4681, opened 2022-11-07
[Request] "File Already Exists" dialogue - Add information on old/new files


Clumpton:
### Is your feature request related to a problem? Please describe.

When you upload a gcode file to Octoprint and a file of the same name is already present a "File Already Exists" dialogue is show offering  to Cancel, Rename or Overwrite it. (see screenshot)
![file exists dialogue](https://user-images.githubusercontent.com/59048696/200374800-88f7138f-aab9-4116-b7d7-46487cfdef38.jpg)
There is no indication of the age, size or upload date/time of the pre-existing file, this could lead to confusion or loss of a precious gcode file.

### Describe the solution you'd like

The "File Already Exists" dialogue would show additional file information to help decide whether it is safe to overwrite.
For example:
**Existing file foobar.gcode:**
- Size 250 kb
- Date 2021-12-25 15:35:20
- Uploaded 3 weeks ago

**New file foobar.gcode:**
- Size 721 kb
- Date 2022-1107 18:30

### Describe alternatives you've considered

I've considered being more careful and remembering old files, but I'm a wrinkly and my memory fails me :-)

max246:
I have made a PR for this https://github.com/OctoPrint/OctoPrint/pull/4720

Clumpton:
> I have made a PR for this #4720

Thanks for that!

foosel:
Implemented by #4721, will be in 1.10.0

Clumpton:
Thanks for that 

⁣Get TypeApp for Android ​

foosel:
1.10.0 has been released.

Clumpton:
Nice implementation, thanks again.

PR #4721, opened 2023-01-31
Add information regarding existing and new file before override

max246:
* [ x] Your changes are not possible to do through a plugin and relevant
  * [x ] If your changes are large or otherwise disruptive: You have
    Brainstorming ticket
  * [ x] Your PR targets OctoPrint's `devel` branch if it's a completely
  * [ x] Your PR was opened from a custom branch on your repository
  * [ x] Your PR only contains relevant changes: no unrelated files,
    if necessary!
  * [ x] Your changes follow the existing coding style
  * [ x] If your changes include style sheets: You have modified the
    with `lessc`)
  * [ x] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [ x] You have run the existing unit tests against your changes and
    nothing broke
  * [ x] You have added yourself to the `AUTHORS.md` file :)

A request regarding to show the difference between the two file before hit override

#4681

Selection_011

### DIFF
--- a/src/octoprint/server/api/files.py
+++ b/src/octoprint/server/api/files.py
@@ -192,6 +192,13 @@ def runFilesTest():
 
         exists = fileManager.file_exists(storage, sanitized)
         if exists:
+            exists_file = [gcode_file for gcode_file in _getFileList(storage) if
+                           gcode_file['name'] == filename]
+            if len(exists_file) > 0:
+                exists_size = exists_file[0]['size']
+                exists_date = exists_file[0]['date']
+            else:
+                abort(400, descrption='Not exists file found to retrive information')
             suggestion = filename
             name, ext = os.path.splitext(filename)
             counter = 0
@@ -205,7 +212,7 @@ def runFilesTest():
             ):
                 counter += 1
                 suggestion = f"{name}_{counter}{ext}"
-            return jsonify(exists=True, suggestion=suggestion)
+            return jsonify(exists=True, suggestion=suggestion, size=exists_size, date=exists_date)
         else:
             return jsonify(exists=False)
 
--- a/src/octoprint/static/js/app/viewmodels/files.js
+++ b/src/octoprint/static/js/app/viewmodels/files.js
@@ -1693,6 +1693,13 @@ $(function () {
                     name: file.name
                 })
             );
+            $("span.existing_size", self.uploadExistsDialog).text(formatSize(response.size));
+            var date_exists = new Date(response.date * 1000);
+            $("span.existing_date", self.uploadExistsDialog).text(date_exists.toLocaleString());
+            var date = new Date(file.lastModified);
+
+            $("span.new_size", self.uploadExistsDialog).text(formatSize(file.size));
+            $("span.new_date", self.uploadExistsDialog).text(date.toLocaleString());
             $("p, form", self.uploadExistsDialog).toggle(!fileSizeTooBig);
             $("span", self.uploadExistsDialog).toggle(fileSizeTooBig);
             $("input", self.uploadExistsDialog)
--- a/src/octoprint/templates/dialogs/files.jinja2
+++ b/src/octoprint/templates/dialogs/files.jinja2
@@ -70,6 +70,19 @@
                 </div>
             </div>
         </form>
+        <div>
+            <b>Existing file</b>
+            <ul>
+            <li>Size: <span class="existing_size"></span></li>
+            <li>Date: <span class="existing_date"></span></li>
+            </ul>
+
+            <b>New file</b>
+            <ul>
+            <li>Size: <span class="new_size"></span></li>
+            <li>Date: <span class="new_date"></span></li>
+            </ul>
+        </div>
         <span hidden>
             {{ _('There is not enough free disk space to keep the existing file and upload this one.') }}
         </span>


src/octoprint/templates/dialogs/files.jinja2
```diff
@@ -70,6 +70,19 @@
                 </div>
             </div>
         </form>
+        <div>
+            <b>Existing file</b>
+            <ul>
+            <li>Size: <span class="existing_size"></span></li>
```
cp2004, line 76: Do the classes actually mean anything? As far as I can see they are used as a way to reference the element in the jquery selector. Perhaps it would make more sense to use actual IDs?

cp2004 (changes requested):

src/octoprint/templates/dialogs/files.jinja2
max246: cool, will update that later today

foosel: Tbh, given how the query is done, I'd prefer we leave it a class so as to not run into namespace issues.

`.existing_size` within `self.uploadExistsDialog` (so almost like it is now, I'd simply get rid of the `span` specification, could be another element at some point) is perfect for the use here.

If we use an id here, we'll need to look into name spacing, `existing_size` is too unspecific for a global site-wide id.

foosel: No, the way it is now it's better, as it separates the value from the label (which btw also still needs to be made translatable, e.g. `{{ _('Size:') }} `).

max246: done

src/octoprint/server/api/files.py
```diff
@@ -192,6 +192,16 @@ def sanitize(storage, path, filename):
...
+            exists_file = [
+                gcode_file
+                for gcode_file in _getFileList(storage)
+                if gcode_file["name"] == filename
+            ]
+            if len(exists_file) > 0:
+                exists_size = exists_file[0]["size"]
+                exists_date = exists_file[0]["date"]
+            else:
+                abort(400, description="Not exists file found to retrieve information")
```
jneilliii, line 204: Wording is a bit strange on this description. Maybe something more like `File not found to retrieve additional information`. It would probably be a rare race condition of the file being deleted and getting into this else block, but just in case.

jneilliii, line 204: Wording is a bit strange on this description. Maybe something more like `File not found to retrieve additional information`. It would probably be a rare race condition of the file being deleted and getting into this else block, but just in case.

jneilliii (commented):
better wording?

src/octoprint/server/api/files.py
max246: Yeah I just covered the basic of failure :) otherwise it will never throw an error if this case happens.

Updated the message.

src/octoprint/static/js/app/viewmodels/files.js
```diff
@@ -1693,6 +1693,17 @@ $(function () {
...
+                formatSize(response.size)
+            );
+            var date_exists = new Date(response.date * 1000);
+            $("span.existing_date", self.uploadExistsDialog).text(
+                date_exists.toLocaleString()
+            );
+            var date = new Date(file.lastModified);
+
+            $("span.new_size", self.uploadExistsDialog).text(formatSize(file.size));
+            $("span.new_date", self.uploadExistsDialog).text(date.toLocaleString());
```
foosel, line 1706:
```suggestion
            $("span.existing_size", self.uploadExistsDialog).text(
                response.size ? formatSize(response.size) : "-"
            );
            $("span.existing_date", self.uploadExistsDialog).text(
                response.date ? (new Date(response.date * 1000)).toLocaleString() : "?"
            );
            $("span.new_size", self.uploadExistsDialog).text(formatSize(file.size));
            $("span.new_date", self.uploadExistsDialog).text(
                (new Date(file.lastModified).toLocaleString()
            );
```

src/octoprint/templates/dialogs/files.jinja2
```diff
@@ -70,6 +70,19 @@
                 </div>
             </div>
         </form>
+        <div>
+            <b>{{ _('Existing file') }}</b>
```
foosel, line 74: `strong`, not `b`

src/octoprint/templates/dialogs/files.jinja2
```diff
@@ -70,6 +70,19 @@
...
+            <ul>
+            <li>{{ _('Size') }}: <span class="existing_size"></span></li>
+            <li>{{ _('Date') }}: <span class="existing_date"></span></li>
+            </ul>
+
+            <b>{{ _('New file') }}</b>
```
foosel, line 80: `strong`, not `b`

src/octoprint/templates/dialogs/files.jinja2
```diff
@@ -70,6 +70,19 @@
                 </div>
             </div>
         </form>
+        <div>
+            <b>{{ _('Existing file') }}</b>
+            <ul>
+            <li>{{ _('Size') }}: <span class="existing_size"></span></li>
+            <li>{{ _('Date') }}: <span class="existing_date"></span></li>
```
foosel, line 77: Indentation

src/octoprint/templates/dialogs/files.jinja2
```diff
@@ -70,6 +70,19 @@
...
+            </ul>
+
+            <b>{{ _('New file') }}</b>
+            <ul>
+            <li>{{ _('Size') }}: <span class="new_size"></span></li>
+            <li>{{ _('Date') }}: <span class="new_date"></span></li>
```
foosel, line 83: Indentation

foosel (changes requested):
Have some minor refactorings (haven't gone through precommit yet!). For the adjustments of the `exists` command on the API, I would actually go for something like this:

``` python
        exists = _getFileDetails(storage, filename) # or sanitized? unsure
        if exists:
            suggestion = filename
            name, ext = os.path.splitext(filename)
            counter = 0
            while fileManager.file_exists(
                storage,
                fileManager.join_path(
                    storage,
                    sanitized_path,
                    fileManager.sanitize_name(storage, suggestion),
                ),
            ):
                counter += 1
                suggestion = f"{name}_{counter}{ext}"
            return jsonify(
                exists=True, 
                suggestion=suggestion,
                size=exists.get("size"),
                date=exists.get("date"))
        else:
            return jsonify(exists=False)
```

Saves us some CPU cycles and is also cleaner as in that we do not try to get the same information twice from different locations. Rest see the comments with suggestions.

Also, in case of SD files, we might not have a size or a date, so we need to compensate for that. Adjusted accordingly.

max246:
I think this can be merged, no?

foosel (approved):

foosel:
Yes, I agree. Alas, I never got a request for another review from you, so I never reviewed again and I fear that now means it'll have to wait until 1.10.0 for release, 1.9.0 has already entered the RC phase and thus the only changes it will see going forward are fixes of regressions.

@cp2004 I saw you previously requested changes as well, as far as I can see those have been resolved however. Go for merge from you too?

cp2004 (approved):
Not tested it, but reading the code looks ok. I was going to ask about documentation updates, but I don't think I can see the `/files/test` endpoint documented anywhere anyway?

foosel:
> I don't think I can see the `/files/test` endpoint documented anywhere anyway?

Yeah, that's still missing, but that's my fault.

max246:
> Yes, I agree. Alas, I never got a request for another review from you, so I never reviewed again and I fear that now means it'll have to wait until 1.10.0 for release, 1.9.0 has already entered the RC phase and thus the only changes it will see going forward are fixes of regressions.
> 
> @cp2004 I saw you previously requested changes as well, as far as I can see those have been resolved however. Go for merge from you too?

its ok dont worry. as long it has been merged I am happy :D

Issue #4740, opened 2023-02-24
CSRF fails to function on non-standard port (non 443)


Fmstrat:
### The problem

This problem is extensively detailed (including links to code lines in posts 11-12) here: https://community.octoprint.org/t/csrf-reverse-proxy-configuration-issue-or-bug/50066/12

The overall problem is that OctoPrint swaps a custom HTTPS port for `443` due to the way it interprets the scheme. So if you run a reverse proxy at `https://octoprint.domain.com:12345` CSRF fails because it suffixes the cookie with `_P443` instead of `_P12345`.

Reposting code locations here for ease of access:

> 
> Hrm, perhaps this is a bug:
> 
> https://github.com/OctoPrint/OctoPrint/blob/a68916f05a251052e139486df9932ecb246ae591/src/octoprint/static/js/app/client/base.js#L133
> 
> This is a theory, but I believe in `getParsedBaseUrl` it's filtering out the custom port, and thus `port != url.port`, so since `url.protocol === "https:"` the port for the cookie suffix is getting set to 443.
> 
> Is `baseurl` still used? Where does this get set:
> 
> https://github.com/OctoPrint/OctoPrint/blob/a68916f05a251052e139486df9932ecb246ae591/src/octoprint/static/js/app/client/base.js#L111
> 
> That would cause the port to not be relayed down the line for the cookie suffix.

> 90% sure this is it now. `baseurl` is not in existance, so `getParsedBaseUrl` is returning back `""` and since it's an `https` scheme, 443 is assigned. 
> 
> `getCookieSuffix` is used in `getCookieName` is used in `getCookie` is used in `getRequestHeaders` which would make `csrfToken == null` here:
> 
> https://github.com/OctoPrint/OctoPrint/blob/a68916f05a251052e139486df9932ecb246ae591/src/octoprint/static/js/app/client/base.js#L182
> 

**EDIT**: I should also mention this can take place with a `baseurl`, as it looks like the `catch` in `getParsedBaseUrl` also disregards `parsed.port`.

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

1.8.6

### Operating system running OctoPrint

OctoPi

### Printer model & used firmware incl. version

n/a

### Browser and version of browser, operating system running browser

Firefox, Chrome

### Checklist of files to include below

- [X] Systeminfo Bundle (always include!)
- [X] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
[...]

PR #4741, opened 2023-02-24
Remove custom port from CSRF cookie suffix

Fmstrat:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

Addresses: https://github.com/OctoPrint/OctoPrint/issues/4740 by utilizing 443 or 80 as the port for the CSRF cookie name even if a custom port is being used.

The best method is to set up an NGINX proxy outside of OctoPi running on `https://octoprint.domain.com:12345` and confirm that `https://octoprint.domain.com:12345/reverse_proxy_test/` is all green. You will see 443 as the port (to align to the server port from OctoPi). Without this fix, logins will fail with a 400, with it, they will succeed regardless of if you are running a standard port, non-standard port, going through `haproxy`, or direct.

https://github.com/OctoPrint/OctoPrint/issues/4740 has a link to the forum post with a full description and debugging of this issue.

https://github.com/OctoPrint/OctoPrint/issues/4740 

N/A

#### Further notes
N/A

### DIFF
--- a/src/octoprint/static/js/app/client/base.js
+++ b/src/octoprint/static/js/app/client/base.js
@@ -119,7 +119,7 @@
             if (!path || path[0] !== "/") {
                 path = "/" + (path ? path : "");
             }
-            var url = new URL(parsed.protocol + "//" + parsed.host + path);
+            var url = new URL(parsed.protocol + "//" + parsed.hostname + path);
         }
 
         return url;


cp2004:
Should the fix not be to go the other way around - rather than get the port wrong in both places, OctoPrint should get the correct port? It should match what is actually happening, which is you using the UI from port 12345, not 443.

Fmstrat:
> Should the fix not be to go the other way around - rather than get the port wrong in both places, OctoPrint should get the correct port? It should match what is actually happening, which is you using the UI from port 12345, not 443.

Yes.. but no?

The current CSRF methodology of OctoPrint is really setting the cookie name via schema, it's just aliasing the port. For instance, `_P80` could be `_SHTTP` and `_P443` could be `_SHTTPS`. While it might "look" odd to have `443`, that's the port the server believes it is listening on, since there is no `baseurl` set. I.E. the current cookie naming convention is somewhat arbitrary.

This could be handled with an overhaul to force setting a `baseurl` with custom port when using a proxy (including `haproxy` in OctoPi), and then updating OctoPrint to always leverage a "true port" for the cookie name instead of P80/P443. But since the cookie name is arbitrary and set via schema (just with the port as the name), it doesn't add any additional security.

So: Yes, if we want the reverse proxy page to "look better." But no if we want the same functionality as today, ease of setup via not requiring users with custom ports to set custom configurations in `config.yaml`, and a simpler fix with little impact to other pieces of code, all while maintaining the same security posture. Given the benefits of the "no", I personally lean that way.

cp2004:
> The current CSRF methodology of OctoPrint is really setting the cookie name via schema, it's just aliasing the port. For instance, _P80 could be _SHTTP and _P443 could be _SHTTPS.

While I agree (to the best of my knowledge) the cookie suffix is a label (and it could be whatever we decide), it's not right to say it only takes either 80 or 443. When I access OctoPrint directly (no proxy) through `localhost:5000` I get the correct suffix:
image

And if I use haproxy to send it through http port 4000, I still get a correct suffix:

image

So it does listen to the port number, not just using the scheme, and when properly setup they match.

In your forum post, it looks to me like the client is generating the cookie suffix correctly as `_P12345` (in the screenshot), but this is not matching the server. The client is correct, the server is wrong, so rather than make the client *also wrong* (but the same wrong as the server) the server should be made right to tell the truth about what port you access OctoPrint on. That could either be a configuration issue or a bug, I haven't yet tried that but it is up next (likely at the weekend)

cp2004:
OK, it didn't take that long to test in the end...

image

I setup haproxy with a self-signed cert, on port 3000 for https, and everything works as expected. So I would be inclined to say there is still something wrong with your nginx configuration, and so I'll hop back to the community forums to have a poke at what might be the issue there.

For reference, the haproxy config I used:

<details>
<summary>haproxy.cfg</summary>

```
global
        maxconn 4096
        user haproxy
        group haproxy
        log /var/log/haproxy.log local1 debug
        tune.ssl.default-dh-param 2048

defaults
        log     global
        mode    http
        compression algo gzip
        option  httplog
        option  dontlognull
        retries 3
        option redispatch
        option http-server-close
        option forwardfor
        maxconn 2000
        timeout connect 5s
        timeout client  15m
        timeout server  15m

frontend public
        bind :::4000 v4v6
        bind :::3000 v4v6 ssl crt /etc/ssl/snakeoil.pem
        option forwardfor except 127.0.0.1
        default_backend octoprint

backend octoprint
        acl needs_scheme req.hdr_cnt(X-Scheme) eq 0

        http-request replace-path ^([^\ :]*)\ /(.*) \1\ /\2
        http-request add-header X-Scheme https if needs_scheme { ssl_fc }
        http-request add-header X-Scheme http if needs_scheme !{ ssl_fc }
        option forwardfor
        server octoprint1 127.0.0.1:5000
```

Fmstrat:
Interesting, I wonder why you're getting a different result. I guess what I'm saying is that the function I changed isn't leveraged by any part of the system other than the CSRF cookie name. The only reason I can think of to even have pre-defined cookie names like this would be to separate HTTP from HTTPS, but if that's the case, there's no reason to use a port in the cookie name at all. It could just be changed to "_SHTTP" and "_SHTTPS" to remove the complexity altogether. The proxy port in headers and/or cookies have no impact on the proxy itself (including file downloads) since the only place it's used is with CSRF (at least that I see calling that function).

In any event, if there's a cleaner way to do this, it works for me. I'm just not finding it with nginx.

Fmstrat:
@cp2004 Trying to figure out what's different about your setup and mine, and I noticed `host_to_server_and_port` does have separate code for determining port when IPV6 vs IPV4. I don't see any "issue" per-se, but if you are willing to take a look, you know this stuff better than me.

cp2004:
That's great that you've managed to sort it out - and we've learnt something about how nginx works with https ports then I guess. I'll have to play around with nginx closer (I tend to always use haproxy as that's what's included with OctoPi) to see if it has any other slight quirks.

PR #4758, opened 2023-03-11
Added D3 to default list of autoUppercaseBlacklist

alexiri:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

This PR adds the gcode command D3 to the default list of gcodes excluded from uppercasing. [D3](https://reprap.org/wiki/G-code#D3:_Read.2FWrite_EEPROM) is a Prusa firmware debugging gcode that allows you to read and write the EEPROM, and the data to be written is encoded as case-sensitive alphanumerics. Uppercasing the data would alter what is being written to the EEPROM, with potentially serious consequences.

This is a Prusa firmware specific gcode (for now), but this is a pretty low-risk change and it could save other Prusa users quite a bit of trouble in the future.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -176,6 +176,7 @@ date of first contribution):
   * [Moiz Rasheed](https://github.com/msrasheed)
   * [Scott Martin](https://github.com/smartin015)
   * [Shyam Sunder](https://github.com/sgsunder)
+  * [Alex Iribarren](https://github.com/alexiri)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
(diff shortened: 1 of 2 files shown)


alexiri:
Digging deeper, I realized that D3 commands wouldn't be uppercased anyway because they [don't match this regex](https://github.com/OctoPrint/OctoPrint/blob/48b13d550cd247776812855fed74a34ff375f95e/src/octoprint/static/js/app/viewmodels/terminal.js#L417), but maybe it's not a bad idea to add this as a default in case it changes at some point.

foosel:
As you already realized, `D3` is not a valid GCODE command but a very much vendor specific command. Hence, OctoPrint won't touch it to begin with, and that won't change either. 

I do my best to leave vendor specific adjustments to plugins these days, to keep the core clean from single vendor workarounds and other kind of fiddling, before it turns completely unmanageable due to vendors refusing to stick to standards. 

For obvious reasons I'd rather not change that, and therefore won't merge this. Sorry, but thank you for your understanding.

PR #4759, opened 2023-03-11
Use webcam container for intersection instead of controls

crysxd:
This is a minor change to allow @LazeMSS and others to move the webcam surface around. We used `#control` to determine if the webcam and if yes which webcam plugin is visible to the user. This causes issues if the webcam is not embedded in `#control`. This PR changes the intersection code to use `#webcam_plugins_container` which is the outer most div containing webcam code.

Functionally, this does not make any difference in vanilla OctoPrint

- Switch away from control tab -> Webcam unloaded after 5s
- Switch browser tab -> Webcam unloaded after 5s

https://discord.com/channels/704958479194128507/708230829050036236/1083833450076848229

https://github.com/LazeMSS/OctoPrint-UICustomizer/issues/281

~

#### Further notes
~

### DIFF
--- a/src/octoprint/plugins/classicwebcam/static/js/classicwebcam.js
+++ b/src/octoprint/plugins/classicwebcam/static/js/classicwebcam.js
@@ -117,18 +117,11 @@ $(function () {
         };
 
         self._enableWebcam = function () {
-            if (
-                OctoPrint.coreui.selectedTab != "#control" ||
-                !OctoPrint.coreui.browserTabVisible
-            ) {
-                return;
-            }
-
             if (self.webcamDisableTimeout != undefined) {
                 clearTimeout(self.webcamDisableTimeout);
             }
 
-            // IF disabled then we dont need to do anything
+            // If disabled then we dont need to do anything
             if (self.settings.webcamEnabled() == false) {
                 console.log("Webcam not enabled");
                 return;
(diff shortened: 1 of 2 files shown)


cp2004:
This makes sense to me. I would be hesitant to support too much hacking of the UI, as it just makes it increasingly difficult to change anything, but in this case there's nothing extra in OctoPrint's code so I don't think there should be an issue.

crysxd:
As pointed out by @jneilliii the initial change was not sufficient, there was a old check in `classicwebcam.js` that prevented the webcam from loading up when it was not embedded in the control tab. I removed this old check as we should fully rely on the `controls.js` to call the proper callbacks when the webcam is shown/hidden

crysxd:
Another small addition: Seemingly the Intersection API cannot handle the webcam being moved to another location on screen. In my example the webcam stayed at the same position but was in another tab. I now allow the intersection API to be reset via `recreateIntersectionObservers()`

crysxd:
We found a nicer solution, initializing the intersection API later in the lifecycle is already enough. Removed the new code again.

crysxd:
And now we have it both:
- onStartupCompleted to initialize the intersection as late as possible
- Separate function to be called to reset the intersection

jneilliii (approved):
These changes were imperative for being able to fix plugin incompatibility.

crysxd:
Thank you! :)

PR #4797, opened 2023-04-18
Don't set ancient date on upload to virtual sd. More features enabled by default in virtual printer.

arekm:
* Drop setting ancient date on files uploaded to virtual sd card, so by default we get current date and that date will be used by OctoPrint.
* Enable extended M20 and LFN_WRITE by default in virtual printer, so people will actually test that when using virtual printer.


foosel (changes requested):
I just realized I forgot to comment with my initial concerns here. I remember when this feature was added, this ancient date was chosen precisely because it mirrors behaviour on various printers out there, what you'll get when you upload a file to the printer's SD card through serial, because the printer has no clue what the current time is (usually no RTC module...) and so it uses this timestamp.

Since the goal of the virtual printer is to behave similar to firmwares out there, I'd prefer to keep this ancient date setting in place.

I'd be fine with just the `EXTENDED_M20`/`LFN_WRITE` changes. At its core however, the virtual printer is a tool for devs - I expect them to adjust settings so they match their current test scenarios, I don't expect people to test all functions of OctoPrint with the virtual printer.

arekm:
That ancient date worked fine when octoprint was showing it in UI. But 4af7bbea5f2877d25910c0c86cbe1862a772df09 introduced MINIMAL_SD_TIMESTAMP and such dates are no longer shown in UI.

Just enabling EXTENDED_M20/LFN_WRITE makes little sense if we never get such dates (unless developer touches files in virtualSd folder or uploads there manually).

So I guess this pull req can be simply closed.

foosel:
> (unless developer touches files in virtualSd folder or uploads there manually).

Which is precisely the intended thing to do here.

> So I guess this pull req can be simply closed.

Roger!

> But [4af7bbe](https://github.com/OctoPrint/OctoPrint/commit/4af7bbea5f2877d25910c0c86cbe1862a772df09) introduced MINIMAL_SD_TIMESTAMP and such dates are no longer shown in UI.

For the record, what the UI does or doesn't show isn't relevant IMHO for how the virtual printer should behave, we need to make sure that parsing of dates works, including the ancient ones, *especially* when making sure that the aforementioned commit still does what it should be doing, so mirroring the behaviour of firmware out there makes the most sense here really.

PR #4798, opened 2023-04-20
M115 extra params

0r31:
Hi,

The goal of this PR is to add the ability to the virtual printer to report extra params provided by M115 report in Marlin firmware.
Here are the kind of extra params M115 can return : https://github.com/MarlinFirmware/Marlin/blob/91cce2a7b39f1b64e37224fe6bef4ce00463fecf/Marlin/src/gcode/host/M115.cpp#L224

Regards

### DIFF
--- a/src/octoprint/plugins/virtual_printer/__init__.py
+++ b/src/octoprint/plugins/virtual_printer/__init__.py
@@ -55,6 +55,10 @@ class VirtualPrinterPlugin(
                 "EMERGENCY_PARSER": True,
                 "EXTENDED_M20": False,
             },
+            "m115ReportExtraParams": True,
+            "extraParams": {
+                "area": "{full:{min:{x:0.00,y:0.00,z:0.00},max:{x:200.00,y:200.00,z:200.00}},work:{min:{x:0.00,y:0.00,z:0.00},max:{x:200.00,y:200.00,z:200.00}}}",
+            },
             "m114FormatString": "X:{x} Y:{y} Z:{z} E:{e[current]} Count: A:{a} B:{b} C:{c}",
             "m105TargetFormatString": "{heater}:{actual:.2f}/ {target:.2f}",
             "m105NoTargetFormatString": "{heater}:{actual:.2f}",
--- a/docs/development/virtual_printer.rst
+++ b/docs/development/virtual_printer.rst
@@ -192,37 +192,44 @@ There many configuration options via ``config.yaml`` for the virtual printer tha
        m115FormatString: "FIRMWARE_NAME: {firmware_name} PROTOCOL_VERSION:1.0"
 
        # Whether to include capability report in M115 output
-       m115ReportCapabilites: false
+       m115ReportCapabilities: false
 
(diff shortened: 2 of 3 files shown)


src/octoprint/plugins/virtual_printer/__init__.py
```diff
@@ -55,6 +55,10 @@ def get_settings_defaults(self):
                 "EMERGENCY_PARSER": True,
                 "EXTENDED_M20": False,
             },
+            "m115ReportExtraParams": True,
+            "extraParams": {
+                "area": "{full:{min:{x:0.00,y:0.00,z:0.00},max:{x:200.00,y:200.00,z:200.00}},work:{min:{x:0.00,y:0.00,z:0.00},max:{x:200.00,y:200.00,z:200.00}}}",
+            },
```
foosel, line 61: From what I understand, this is tied to a capability (`M115_GEOMETRY_REPORT`), so don't make it `extraParams` or something like that but rather tie it to the configured capabilities here as well.

Also, this should return more realistic parameters, 10x10x10cm buildvolumes are rather uncommon I'd say. I suggest going with the values from the default printer profile here. The virtual printer could even read that from the global config and return it here to be consistent. Ideally this should use place holders and be considered a format string for that matter.

foosel (changes requested):
As mentioned on Discord, before this can be merged, documentation of this output format in [Marlin's own docs](https://marlinfw.org/docs/gcode/M115.html) is required, and also expected in the [GCODE wiki node](https://reprap.org/wiki/G-code#M115:_Get_Firmware_Version_and_Capabilities). Only then can we assume a certain amount of guaranteed stability and rely on this for implementation. That last thing I want is trying to shoot a moving target with this over the next handful of releases - code changes quickly, only when stuff is documented it can be depended on.

From what I understand, this is tied to a capability (`M115_GEOMETRY_REPORT`), so don't make it `extraParams` or something like that but rather tie it to the configured capabilities here as well.

Also, this should return more realistic parameters, 10x10x10cm buildvolumes are rather uncommon I'd say. I suggest going with the values from the default printer profile here. The virtual printer could even read that from the global config and use it here to be consistent. Then just format it accordingly on output, no need for config parameters apart from a check of "capability set or not".

0r31:
Actually, I added `extraParams`  as it is not echoed the same way other capabilities are ("Cap:" prefix).

```
Recv: Cap:AUTOREPORT_TEMP:1
Recv: Cap:AUTOREPORT_SD_STATUS:1
Recv: Cap:AUTOREPORT_POS:0
Recv: Cap:EMERGENCY_PARSER:1
Recv: Cap:EXTENDED_M20:0
Recv: area:{full:{min:{x:-3.75,y:0.00,z:0.00},max:{x:430.00,y:320.00,z:360.00}},work:{min:{x:-3.75,y:0.00,z:0.00},max:{x:430.00,y:320.00,z:360.00}}}
```

Regarding dimension values, i totally understand what you mentioned. Will do that.

0r31:
@foosel : i am closing this PR and will create another one according to what you said.

PR #4799, opened 2023-04-20
Handles M115 GEOMETRY capability report

0r31:
Hi,

The goal of this PR is to add the ability to the virtual printer to report GEOMETRY capability provided by M115 report in Marlin firmware.
For more information check https://github.com/MarlinFirmware/Marlin/blob/91cce2a7b39f1b64e37224fe6bef4ce00463fecf/Marlin/src/gcode/host/M115.cpp#L224

( replaces #4798 )

Regards


0r31:
Hi @foosel ,

@thinkyhead merged the PR i did regarding geometry report documentation in Marlin (https://github.com/MarlinFirmware/MarlinDocumentation/pull/496).

It is already online : https://marlinfw.org/docs/gcode/M115.html

foosel:
I have to admit that I'm very much confused right now. I understood it so far that this was a capability in itself. However as I read it now in the official docs it's an *additional line* generated when a specific config setting is provided. Is this correct?

If so, my last review was wrong and we need to do this differently. Can you confirm that it's only a Marlin config setting that then generates an additional line in the `M115` report, not related to the capability reports themselves? If yes, I'll see that I do the fix for that myself and then merge.

foosel:
Roger, sorry then for the initial confusion that caused you unnecessary work. I'll take care of basically merging the prior PR and this one and will get it merged, so that the virtual printer supports this. Then we can proceed with comm layer parsing.

0r31:
> Roger, sorry then for the initial confusion that caused you unnecessary work. I'll take care of basically merging the prior PR and this one and will get it merged, so that the virtual printer supports this. Then we can proceed with comm layer parsing.

No pb, i may have misexplained what's going on here.
If i could help in merging stuff, tell me. I can create another clean PR which will be a kind of merge of #4798 and this one.

PR #4820, opened 2023-06-05
log extruder and file position on pause

danielkucera:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

It logs extruder and file position to log in case of failure during pause state.

I run it during printing and the log showed added values

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -1691,6 +1691,9 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
             position=self._comm.cancel_position.as_dict()
             if self._comm and self._comm.cancel_position
             else None,
+            fileposition=self._comm.getFilePosition()
+            if self._comm
+            else None,
             action_user=user,
         )
         if payload:
@@ -1702,11 +1705,13 @@ class Printer(PrinterInterface, comm.MachineComPrintCallback):
                 {"type": "cancel", "label": "Cancel"},
             )
             self._logger_job.info(
-                "Print job cancelled - origin: {}, path: {}, owner: {}, user: {}".format(
+                "Print job cancelled - origin: {}, path: {}, owner: {}, user: {}, fileposition: {}, position: {}".format(
                     payload.get("origin"),
                     payload.get("path"),
                     payload.get("owner"),
                     payload.get("user"),
+                    payload.get("fileposition"),
+                    payload.get("position"),
                 )
             )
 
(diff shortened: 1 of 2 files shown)


src/octoprint/printer/standard.py
```diff
@@ -1744,12 +1744,19 @@ def on_comm_print_job_paused(self, suppress_script=False, user=None):
         )
         if payload:
             eventManager().fire(Events.PRINT_PAUSED, payload)
+            if "position" in payload:
+                position = str(payload["position"])
+            else:
+                position = "unknown"
+            fileposition = str(self._comm.getFilePosition())
```
foosel, line 1751: `self._comm` might be `None` under certain (very unlikely but still possible) circumstances, so this needs a null-check.

src/octoprint/printer/standard.py
```diff
@@ -1744,12 +1744,19 @@ def on_comm_print_job_paused(self, suppress_script=False, user=None):
         )
         if payload:
             eventManager().fire(Events.PRINT_PAUSED, payload)
+            if "position" in payload:
+                position = str(payload["position"])
+            else:
+                position = "unknown"
```
foosel, line 1750: We can get rid of that and instead copy the fetch of `origin` etc below

src/octoprint/printer/standard.py
```diff
@@ -1744,12 +1744,19 @@ def on_comm_print_job_paused(self, suppress_script=False, user=None):
...
+            fileposition = str(self._comm.getFilePosition())
             self._logger_job.info(
-                "Print job paused - origin: {}, path: {}, owner: {}, user: {}".format(
+                "Print job paused - origin: {}, path: {}, owner: {}, user: {}, fileposition: {}, position: {}".format(
                     payload.get("origin"),
                     payload.get("path"),
                     payload.get("owner"),
                     payload.get("user"),
+                    fileposition,
+                    position,
```
foosel, line 1759:
```suggestion
                    payload.get("position"),
```

foosel (changes requested):
First of all, I agree that getting that into the log makes sense 👍️

However, since the log line currently pretty much copies what's inside the event (minus the position), I wonder if it wouldn't make sense to extend the event here as well to include the file position (origin and file name are already there, your current code duplicates that information as it's also part of the `getFilePosition` response). 

And then we probably should also do the same for cancelling - not that much of interested in case of an issue during pausing, but rather very much helpful for people who want to attempt to rescue a print that got cancelled.

Event extension then also needs some docs.

Also, some small issues, see individual comments.

foosel:
Also, as I just noticed, this should be filed against `maintenance`, not `devel`. It's an improvement of existing functionality, not a biiiiiig new feature that requires a ton of testing and lengthy integration processes first.

danielkucera:
@foosel , I've put everything into event data and included cancel event.

danielkucera:
It's updated, please check

PR #4823, opened 2023-06-06
improve pause gcode

danielkucera:
Without this command, steppers get disabled after a while (at least in Marlin) and axes can move during maintenance operations ( filament change, magnets inserts, etc.)


foosel:
There doesn't seem to be an agreement across firmwares whether that `S` parameter is supported or not. It seems that Prusa does support it, potentially Marlin main as well. However, Klipper is marked as unsupported here - and actually even Prusa is. 

So I'm a bit unsure on whether this should be merged as is. I wouldn't see an issue if we left it commented out in the script with a note to tell people to comment it in if they know their printer supports it.

danielkucera:
@foosel , please see latest version

Issue #4822, opened 2023-06-05
webcam preview and settings causing the page to freeze and crash in Firefox


TheCorruptedEngineer:
### The problem

every time the control tab is opened and the webcam attempts to load it causes the site to freeze and crash. This persists across multiple devices running Firefox (even the mobile version) and isn't caused by any installed extensions, the freezing of the website also happens inside of the settings. Other browsers running on Chromium are able to load the webcam preview and run the website fine.

### Did the issue persist even in safe mode?

Yes, it did persist

### Version of OctoPrint

1.9.0

### Operating system running OctoPrint

Android

### Printer model & used firmware incl. version

Ender3 Marlin 1.1.4

### Browser and version of browser, operating system running browser

Firefox 113.0.2 (64-bit)

### Checklist of files to include below

- [X] Systeminfo Bundle (always include!)
- [X] Contents of the JavaScript browser console (always include in cases of issues with the user interface)
- [X] Screenshots and/or videos showing the problem (always include in case of issues with the user interface)
- [ ] GCODE file with which to reproduce (always include in case of issues with GCODE analysis or printing behaviour)

### Additional information & file uploads

[octoprint-systeminfo-20230605204651.zip](https://github.com/OctoPrint/OctoPrint/files/11657810/octoprint-systeminfo-20230605204651.zip)
J8s36m6o?usp=sharing)

https://github.com/OctoPrint/OctoPrint/assets/101986148/e12f84ef-10e3-4b66-b6c7-f4e539484e8f

[Browser console.txt](https://github.com/OctoPrint/OctoPrint/files/11657812/Browser.console.txt)

cp2004:
Firefox has been fine using mjpg streams since OctoPrint existed, but this is the second report (first was a convo on discord) of it choking when loading a stream recently, but other browsers worked fine. Wonder if it's a bug with a specific version perhaps.

TheCorruptedEngineer:
> Firefox has been fine using mjpg streams since OctoPrint existed, but this is the second report (first was a convo on discord) of it choking when loading a stream recently, but other browsers worked fine. Wonder if it's a bug with a specific version perhaps.

I am having this issue on several devices running firefox, including the mobile Firefox version on my phone

foosel:
Hm... Firefox is my primary browser on both desktop and mobile. I'm currently on 113.0.1 (64-bit) on desktop. Can't say I've noticed anything out of the ordinary.

[...]

PR #4824, opened 2023-06-06
:bug: Reduce intersection observer threshold

cp2004:
When this value is 1, scrolling down the page triggers the value to change - probably something changes invisibly. Setting it to anything less than 1 seems to cure the issue, and there are no longer unwanted triggers of changes.

Added logging to the 'Classic Webcam' plugin, and my own for when the visibility changed.

I haven't tested it with *just* the classic webcam plugin installed, to see whether it is only triggered by something I have done.

Once this was changed, the plugins stopped logging visibility changes on scrolling and the stream was stablez

Working on CameraStreamerControl

None

The video I captured is a large file, too much for here, but it is posted to discord alright

https://discord.com/channels/704958479194128507/708230829050036236/1115668233874321549

#### Further notes

This may be my fault, I'm not sure.

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -122,7 +122,7 @@ $(function () {
                     var options = {
                         root: document.querySelector("#webcam_plugins_container"),
                         rootMargin: "0px",
-                        threshold: 1.0
+                        threshold: 0.9
                     };
                     var callback = function (entries) {
                         var visible = entries[0].isIntersecting;
--- a/src/octoprint/static/js/app/viewmodels/control.js
+++ b/src/octoprint/static/js/app/viewmodels/control.js
@@ -57,19 +57,20 @@ $(function () {
         self._visibleWebcam = undefined;
 
         self._dispatchWebcamVisibilityChange = function (target, visible) {
+            log.debug(`Webcam visibility of #${target.id} changed to ${visible}`);
             var vm = ko.dataFor(target.children[0]);
             if (vm === self) {
-                console.debug(
+                log.debug(
                     `VM for webcam #${target.id} is not bound, skipping visibility update`
                 );
             } else if (vm === undefined) {
-                console.debug(
+                log.debug(
                     `VM for webcam #${target.id} not found, skipping visibility update`
                 );
             } else if (typeof vm.onWebcamVisibilityChange === "function") {
                 vm.onWebcamVisibilityChange(visible);
             } else {
-                console.debug(
+                log.debug(
                     `VM for webcam #${target.id} does not declare 'onWebcamVisibilityChange(visible)', skipping visibility update (vm=${vm.constructor.name})`
                 );
             }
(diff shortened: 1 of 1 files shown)


foosel:
I'm wondering, could this possibly explain #4822 ?

foosel:
Hm, reading through MDN's entry on the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) I'm wondering why 100%/1.0 was chosen in the first place here, and what potential side effects we could see by lowering it.

I mean, from my understanding we basically check here how much of the webcam interacts with the control tab. What I now wonder is, why does that change when scrolling at all? Are we maybe only looking at the *visible* part of the control tab? If so, shouldn't that rather be set to 0.01/1% then?

Any thoughts?

cp2004:
The below video demonstrates the issue that I found with the observer at 1.0. It is quite strange as it doesn't seem to behave how I thought it would. From what I can see, the observer is bound to the right element, like it's not attached in the wrong place so measuring wrong. In the below case I had UI Customizer enabled as I was able to demonstrate the problem easier with the long column scrolling, but initially I did find it with core OP - it's not a plugin issue.

I'm still confused by it.

https://github.com/OctoPrint/OctoPrint/assets/31997505/28a98595-5cba-410e-950b-a292b1ecab6c

foosel:
Maybe @crysxd has some ideas?

crysxd:
I can't comment on why `1` was choosen - I found this solution online, tested it and it worked as expected. I never had the need to change this value. From a logic point of view, I'd agree that `0.01` should do the job fine. As long as any part of the webcam is shown we keep the stream active. 

@cp2004 just for my understanding, the video shown is with `1`?

cp2004:
> @cp2004 just for my understanding, the video shown is with 1?

Correct, yes

crysxd:
I don't see any reason why it would behave like that. Might just be a bug in the intersection implementation?

If a lower value solves this issue, I'd just go for it. But then I think 0.01 is the best choice. Basically 1 means "If the webcam is fully visible, start the stream" and 0.01 means "If a small section of the webcam is visible, start the stream". The latter seems to be more desirable for me as it also would speed up the connection a bit when you scroll back to the webcam. Also gives us the most buffer to 1, which is known to be broken. 0.9 might still trigger the same issue in rare cases depending on screen resolution etc....

foosel:
Hm... So I'm trying to reproduce the original issue. Checked out the PR, changed the value back to 1.0, configured a webcam stream, adjusted the classic webcam to log visibility changes, and also added a debug output to the triggering of the visibility change callbacks. I can't seem to make this happen. Have changed the size of the window, scrolling like mad, tried both Firefox and Chromium, can't reproduce this at all 😐️ Checked with the testwebcam plugin and classic one enabled, so two webcams just like in your case as well.

@cp2004 just to make sure, have you also tried this on a "clean" setup?

Intuitively, I'd say we change this to 0.01 regardless, because it simply makes more sense for us, but still, I don't get why I cannot reproduce this here at all.

foosel:
Will go with 1% for 1.9.1 and just hope it doesn't break anything while fixing the issues that @cp2004 observed.

PR #4830, opened 2023-06-16
Optimized Translation

bigfoxtail:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

Optimized Translation

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,7 +1,7 @@
 [main]
 host = https://www.transifex.com
 
-[octoprint.messages]
+[o:octoprint:p:octoprint:r:messages]
 file_filter = translations/<lang>/LC_MESSAGES/messages.po
 source_file = translations/messages.pot
 source_lang = en
--- a/src/octoprint/templates/sidebar/state.jinja2
+++ b/src/octoprint/templates/sidebar/state.jinja2
@@ -11,7 +11,7 @@
 <span title="{{ _('Current timelapse configuration')|edq }}">{{ _('Timelapse') }}</span>: <strong data-bind="text: timelapseString"></strong><br>
 <!-- /ko -->
 <!-- ko foreach: filament -->
-<span data-bind="text: 'Filament (' + name() + '): ', title: 'Filament usage for ' + name()"></span><strong data-bind="text: formatFilament(data())"></strong><br>
+<span data-bind="text: '{{ _('Filament')|edq }} (' + name() + '): ', title: 'Filament usage for ' + name()"></span><strong data-bind="text: formatFilament(data())"></strong><br>
 <!-- /ko -->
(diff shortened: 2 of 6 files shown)


.tx/config
```diff
@@ -1,7 +1,7 @@
 [main]
 host = https://www.transifex.com
 
-[octoprint.messages]
+[o:octoprint:p:octoprint:r:messages]
```
foosel, line 4: Could you please explain what this is supposed to achieve?

src/octoprint/templates/sidebar/state.jinja2
```diff
@@ -11,7 +11,7 @@
 <span title="{{ _('Current timelapse configuration')|edq }}">{{ _('Timelapse') }}</span>: <strong data-bind="text: timelapseString"></strong><br>
 <!-- /ko -->
 <!-- ko foreach: filament -->
-<span data-bind="text: 'Filament (' + name() + '): ', title: 'Filament usage for ' + name()"></span><strong data-bind="text: formatFilament(data())"></strong><br>
+<span data-bind="text: '{{ _('Filament')|edq }} (' + name() + '): ', title: 'Filament usage for ' + name()"></span><strong data-bind="text: formatFilament(data())"></strong><br>
```
foosel, line 14: This will fail if the translation for whatever reason contains single quotes (it should never do this, but still, this code is wrong this way).

Given that we have a bit of a longer text here with some placeholders, it would probably make more sense to turn this into a formatting string with actual placeholders...

Also, "Filament usage for ..." is still not translatable, should probably also get adjusted as its user visible as well.

foosel (changes requested):
Thank you for this!

Apart from what the PR validation already mentioned, could you please also look into the points mentioned below? Thank you!

.tx/config
bigfoxtail: The Transifex client has been updated. Using an outdated configuration file will result in an error.
<https://developers.transifex.com/docs/migrating-from-older-versions-of-the-client>

src/octoprint/templates/sidebar/state.jinja2
bigfoxtail: Furthermore, I didn't see any tooltip or hint for "Filament usage for..." on the page when I hovered my mouse over the content.

.tx/config
foosel: Ah, gotcha!

src/octoprint/templates/sidebar/state.jinja2
foosel: > Furthermore, I didn't see any tooltip or hint for "Filament usage for..." on the page when I hovered my mouse over the content.

Indeed, that's a bug in the current code 😅 Sets `title: ...`, should be setting `attr: {title: ...}`.

I'll take a look at this

foosel (approved):

PR #4739, opened 2023-02-23
Adding A Remote Access OctoPrint Setup Wizard Page

QuinnDamerell:
* [x] Your changes are not possible to do through a plugin and relevant
    to a large audience (ideally all users of OctoPrint)
  * [x] If your changes are large or otherwise disruptive: You have
    made sure your changes don't interfere with current development by
    talking it through with the maintainers, e.g. through a
    Brainstorming ticket
  * [x] Your PR targets OctoPrint's `devel` branch if it's a completely
    new feature, or `maintenance` if it's a bug fix or improvement of
    existing functionality for the current stable version (no PRs
    against `master` or anything else please)
  * [x] Your PR was opened from a custom branch on your repository
    (no PRs from your version of `master`, `maintenance`, or `devel`
    please), e.g. `dev/my_new_feature` or `fix/my_bugfix`
  * [x] Your PR only contains relevant changes: no unrelated files,
    no dead code, ideally only one commit - rebase and squash your PR
    if necessary!
  * [x] Your changes follow the existing coding style
  * [x] If your changes include style sheets: You have modified the
    `.less` source files, not the `.css` files (those are generated
    with `lessc`)
  * [x] You have tested your changes (please state how!) - ideally you
    have added unit tests
  * [x] You have run the existing unit tests against your changes and
    nothing broke
  * [x] You have added yourself to the `AUTHORS.md` file :)

#### What does this PR do and why is it necessary?

Gina and I were discussing offline back in December that it would be nice to add a setup page for remote access, mostly to make it clear to users that doing things such as port forwarding aren't secure at all. Having a page about remote access also helps OctoPrint stay competitive with the new "all in one" solutions, where the 3D printer comes with control software, a cloud service, and an app that's all wrapped up into a closed ecosystem. 

This page informs the user that they should never port forward, links to the remote access plugins category so the user can see all of the plugins, and suggests the top 2 plugins as possible options.

Gina and I were originally talking about allowing this page do to a plugin install directly from the wizard, so then, after one more quick wizard page, the remote access was totally set up. But due to the difficulty of invoking the plugin install from the wizard, this is just the first step of getting something in place.

#### How was it tested? How can it be tested by the reviewer?

I tested this using a local raspberry pi that I flashed to the latest version of OctoPrint, applied the changes to, and then walked through the setup. I tried to test all of the scenarios, such as a fresh install, a user who already had an OctoPrint setup, and such.

#### Any background context you want to provide?

See above :)

#### What are the relevant tickets if any?

None

[...]

PR #4833, opened 2023-06-21
Improve first run wizard re safety hints

foosel:

Safety hints in the first run end dialog have been reworked to be more visually alerting and a checkbox has been added for each that has to be checked in order to proceed with the wizard.

See discussion in #4739

Visual test by setting `server.firstRun` to true, restarting and clicking through the wizard, again and again

Other PR, #4739

image

image

[Screencast from 2023-06-21 16-23-27.webm](https://github.com/OctoPrint/OctoPrint/assets/83657/7cbc8acc-5780-4f32-8c04-30011f9fcf1b)


foosel:
@QuinnDamerell @jneilliii @cp2004 @fieldOfView would love some feedback on this approach 😅

jneilliii:
Looks good to me and isn't any more intrusive to the end user than any of the other wizard pages. I'm assuming this won't impact restore as that happens ahead of this step and OP gets restarted anyway.

QuinnDamerell:
I concur, I think it’s the perfect level of annoyance and yet getting users to actually pay attention.

PR #4870, opened 2023-08-11
Webrtc: fix RTCSessionDescription

havalli:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

Fixed the connection to the webrtc server. 

Tested with a [mediamtx](https://github.com/bluenviron/mediamtx) Webrtc server
Webcam url: "webrtcs://webrtc-server-host/cam/whep"

image

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -180,6 +180,7 @@ date of first contribution):
   * [Christopher Perrin](https://github.com/cperrin88)
   * [Martin Bartsch](https://github.com/kaenguruhs)
   * [Matthias Weis](https://github.com/mad73923)
+  * [Valentin Hamm](https://github.com/havalli)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/static/js/app/helpers.js
+++ b/src/octoprint/static/js/app/helpers.js
@@ -1720,16 +1720,18 @@ var negotiateWebRTC = function (streamUrl) {
             return $.ajax({
                 url: streamUrl,
                 type: "POST",
-                dataType: "json",
-                data: JSON.stringify({
-                    sdp: offer.sdp,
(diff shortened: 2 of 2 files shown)


cp2004:
The (incomplete) WebRTC implementation in OctoPrint is only compatible with the aiortc server it was developed for (by another contributor). This change will probably break that and any similar servers that communicate the same way.

Because there is no standard to the WebRTC offer communication, these should use a Webcam Plugin introduced in 1.9.0, which allows you to customise the connection to whichever server you are using. I've written one for camera-streamer.

havalli:
There is a standard for the WebRTC offer information. [Whep](https://www.ietf.org/id/draft-murillo-whep-02.html). Yes, it is still a draft. But it fits this scenario exactly. (Just consuming a video stream)
My changes follow this standard. Wouldn't it make sense to put the existing logic in an "aiortc" webcam plugin and implement the Whep standard in the Classic Webcam?

foosel:
Are these changes all that is needed to make it Whep compliant? If so I would suggest to add a config option to the plugin, something like `negotiate_flavor`, and depending on whether this is set to `aiortc` or `whep` do one or the other thing.

havalli:
In my setup, it works. I think it's more of a starting point. Whep is also still a draft. I like the configuration option, but my Python skills are very bad. Maybe someone else can implement it.

foosel:
Converting this to draft for now, as we cannot merge it as is. Requires a `negotiate_flavor` configuration option from the current point of view, in order to not break compatibility with existing aiortc based setups.

foosel:
FWIW, latest document on WHEP seems now to be here: https://datatracker.ietf.org/doc/draft-ietf-wish-whep/, the prior link gives a 404.

Still a draft though, currently expiring in a month.

And given how long that's now been in draft-only state, I'd rather not do anything here (trying to hit a potentially moving target) and rather wait until it turns from draft to an RFC and thus something more stable. Thus, I'm closing this.

PR #4874, opened 2023-08-21
🚸 New thermal kill error message

thinkyhead:
Added in MarlinFirmware/Marlin#23373

This PR adds a new string "Thermal Malfunction" to `regex_marlinKillError`. This message is emitted as part of the KILL process when the temperature variance monitor sees a frozen temperature sensor.

Not tested, but it's an obvious change. Marlin can be forced to emit this string to test the host response by sending `M118 Thermal Malfunction`.


foosel:
Thanks! 🚀

PR #4886, opened 2023-09-16
Ability to ignore Printer Notifications based on regex

jatin-47:
Added new regex setting, new Filtering regex setting UI and filter on the incoming messages.

This PR solves the issue #4326 which is a feature request on the ability to ignore Printer Notifications based on regex.
It was necessary as mentioned in the issue.

Entered the following regex into the settings page under the "Filtering Regex" input field:
`(^\d+%)|(^Layer \d+\/\d+)|(^Done In .*)`

Tested using the debug commands of the virtual printer:
* `!!DEBUG:send //action:notification Layer 15/23`
* `!!DEBUG:send //action:notification 28%`
* `!!DEBUG:send //action:notification Done In 02:45` 

The notifications were not shown in the "Printer Notifications" area as well as no popups were generated in all the cases as expected.

Nope

#4326 

image

#### Further notes
Nope


src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
```diff
@@ -16,4 +16,11 @@
             </label>
         </div>
     </div>
+    <div class="control-group" data-bind="css: { error: !regexValid() }">
+        <label class="control-label" for="settings-filterRegex"> {{ _('Filtering Regex') }}</label>
+        <div class="controls">
+            <textarea rows="4" class="block" id="settings-filterRegex" data-bind="value: regex, valueUpdate: 'afterkeydown'"></textarea>
```
jneilliii, line 22: I think changing the value data-bind here to `settings.plugins.action_command_notification.regex` instead of using the one created in onBeforeBinding will fix the saving issue. And also in that perspective, replacing that in the js side as well would remove the need for onBeforeBinding altogether?

src/octoprint/plugins/action_command_notification/static/js/action_command_notification.js
```diff
@@ -7,11 +7,27 @@ $(function () {
...
+        };
+
+        self.regexValid = function () {
+            try {
+                new RegExp(self.regex());
+                return true;
+            } catch (e) {
+                return false;
+            }
+        };
```
jneilliii, line 29: Not sure if this would benefit from being a ko.pureComputed function or not, but relative to my other comment you could change this to use `self.settings.settings.plugins.action_command_notification.regex` and get rid of the internal regex observable and eliminate the onBeforeBinding necessity.

jatin-47:
I made the changes you suggested, but still, the settings are not getting saved!! I can't figure out the reason why this is happening, I am getting this error in the console of my browser:

image

src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
jatin-47: But why is the current code not working? I have seen other code doing the same thing. Although, I made this change you suggested but still it doesn't save the settings.

jatin-47: Actually, I named the variable incorrectly. I was now using custom binding and used the settingsViewModel as the dependency in the custom ViewModel and stored that passes parameter in the self.settings variable so I must use : `settings.settings.plugins.action_command_notification.regex` and NOT `settings.plugins.action_command_notification.regex`

jneilliii (approved):
haven't tested yet, but this looks like it would work as expected.

src/octoprint/plugins/action_command_notification/templates/action_command_notification_settings.jinja2
```diff
@@ -4,16 +4,23 @@
...
+                <input type="checkbox" data-bind="checked: settings.settings.plugins.action_command_notification.enable_popups"> {{ _('Enable popups') }}
                 <span class="help-block">{{ _("Check this if you want notifications to also show up as popups in addition to the notification list in the sidebar") }}</span>
             </label>
         </div>
     </div>
+    <div class="control-group" data-bind="css: { error: !regexValid() }">
+        <label class="control-label" for="settings-filterRegex"> {{ _('Filtering Regex') }}</label>
+        <div class="controls">
+            <textarea rows="4" class="block" id="settings-filterRegex" data-bind="value: settings.settings.plugins.action_command_notification.regex, valueUpdate: 'afterkeydown'"></textarea>
+            <span class="help-inline">{{ _('Use this to write a valid <em>regular expression</em> to hide notifications having the pattern defined in it. Chain multiple filters using "|". For more information, <a href="%%(tut_url)s" target="_blank" rel="noopener noreferer">Regular Expressions</a>')|format(tut_url="https://docs.python.org/3/library/re.html") }}</span>
```
foosel, line 23:
```suggestion
            <span class="help-inline">{{ _('Use this to write a valid <em>regular expression</em> to hide notifications having the pattern defined in it. Chain multiple filters using "|". For more information, please see <a href="%%(tut_url)s" target="_blank" rel="noopener noreferer">the Python docs on regular expressions</a>')|format(tut_url="https://docs.python.org/3/library/re.html") }}</span>
```

src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -60,7 +62,20 @@ def on_event(self, event, payload):
     # ~ SettingsPlugin
 
     def get_settings_defaults(self):
-        return {"enable": True, "enable_popups": False}
+        return {"enable": True, "enable_popups": False, "regex": ""}
```
foosel, line 65: I'd prefer we call that `filter` since it's also stored in `self._filter` and better communicates what that regex is *for*.

src/octoprint/plugins/action_command_notification/__init__.py
```diff
@@ -120,6 +135,10 @@ def action_command_handler(self, comm, line, action, *args, **kwargs):
             return
 
         message = parameter.strip()
+
+        if self._filter.search(message):
+            return
```
foosel, line 140: For debugging purposes, I think it would be a good idea to add a **debug** log output here, so that once enabled the log would clarify why certain notifications don't show up in the sidebar/the popups anymore.

```suggestion
            self._logger.debug(f"Notification matches filter regex: {message}")
            return
```

foosel (changes requested):
Finally got around to do a review as well, sorry for the delay, I was out of office.

First of all, I really like this! 👍️

I only have some minor nitpicks as listed below, once those are taken care of I think this is ready to be merged 😊

foosel (approved):

foosel:
Thanks, looking good now! Merged and will be in 1.10.0! 👍️

PR #4964, opened 2024-03-02
Filename sanitization improvement in download_file function

jacopotediosi:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

The `download_file` function in `/octoprint/OctoPrint/src/octoprint/util/net.py` suffers from a *quasi* - Path Traversal vulnerability:

https://github.com/OctoPrint/OctoPrint/blob/65a5533c5d645f7033f01056c8ef735a938339f0/src/octoprint/util/net.py#L260-L283

As visible on line 277, the filename of the file to be downloaded, taken from the `Content-Disposition` header received from the remote server during file download, is not sanitized and is appended directly to the folder name.

The presence of the `assert path.startswith(folder)` instruction on line 278 prevents Path Traversal towards the parent of the folder, but not within subfolders. Taking for example a case in which the function was called with the intention of storing the file inside `/`, if the server responded with the header `Content-Disposition: attachment; filename=etc/passwd`, the downloaded file would overwrite `/etc/passwd`.

I'm sending a PR instead of using the GitHub Advisory tool because **in the current codebase** the `download_file` function is only called by specifying temporary folders as the folder, so **this vulnerability is not exploitable**. However, I believe that improving filename sanitization in this case is easy and prevents any vulnerabilities that might arise by reusing the `download_file` function differently in the future.

I have verified that the unit tests pass.

I don't seem to see any situations where this change would break anything.

The function I added, `secure_filename`, takes care of sanitizing file names and is specifically designed to pass the sanitized filename to `os.path.join()`, as documented in the [Flask documentation](https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html).

However, I'm open to discussion if you notice any edge cases that I missed.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -182,6 +182,7 @@ date of first contribution):
   * [Martin Bartsch](https://github.com/kaenguruhs)
   * [Matthias Weis](https://github.com/mad73923)
   * [Rob Emery](https://github.com/mintsoft)
+  * [Jacopo Tediosi](https://github.com/jacopotediosi)
 
 OctoPrint started off as a fork of [Cura](https://github.com/daid/Cura) by
 [Daid Braam](https://github.com/daid). Parts of its communication layer and
--- a/src/octoprint/util/net.py
+++ b/src/octoprint/util/net.py
@@ -10,6 +10,7 @@ import netaddr
 import netifaces
 import requests
 import werkzeug.http
+from werkzeug.utils import secure_filename
 
 _cached_check_v6 = None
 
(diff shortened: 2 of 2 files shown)


foosel:
FYI, I've seen this, I'm just still out of commission after a surgery last week. Will look into this once fully recovered.

jacopotediosi:
> FYI, I've seen this, I'm just still out of commission after a surgery last week. Will look into this once fully recovered.

Thanks Gina, I already knew that. Take your time and stay safe out there.

Issue #1313, opened 2016-04-15
[Request] Option to render timelapse later instead of immediately


mwhuss:
Feature Benefits:

Sometimes prints fail or I want to start another print immediately which is not recommended when rendering a timelapse. I'd like the ability to go to "Unrendered Timelapses" and start the render myself instead of it happening immediately after a print finishes.

philphilphil:
I would like this as well, my rpi1 is soo slow and often it's not even worth rendering when the print fails.
Maybe with a new setting "Autostart Timelapse renders" so nothing changes for users who like the auto-start of the render.

If foosel likes this idea i would also be able to implement it :).

foosel:
An "Autostart timelapse renders" checkbox on the timelapse tab sounds great, feel free to go ahead with that @philphilphil! Just a word of a warning, I'm currently a bit behind with reviewing code changes due to having my hands full with the [funding situation](https://plus.google.com/u/0/+OctoprintOrg/posts/5m53jf6cdme), so I don't want to promise to be able to immediately merge it when you send a PR. Sorry in advance for that.

philphilphil:
@foosel no worries :). I think i'll need some days to finish and test it correctly anyway.

mwhuss:
@philphilphil let me know if you need any help testing it

fraternl:
A lot of days have passed since April 18th ;-)
I just created a ticket that's somewhat related:
https://github.com/foosel/OctoPrint/issues/1526

philphilphil:
@fraternl i kind of forgot about this ticket.. I have already some changes from back in April on my computer, i will review them and finish it, still an important feature i think :)

fraternl:
@philphilphil  do you consider https://github.com/foosel/OctoPrint/issues/1526 as well in your changes?

fiveangle:
I can see this isn't getting much traction, but for consideration of posterity, a preference to "Automatically delete Cancelled print time lapses" (low hanging fruit) as well as "Render timelapse when not printing" options would be ideal.  For the when-not-printing scenario, the script could subscribe to the log file (or port to syslog if not already) to monitor when the job is finished, or cancelled, then automatically kick off the render(s).  There would have to be some coordination trigger mechanism so that if OP started a new print job, it would notify the render script to quit or pause render/resume "monitor" mode, waiting for completion, then restart the job.

Oninaig:
Any update(s) on this?

grantx2016:
I would like this also.

[...]

PR #4994, opened 2024-04-19
Timelapse render option

WisdomCode:
This Request wants to take care of the suprisingly old #1313
It offers a combo box for setting what is to be done with the timelapse after a print is finished. To not break previous setups, the default is to render always after a print. However, it is now also possible to render only successful prints after the print is finished, or to never render a print after finishing printing. 

However, I am still not done with this request, because there is a bug I cannot really get my head around. Maybe someone a bit more acquainted can help me out here?

Currently, the render is successfully skipped, however the according print files are still kept as currently active, as if the recording did not stop. This must be due to the timelapse being saved as "current" (the variable in src/octoprint/timelapse.py), but I cannot find where this is happening and how to stop it. A restart of the server fixes it, as the "current" variable gets reinitialized then. 
I am also not really sure how to test it without printing, which is just a bit painful for debugging this efficiently. 
I would appreciate any help. Thank you for this great project @foosel !

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.11
+  python: python3.10
 exclude: ^(src/octoprint/vendor/|src/octoprint/static/js/lib|src/octoprint/static/vendor|src/octoprint_setuptools|tests/static/js/lib|tests/util/_files|scripts/|translations/|.*\.css|.*\.svg)
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.10
+  python: python3.11
 exclude: ^(src/octoprint/vendor/|src/octoprint/static/js/lib|src/octoprint/static/vendor|src/octoprint_setuptools|tests/static/js/lib|tests/util/_files|scripts/|translations/|.*\.css|.*\.svg)
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/docs/api/timelapse.rst
+++ b/docs/api/timelapse.rst
@@ -258,7 +258,7 @@ For timelapse type ``zchange``.
    * - ``renderAfterPrint``
      - 1
      - string
-     - Determines whether the timelapse should be rendered automatically after the print finishes. 
+     - Determines whether the timelapse should be rendered automatically after the print finishes.
 
 .. _sec-api-timelapse-datamodel-config-timed:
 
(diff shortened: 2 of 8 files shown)


foosel:
@WisdomCode Just a quick heads-up, I've seen this and am happy to help, but it might take until next week.

foosel:
So now that 1.10.0 is out and it also looks like it's a fairly solid release, I finally found the time to take a closer look at this :)

First of all, please make sure to [run `pre-commit install`]() in your checkout, as that not having been done caused the build failure - your code changes did not pass the linter and formatter.

Also, this is an improvement of an existing feature, not a huge change, so I've rebased your changes to target the `maintenance` branch (and thus a 1.11.0 release instead of the 2.0.0 release) instead.

Now, about your problem, just to make sure, we are talking about this staying going:

image

right?

If so, the issue is not `current` staying set, the issue is `current._file_prefix` staying set, which is used to determine the `processing` state of the unrendered timelapse.

I'd suggest to introduce a new method on `Timelapse`:

``` python
    def _reset_metadata(self):
        self._image_number = None
        self._in_timelapse = False
        self._gcode_file = None
        self._file_prefix = None

        self._capture_errors = 0
        self._capture_success = 0
```

Then replace all calls to `reset_image_number` with that (`self._reset_metadata`) and make sure to fetch the required information for the render process *before* this happens, to pass to the movie render method:

``` python
        def reset_and_create():
            file_prefix = self._file_prefix
            gcode_file = self._gcode_file

            self._reset_metadata()

            render_unrendered_timelapse(
                file_prefix,
                gcode=gcode_file,
                postfix=None if success else "-fail",
                fps=self._fps,
            )
```

That should take care of things.

I'd also suggest to keep the logic whether to render a timelapse or not in *one* place, preferably `Timelapse.on_print_done` - right now it's there and in `reset_and_create`. Speaking of `reset_and_create`, that also had a bug in referring to the unknown `self.render_failed_print` 😬

Also... what do you think, should we maybe also have a "only render if failed" option? Personally I could see myself using this (I'm not interested in timelapse recordings of all my successful prints, but it can be quite helpful for failed prints).

WisdomCode:
I will work on this next week, I got sick unfortunately...
But beforehand, thanks for your extensive analysis of the problem! You suspected correctly that I was referring to the action column showing the loading/ongoing indicator, and your conclusion sounds plausible, I will check it out! Didn't find the linting explicitly as I was testing the code simply by pasting it into my live system and restarting the service :D. Thanks for pointing it out.

While I am not sure I would want such a feature (as one probably sometimes wants to retry immediatly after a failed print with a changed setting or some manual intervention), but its just very easy to add right now so I would put it in so no furher issue arises from that, just having all possible options available from the start. As they are hidden in the drop-down, there is just no disadvantage from a designer standpoint in adding another option for the user I think. :+1:

foosel:
Get well soon!

WisdomCode:
Alright, I tested never, always, only successful with a failed and a successful print and only failed with a failed print. Meaning beside the scenario of a successful print with the fail option, everything has been tested. As its pretty easy boolean logic, I am pretty confident that this should work as well. 
So I set this as ready for review.

Issue #5039, opened 2024-07-05
[Request] Improve preview of files using "scarf joint seam" (orcaslicer & others)


jepler:
### Is your feature request related to a problem? Please describe.

I've recently started to use the "scarf joint seam" feature of orcaslicer, which can improve the surface finish of many 3d prints.

However, this feature emits XYZE moves which confuse the gcode preview in octoprint, making it report an incorrect number of layers, as well as making the layer-by-layer preview not useful.

### Describe the solution you'd like

Octoprint needs to use a different method to analyze gcode files that use scarf join. However, I don't know enough about the internals to have a concrete suggestion.

### Describe alternatives you've considered

Many slicers emit special comments to delimit the slicer's idea of a layer. Octoprint could (optionally?) look for these in lieu of what it appears to do now, analyze the Z-heights where extrusion takes place. 

### Additional context

I can provide a gcode file reproducer if needed. I'm using OrcaSlicer 2.1.1 on Linux.

This might also improve gcode visualization of vase mode prints, as well.

jneilliii:
This is an interesting challenge. I know @JoveToo was one of the last contributors to the GCode Viewer, wonder if he might have some ideas.

JoveToo:
Please upload a gcode file from a benchie or something with scarf joint seam.

I will take a look.

jepler:
I've updated the initial comment with a zipped gcode file that demonstrated the problem.
> Model size: 154.79mm × 123.30mm × 48.00mm
> Estimated layer height: 0.00mm
> Estimated total print time: 00:49:52
> Layers with extrusion: 13327

PR #5040, opened 2024-07-07
GCodeViewer: Scarf Seam support

JoveToo:
    Brainstorming ticket
    if necessary!
    have added unit tests

This modifies the GCodeViewer to support Scarf Seams.

Current status: DRAFT

This still contains debug logging and isn't entirely finished. I created this so others can take a look.

It works for most layers but in some cases, additional layers are created and it messes up. (for example, if you load the gcode benchy in #5039 , layers at heights 10.6 and 10.8 are not correct, there are multiple copies of those layers, some with  multiple layers merged.)

It was tested with the sample gcode in issue #5039

#5039

### DIFF
--- a/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
+++ b/src/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js
@@ -447,6 +447,11 @@ var doParse = async function () {
     var zLiftLayer = undefined;
     var zLiftMoves = [];
 
+    var scarfJoint = false;
+    var scarfJointMoves = [];
+    var scarfJointPossibleLayer = undefined;
+    var scarfJointActualLayer = undefined;
+
     var dcExtrude = false;
     var assumeNonDC = false;
 
@@ -727,7 +732,7 @@ var doParse = async function () {
         }
 
         if (typeof z !== "undefined" && z !== prevZ) {
-            if (!extrude && !zLift) {
+            if (!scarfJoint && !extrude && !zLift) {
                 // possible z-lift
                 zLift = true;
                 zLiftZ = maxLiftZ = prevZ;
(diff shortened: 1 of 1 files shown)


adelton:
Hello @JoveToo,

I recently hit the GCode Viewer acting out on my GCode files, it turns out the scarf joints were the cause of the problem. I used your WIP code, mounted it to my `docker.io/octoprint/octoprint:latest` container using
```
   -v $(pwd)/pr-5040-worker.js:/usr/local/lib/python3.10/site-packages/octoprint/plugins/gcodeviewer/static/js/viewer/worker.js:ro
```
and it solved all issues on my data -- the layers in the viewer are now nice and sane. I did not notice any ill effect of the change.

I know that you are not quite happy with the code yet, for example with its behaviour of Benchy from https://github.com/OctoPrint/OctoPrint/issues/5039, but I wonder: does the current code in the PR change cause actual regression in any situations? Wouldn't it be an improvement to merge the code as is, give it more exposure, and work on the corner cases afterwards?

adelton:
If I understand the dilemma correctly, it is not about scarf seams specifically but about the concept of layer in general. Where does the previous layer end and a new one starts, to show something that the user could also consider to be a view of a layer.

Looking at a `.gcode` generated by Prusa Slicer (2.9.2), I can see lines
```
;LAYER_CHANGE
;Z:0.2
;HEIGHT:0.2

[...]

;BEFORE_LAYER_CHANGE

[...]

;AFTER_LAYER_CHANGE
```

Could these be used as the ultimate markers, if present?

adelton:
[Basic bodies - cylinder_8m_0.20mm_215C_PLA_ENDER3.gcode](https://github.com/user-attachments/files/21697080/Basic.bodies.-.cylinder_8m_0.20mm_215C_PLA_ENDER3.txt)

Also, looking at the `master` code and the PR code, the logic determining the scarf join seems to be
```
            // we have a Z increase while extruding
            // this is most likely a scarf joint seam
            if (extrude && z > prevZ) {
```

So we consider it part of the current layer when the Z gets increased while extruding.

But nowhere do I see the maximal extruded Z height recorded; there is `maxLiftZ` but that's for the non-extrude `zLiftMoves`, if I read the code correctly.

At least the way PrusaSlicer does the scarf joints, it's not as much about the Z axes moving up, it's that the extrusion is below the the current layer's Z value. So the start of the scarf joint dives under, basically to the height of the previous layer, and then gradually moves up.

If we are looking for some heuristics, could we use this aspect? Basically, once we've extruded at some Z offset for the current layer, any extrusions below that Z offset belong to that layer without any additional logic needed.

JoveToo:
So the main problem I encountered is not recognizing zlift, layer height and scarf seams. It is when the slicers combines those and I can't tell one from the other. For example, I can only tell if I am zlifting because we come back down without extruding. So I can't actually tell we are in a zlift until we come back down. But it is entirely possible that we come back down, into a scarf seam (so below the current layer zHeight -- a layer that hasn't started yet) and then move back up to a new layer. The problem is recognising that the previous move was a zlift and this one gcode command both terminates the zlift and starts the scarf seam and thus a new layer. But I don't actually know this is a scarf seam until it goes gradually up. But moving a bit up is also an indication of the start of a new layer. The code gets really complex and will be difficult to maintain without understanding all these subtleties. I really don't think this is a good approach, so I abandoned it.

And, indeed, I was considering using the comments in the gcode to tell whether we are starting a new layer, this would solve all the problems and all that other zlift, scarf seam and even layer detection logic isn't necessary anymore.
The main problem with that is that I would need to recognize and understand all the comments from all the possible slicers. And since comments are optional, who knows when they are present and when not?

PR #5060, opened 2024-09-13
Manage groups of remote user based on header

MaienM:

This extends the existing functionality to automatically manage the users based on headers provided by an auth proxy in front of OctoPrint to include group management. This allows one to use an external source of users with varying levels of access.

By configuring a header that contains a comma-separated list of groups and optionally a mapping of the names in this list to the names of the matching OctoPrint groups (in case these don't match) this will manage the list of groups that a user belongs to.

I've mostly tested this by configuring it to look at the groups header that OAuth2 Proxy sends. (Note that it doesn't include this header by default, the `--set-xauthrequest` flag enables this.) It can also be tested by simply including these headers in a manual request.

Example configuration:

```yaml
accessControl:
  trustRemoteUser: true
  addRemoteUsers: true
  remoteUserHeader: 'X-Auth-Request-Preferred-Username'
  remoteGroupsHeader: 'X-Auth-Request-Groups'
  remoteGroupsMapping:
    foo: readonly
    bar: users
```

This can be tested by performing a login request with e.g. curl:

```shell
curl 'http://127.0.0.1:8083/api/login?passive' -X POST -H 'X-Auth-Request-Preferred-Username: testuser' -H 'X-Auth-Request-Groups: foo,bar'
```

In the output of this request it can be observed that the user is in the `readonly` and `users` groups. (Most fields have been omitted from this output for the sake of brevity.)

```json
{
  "name": "testuser",
  "groups": [
    "readonly",
    "users"
  ]
}
```

Repeating the request with one of the groups removed will result in the user also losing said group:

```shell
curl 'http://127.0.0.1:8083/api/login?passive' -X POST -H 'X-Auth-Request-Preferred-Username: testuser' -H 'X-Auth-Request-Groups: foo'
```

```json
{
  "name": "testuser",
  "groups": [
    "readonly"
  ]
}
```

No.

None that I am aware of.

N/A.

#### Further notes

N/A.

### DIFF
--- a/src/octoprint/server/util/__init__.py
+++ b/src/octoprint/server/util/__init__.py
@@ -234,6 +234,17 @@ def get_user_for_remote_user_header(
         )
         user = octoprint.server.userManager.find_user(userid=header)
 
+    if user and settings().getBoolean(["accessControl", "trustRemoteGroups"]):
+        groupHeader = request.headers.get(
+            settings().get(["accessControl", "remoteGroupsHeader"])
+        )
+        if groupHeader:
+            groups = groupHeader.split(",")
+            mapping = settings().get(["accessControl", "remoteGroupsMapping"])
+            if mapping:
+                groups = [mapping.get(group, group) for group in groups]
+            octoprint.server.userManager.change_user_groups(header, groups)
+
     if user:
         _flask.session["login_mechanism"] = LoginMechanism.REMOTE_USER
         _flask.session["credentials_seen"] = datetime.datetime.now().timestamp()
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -186,6 +186,7 @@ date of first contribution):
   * [Jeff Mixon](https://github.com/zaventh)
   * [lauren n. liberda](https://selfisekai.rocks/)
   * [Nathan Schulte](https://github.com/nmschulte)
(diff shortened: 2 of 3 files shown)


src/octoprint/server/util/__init__.py
```diff
@@ -234,6 +234,17 @@ def get_user_for_remote_user_header(
         )
         user = octoprint.server.userManager.find_user(userid=header)
 
+    if user:
+        groupHeader = request.headers.get(
+            settings().get(["accessControl", "remoteGroupsHeader"])
+        )
+        if groupHeader:
```
foosel, line 241: I think, just to be on the safe side, we should also check for a `trustRemoteGroups` setting here.

I could imagine existing setups out there were whatever is sending the user header is also already sending an empty group header, and that would immediately break the existing user group assignment then. Having to enable the use of sent groups would avoid this.

The risk is minimal, but it's there, and I'd rather make sure that this change doesn't have any chance to break existing workflows.

foosel (changes requested):
Love it! Just one open point, see the comment below. Once that's done this can be merged!

foosel:
Thank you! :+1: Merged and ready for 1.11.0!

PR #5070, opened 2024-10-18
🔒 Use secrets lib to generate Flask secret key

jacopotediosi:
    Brainstorming ticket --> Not large/disruptive
    if necessary!
    with `lessc`) --> No style sheets
    have added unit tests
    nothing broke

The current server implementation generates the Flask secret key using `random.choice`, which is not a cryptographically secure pseudo-random number generator (CSPRNG) and should not be used for cryptographic purposes. In fact, [it is known](https://stackered.com/blog/python-random-prediction/) that an attacker can predict the outputs of the `random` library after analyzing a few hundred of them.

However, it should be noted that this attack is difficult or impossible to exploit, as it requires the attacker to be able to observe the outputs of the `random` library before the Flask secret key is stored in the yaml configuration file when OctoPrint is first started. For this reason, this vulnerability is to be considered merely theoretical and is handled via a Pull Request rather than a Github Advisory.

This PR fixes secret key generation using the `secrets` library, as suggested by the Flask [documentation](https://flask.palletsprojects.com/en/latest/config/#SECRET_KEY).

All OctoPrint installations prior to this PR will retain the previous secret key, which should not pose a security concern as this attack is not even retroactive.

I tested this PR by applying it to the `maintenance` branch in a virtualenv with python 3.10. It passed all unit tests.

By deleting the `server.secretKey` entry from the `~/.octoprint/config.yaml` file, I verified that OctoPrint recreates a different new secret key inside the file after a restart.

The `secrets` library that introduces this PR does not appear to be used anywhere else in the OctoPrint code, but [it is native from python3.6 onwards](https://docs.python.org/3/library/secrets.html). OctoPrint dropped support for older Python versions years ago. Users with older versions can still apply an alternative fix by implementing `os.urandom` or installing the [python2-secrets](https://pypi.org/project/python2-secrets/) library.

### DIFF
--- a/src/octoprint/access/users.py
+++ b/src/octoprint/access/users.py
@@ -4,9 +4,9 @@ __copyright__ = "Copyright (C) 2014 The OctoPrint Project - Released under terms
 import hashlib
 import logging
 import os
+import secrets
 import shutil
 import time
-import uuid
 
 import wrapt
 from flask_login import AnonymousUserMixin, UserMixin
@@ -1453,7 +1453,7 @@ class SessionUser(wrapt.ObjectProxy):
     def __init__(self, user):
         wrapt.ObjectProxy.__init__(self, user)
 
-        self._self_session = "".join("%02X" % z for z in bytes(uuid.uuid4().bytes))
+        self._self_session = secrets.token_hex()
         self._self_created = self._self_touched = time.monotonic()
 
     @property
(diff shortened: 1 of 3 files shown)


foosel:
Oh, it's definitely a good idea to get rid of this tech debt.

I'd go one step further however and use `secrets` not only for flask here, but also for generating any API keys or session identifiers. I've adjusted the PR accordingly.

Might also make sense to include this in the next bugfix release.

foosel:
I'm aware, which is why for the API keys I switched to `token_urlsafe` instead of `token_hex`. 

And the idea in switching away from uuid here is rather because it felt like a weird approach in any case to use a uuid but convert it manually to a hex string, when what we actually want is a secure random url safe token.

PR #5073, opened 2024-10-26
include file data payload in onBeforePrintStart viewModel callback

### LINKED ISSUE #4951: [Request] extend onBeforePrintStart to include selected file data

While developing the Bambu plugin I was attempting to use the onBeforePrintStart viewmodel callback to pop my own print options modal prior to starting. If you use the load and print button from the file list the filesViewModel.selectedfile property is not populated with the file that the button was pressed from but the previously selected file if any. This means I have no way of knowing which file was selected to print. 

Would like to get the selected file data object (including name, path, storage, refs, etc.) added to the onBeforePrintStart callback to allow for further processing.  

* monkey patching filesViewModel.loadfile (probably will be my interim solution)
* reordering the loadfile function to have the callbacks after file selection so that filesViewModel.selectedfile is populated prior
* disabling the load and print button entirely and force selecting the file

### Additional context

_No response_

jneilliii:
    Brainstorming ticket
    if necessary!
    with `lessc`)
    have added unit tests
    nothing broke

Adds a second parameter to the onBeforePrintStart viewModel callback that includes selected file details. This allows plugins utilizing the callback to have access to the started file's name, path, and origin (at minimum, see notes below) to use for further processing. 

Locally in development environment utilizing the callback in my Bambu plugin to validate the data was coming through. 

resolves #4951 

#### Further notes
It seems that the printerStateViewModel does not include the full file metadata that is provided from the filesViewModel "load and print" button, so the data object passed to the onBeforePrintStart callback varies based on where you start the print from in the UI. I attempted to get around this by adding a dependency to filesViewModel in printerstate.js so I could pass filesViewModel.selectedFile to the callback, but that broke the UI completely due to a dependency loop. Not sure if there's a better way to handle this, like include all the metadata from the server side for data.job in `self._processJobData(data.job);`; but that seems like it might be too much information to send over the websocket continuously. 

Should not impact any other plugins currently using the callback, as it was added as second parameter.

### DIFF
--- a/docs/plugins/viewmodels.rst
+++ b/docs/plugins/viewmodels.rst
@@ -225,11 +225,12 @@ onUserPermissionsChanged(user)
    will be provided as only parameter. Note that this may also be triggered for not logged in guests if the guest
    group is modified. In this case ``user`` will be undefined.
 
-onBeforePrintStart(callback)
+onBeforePrintStart(callback, data)
    Called before a print is started either by clicking the "Print" button in the state panel or the select & print icon
-   in the file list. The callback to actually proceed with starting the print is provided as the only parameter. By returning
+   in the file list. The callback to proceed with starting the print is provided as first parameter. By returning
    ``false`` from this, plugins may prevent a print from actually starting, optionally starting it at a later date by
-   calling ``callback`` themselves. This can be used for example to implement an additional confirmation dialog.
+   calling ``callback`` themselves. This can be used for example to implement an additional confirmation dialog. The
+   second parameter is a file data object that includes details such as name, path, and origin.
 
 onTabChange(next, current)
    Called before the main tab view switches to a new tab, so `before` the new tab becomes visible. Called with the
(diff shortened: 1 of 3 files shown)


jneilliii:
Had an aha moment when I woke up this morning and realized I could just assign the printerStateViewModel file data from the filesVewModel load function so that it can be consistent between both start print methods.

foosel:
I've changed things around a bit because it felt extremely wrong to modify data of one viewmodel directly from another one, and this also wouldn't have worked in case of another client changing the selected file.

So instead, I now sync the selected file data through the file API, which means one additional backend request whenever the file *changes* - which I think should not cause issues. I also refactored the access to name, path etc to not be its own observable but rather a `pureComputed`, that way we don't have keep the data in several places, but only the object fetched from the API is the source of everything.

jneilliii:
That totally makes sense to me and thanks for taking the time to review and adjust. I was curious about another client connecting, but figured if a print was already started they wouldn't be able to select a file anyway, so thought it wasn't that big of a deal.

PR #5256, opened 2026-02-17
fix(achievements): distinguish between failed and cancelled prints

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
- [ ] You have tested your changes (please state how!) - ideally you
  have added unit tests
  nothing broke (`pytest`)

This PR fixes a broken comparison that always evaluated to `True`, causing achievements for cancelled prints to be assigned also for failed prints.

This one affected also the `main` branch, so I think it could be bugfix relevant.

I didn't really test this, it just seems legit.

No

### DIFF
--- a/src/octoprint/plugins/achievements/__init__.py
+++ b/src/octoprint/plugins/achievements/__init__.py
@@ -346,7 +346,7 @@ class AchievementsPlugin(
             self._data.stats.print_duration_total += payload["time"]
             self._current_year_stats.print_duration_total += payload["time"]
 
-            if Events.PRINT_CANCELLED:
+            if event == Events.PRINT_CANCELLED:
                 self._trigger_achievement(
                     Achievements.ALL_BEGINNINGS_ARE_HARD, write=False
                 )


foosel:
Hmm... you know, the code as it currently is is obviously broken, but I'm seriously wondering if the outcome isn't better.

Imagine a problem with the printer causing your first print to fail, and you get an achievement for that. Isn't that maybe a good idea too? I'm seriously wondering what's the better user experience here, the way it is (accidentally) now, or the way it will be after this change (only actually actively cancelled prints count).

Currently unsure.

jacopotediosi:
I think so too. Those 3 current achievements seem good to obtain even in case of failed prints, maybe we'd just need to adapt their description.

jneilliii:
I think they should all be lumped together. Yes you can cancel a print for a non-failure reason, but it's usually something like oh...I loaded the wrong filament, or need to adjust my slicer settings because I messed up.

foosel:
Have gone with removing the broken `if` altogether, effectively lumping together failed and cancelled prints intentionally, and changed the description of the related achievements.

PR #5263, opened 2026-02-19
fix(ui): prevent double password prompt on change

jacopotediosi:
When changing the current user's password via OctoPrint Settings -> Access Control, a re-authentication prompt was triggered. This is redundant because the user is already required to enter their current password in the change password form.

  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

This PR prevents the re-authentication prompt from appearing when changing your own password via OctoPrint Settings -> Access Control. The check was redundant, as the current password is already required within the change password form.

Reading the code, there was already logic in place to avoid re-authentication in this case, but `self.isCurrentUser()` was being called without passing the user whose password was being changed, thus it always returned `false`. The naming convention for these things seems confusing to me (`self.currentUser()` is an observable of the user whose password is being changed, not the actual current user), but I guess we'll have to live with it.

This bug affects both the `dev` and `main` branches (bugfix release relevant).

I verified the following:
- Before the fix, re-authentication was requested whether changing your own password or another user's password.
- After the fix, re-authentication is only requested when changing another user's password.
- The password change continues to work correctly both when changing your own password and those of other users.

No

### DIFF
--- a/src/octoprint/static/js/app/viewmodels/access.js
+++ b/src/octoprint/static/js/app/viewmodels/access.js
@@ -342,7 +342,8 @@ $(function () {
                         });
                 };
 
-                if (self.isCurrentUser()) {
+                const user = self.currentUser();
+                if (self.isCurrentUser(user)) {
                     proceed();
                 } else {
                     access.loginState.reauthenticateIfNecessary(proceed);


foosel:
> The naming convention for these things seems confusing to me (`self.currentUser()` is an observable of the user whose password is being changed, not the actual current user), but I guess we'll have to live with it.

I agree, that is very much confusing. I'll see if this can maybe be changed with a rename (+ compat layer for old calling code)

PR #5273, opened 2026-02-21
fix(ui): improve progress bars readability

### LINKED ISSUE #5267: Timelapse: text overlaying progress bar partially unreadable
### The problem

1. set up timelapse
2. finish a long print
3. attempt to watch the progress of the rendering

While the white text on the blue (finished) part is perfectly readable, the remaining part (that contains the lion's share of the print name, and the percentage) isn't since the background is light grey. Dragging the mouse over this part makes the text visible though

### If you could not test in safe mode, please state why ("currently printing" is NOT an excuse!)

This is a design issue of the web interface and not a fault related to a plugin.

### Version of OctoPrint

1.11.7

OctoPi 1.1.0

Firefox 147.0.4 on macos Sonoma

Prusa MK3S 3.14.1

- [x] Systeminfo Bundle (always include!)

N/A

jacopotediosi:
  Brainstorming ticket
  if necessary!
- [x] If your changes include style sheets: You have modified the
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

Improve readability in progress bars (see #5267).

I increased the text size, made it permanently black and bold, and lightened the blue color of the progress bar by 30%.

Since the color change to white is no longer performed, there is no longer a need to swap the `progress-text-back` class with `progress-text-front` in the Jinja files based on progress percentage. The `progress-text-front` class remains defined to maintain backward compatibility with plugins that may still use it.

See screenshots below. Other parts of OctoPrint where the progress bar appears might need further testing; the fix is global but I have only tested it within the timelapse section.

No.

Fixes #5267.

Before the fix:

before

After the fix:

after


steve8x8:
Like it ;-) (IMHO the blue background could be even lighter as long as there's enough contrast at the transition point)

jneilliii:
I wonder if this wouldn't be a better solution? 

https://stackoverflow.com/questions/21909494/how-to-create-a-progressbar-with-inverted-text-color-for-progressed-part

jacopotediosi:
> I wonder if this wouldn't be a better solution?
> 
> https://stackoverflow.com/questions/21909494/how-to-create-a-progressbar-with-inverted-text-color-for-progressed-part

It requires a change of HTML structure and I don't see any way to remain backwards compatible with plugins that use the current CSS classes. If we want to go this route, it's a good idea to also create a helper/wrapper to create the progress bars.

I can try to work on it, but not before 03/03/2026.

foosel:
I think I have a better solution (also with a new ko `progressbar` binding) based on the link posted by @jneilliii:

``` html
<div data-bind="progressbar: { percentage: progressString, text: progressBarString }"></div>
```

``` js
ko.bindingHandlers.progressbar = {
    init: (element, valueAccessor, allBindings, viewModel, bindingContext) => {
        const bar = $("<div class='bar'></div>");
        const textBack = $("<span class='progress-text-back'></span>");
        const textFront = $("<span class='progress-text-front'></span>");

        $(element)
            .addClass("progress progress-text-centered")
            .empty()
            .append(bar)
            .append(textBack)
            .append(textFront);
    },
    update: function (element, valueAccessor, allBindings, viewModel, bindingContext) {
        const val = ko.utils.unwrapObservable(valueAccessor());

        const defaultOptions = {
            text: "",
            percentage: 0,
            active: false
        };

        const options = {};
        _.each(defaultOptions, function (value, key) {
            options[key] = ko.utils.unwrapObservable(val[key]) || value;
        });

        $(".bar", element).css("width", `${options.percentage}%`);
        $(".progress-text-back", element)
            .text(options.text)
            .css("clip-path", `inset(0 0 0 ${options.percentage}%)`);
        $(".progress-text-front", element)
            .text(options.text)
            .css("clip-path", `inset(0 calc(100% - ${options.percentage}%) 0 0)`);

        if (options.active) {
            $(".bar", element).addClass("progress-striped active");
        } else {
            $(".bar", element).removeClass("progress-striped active");
        }
    }
};
```

image

Doesn't need any changes on the CSS (so plugins are safe), and works like things used to work before we got rid of a lib for performance reasons in #4105.

foosel:
Closing as I went with e8c4c66bf599577bc09a07c422c5b16352fb7fe5 instead - but thank you for the work in any case!

I think going back to the look of things before #4105 got merged (now that the browser support for CSS stuff is better) is matching the general behaviour of progress bars out there better.

A new ko-binding `progressbar` is available too, and existing progressbars in third party plugins should not be affected at all, as the CSS didn't have to be changed at all.

PR #5294, opened 2026-03-11
fix: wrong FILE_SELECTED path on printer upload

jacopotediosi:
- [x] You have read through `CONTRIBUTING.md`
- [x] Your changes are not possible to do through a plugin and relevant
  to a large audience (ideally all users of OctoPrint)
- [ ] If your changes are large or otherwise disruptive: You have
  made sure your changes don't interfere with current development by
  talking it through with the maintainers, e.g. through a
  Brainstorming ticket
- [x] Your PR targets OctoPrint's `dev` branch
- [x] Your PR was opened from a custom branch on your repository
  (no PRs from your version of `main`, `bugfix`, `next` or `dev`
  please), e.g. `wip/my_new_feature` or `wip/my_bugfix`
- [x] Your PR only contains relevant changes: no unrelated files,
  no dead code, ideally only one commit - rebase and squash your PR
  if necessary!
- [ ] If your changes include style sheets: You have modified the
  `.less` source files, not the `.css` files (those are generated
  with `lessc`)
- [x] You have tested your changes (please state how!) - ideally you
  have added unit tests
- [x] You have run the existing unit tests against your changes and
  nothing broke (`pytest`)
- [x] You have run the included `pre-commit` suite against your changes
  and nothing broke (`pre-commit run --all-files`)
- [x] You have added yourself to the `AUTHORS.md` file :)

#### What does this PR do and why is it necessary?

While trying to copy the file "a.gcode" from Local to Printer storage, I got this stacktrace:

~~~stacktrace
2026-03-11 23:44:08,192 - octoprint.printer.standard.job - INFO - Print job selected - origin: local, path: a.gco, owner: None, user: None
2026-03-11 23:44:08,194 - octoprint.plugins.serial_connector.serial_comm - INFO - Changing monitoring state from "Operational" to "Sending file to SD"
2026-03-11 23:44:08,198 - octoprint.plugins.serial_connector.serial_comm - INFO - Finished in 0.006 s. Approx. transfer rate of 164.314 lines/s or 6.086 ms per line
2026-03-11 23:44:08,199 - octoprint.plugins.file_check - WARNING - Error raised by native grep, falling back to python implementation: grep: /home/user/.octoprint/upload/a.gco: No such file or directory
2026-03-11 23:44:08,199 - octoprint.plugins.file_check - ERROR - Something unexpectedly went wrong while trying to search for {travel_speed} in /home/user/.octoprint/upload/a.gco
Traceback (most recent call last):
[...]

PR #5297, opened 2026-03-12
fix(storage): fix wrong storage on UploadJob

foosel:

Changes the `UploadJob`'s storage to `printer` when uploading to the printer's SD card, deprecates `local` key on related event payloads and adjusts frontend accordingly.

Uploading files to the printer from the UI.

Hahahaha... no.

As discussed in #5294

#5294

#### Further notes

Draft for now as dinner's ready and I can't wait for tests, CI, do your thing.

### DIFF
--- a/docs/events/index.rst
+++ b/docs/events/index.rst
@@ -415,7 +415,7 @@ FileSelected
 
    * ``name``: the file's name
    * ``path``: the file's path within its storage location
-   * ``origin``: the origin storage location of the file, either ``local`` or ``sdcard``
+   * ``origin``: the origin storage location of the file, either ``local`` or ``printer``
 
    .. versionchanged:: 1.4.0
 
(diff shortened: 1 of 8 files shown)


jacopotediosi:
Looks good to me.

I noticed that when loading `a.gcode` the following appears in the logs:
```stacktrace
2026-03-12 19:10:37,613 - tornado.access - WARNING - 404 GET /api/files/printer/a.gco (127.0.0.1) 3.87ms
```
This happens because the State sidebar tries to load data from the destination file, which obviously doesn't exist yet.
However it doesn't seem like a big problem to me.

foosel:
I noticed that yesterday as well and was about to say, we can't do much about it, and then I got the idea to simply flag upload jobs in the state data and now that's a thing of the past too.

PR #5309, opened 2026-03-14
fix: fix terminal suppression

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

Suppression checkboxes in Terminal tab were not working anymore because regexes were looking for `Recv:` and `Send:` but since 0b12b9494 the serial connector was logging `<<<` and `>>>`.
This PR proposes to revert to `Recv:` and `Send:` to fix this bug.

To be honest, I preferred `<<<` and `>>>` myself, because they were shorter and not language (English) dependent. However, I did a quick `grep` and found out that some known plugins (`atxpihat`, `bettergrblsupport`, `prettygcode`, `terminalmessaging`) actually read theese logs and look for the old format. So, if we want to keep the new format, I think it should be warned as a breaking change. Also we'll need to update multiple residues in the codebase and documentation that still references `Recv:` and `Send:`.

This bug affected only the `dev` branch. 

Connect to the virtual printer, go to the Terminal tab and check all the suppression checkboxes. Before this fix they don't work. After this fix they work.

No

### DIFF
--- a/src/octoprint/plugins/serial_connector/serial_comm.py
+++ b/src/octoprint/plugins/serial_connector/serial_comm.py
@@ -4278,10 +4278,10 @@ class MachineCom:
 
         if ret != "":
             try:
-                self._log(f"<<< {sanitize_ascii(ret)}")
+                self._log(f"Recv: {sanitize_ascii(ret)}")
             except ValueError as e:
                 self._log(f"WARN: While reading last line: {e}")
-                self._log(f"<<< {ret!r}")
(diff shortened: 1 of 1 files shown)


foosel:
> So, if we want to keep the new format, I think it should be warned as a breaking change.

One more reason then to go with 2.0.0 for the upcoming release, vs 1.12.0.

foosel:
I've created a task issue for the communication of the breaking change here.

PR #5315, opened 2026-03-16
chore(access): minor code improvements

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

This PR fixes some low hanging fruits I found while reviewing code in the access package.

These are just small improvements. Neither issue had practical impact. For further details, please refer to the commit description.

These harmless bugs also exists in `main` branch, so this PR could be bugfix release relevant.

Actually nothing to test, these bugs were harmless. Pytest passed before this PR and still passes.

No

### DIFF
--- a/src/octoprint/access/groups.py
+++ b/src/octoprint/access/groups.py
@@ -472,9 +472,9 @@ class FilebasedGroupManager(GroupManager):
             added_subgroups = list(set(subgroups) - set(group._subgroups))
 
             if removed_subgroups:
-                self._dirty = group.remove_subgroups_from_group(removed_subgroups)
+                self._dirty |= group.remove_subgroups_from_group(removed_subgroups)
             if added_subgroups:
-                self._dirty = group.add_subgroups_to_group(added_subgroups)
+                self._dirty |= group.add_subgroups_to_group(added_subgroups)
 
             notifications.append(
                 (
@@ -611,7 +611,7 @@ class Group:
 
         dirty = False
         for permission in permissions:
-            if permissions not in self.permissions:
+            if permission not in self.permissions:
                 self._permissions.append(permission)
                 dirty = True
 
--- a/src/octoprint/access/groups.py
+++ b/src/octoprint/access/groups.py
@@ -472,9 +472,9 @@ class FilebasedGroupManager(GroupManager):
             added_subgroups = list(set(subgroups) - set(group._subgroups))
 
             if removed_subgroups:
-                self._dirty |= group.remove_subgroups_from_group(removed_subgroups)
+                self._dirty = group.remove_subgroups_from_group(removed_subgroups) or self._dirty
             if added_subgroups:
-                self._dirty |= group.add_subgroups_to_group(added_subgroups)
+                self._dirty = group.add_subgroups_to_group(added_subgroups) or self._dirty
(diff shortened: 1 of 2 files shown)


src/octoprint/access/groups.py
```diff
@@ -472,9 +472,9 @@ def update_group(
             added_subgroups = list(set(subgroups) - set(group._subgroups))
 
             if removed_subgroups:
-                self._dirty = group.remove_subgroups_from_group(removed_subgroups)
+                self._dirty |= group.remove_subgroups_from_group(removed_subgroups)
```
foosel: That's a bitwise or, not a logical or. While this might work, it's semantically incorrect. I'm also not sure about the execution order here, so let's make this explicit by putting the `or` expression at the end, to make sure the remove is always evaluated before calculating the overall result.

```suggestion
                self._dirty = group.remove_subgroups_from_group(removed_subgroups) or self._dirty
```

src/octoprint/access/groups.py
```diff
@@ -472,9 +472,9 @@ def update_group(
...
             if added_subgroups:
-                self._dirty = group.add_subgroups_to_group(added_subgroups)
+                self._dirty |= group.add_subgroups_to_group(added_subgroups)
```
foosel: Bitwise vs logical, as mentioned.


```suggestion
                self._dirty = group.add_subgroups_to_group(added_subgroups) or self._dirty
```

foosel (changes requested):

jacopotediosi:
@foosel about the bitwise or, I copied them from the other points in the code, which uses them and are still there:

https://github.com/OctoPrint/OctoPrint/blob/b7255eaf0dd4ed821a8e77a371c323d6e462f6c0/src/octoprint/access/groups.py#L455-L458

https://github.com/OctoPrint/OctoPrint/blob/b7255eaf0dd4ed821a8e77a371c323d6e462f6c0/src/octoprint/access/users.py#L787-L790

foosel:
Ah, good point then, I'll take care of those too. Not sure how that happened.

jacopotediosi:
May I suggest putting `or self._dirty` at the beginning of the assignment and not at the end, to make the code more readable?

Edit: too late :(

foosel:
Can't do that - boolean operators are shortcutting, meaning, this:

``` python
a = a or method()
```

would immediately evaluate to `True` on the first argument `a` if that's `True` and the `method` call would never get executed.

That's what I meant above with

> I'm also not sure about the execution order here, so let's make this explicit

jacopotediosi:
> Can't do that - boolean operators are shortcutting, meaning, this:
> 
> ```python
> a = a or method()
> ```
> 
> would immediately evaluate to `True` on the first argument `a` if that's `True` and the `method` call would never get executed.

Oh, whops, you are right

foosel:
I actually like the short format bitwise or here, but I couldn't find information on whether the `|` is actually overloaded for booleans *officially*, or if we have type coercion here (boolean gets turned to integer and then bitwise or'ed). And as I don't have time to dig through the cpython code at the moment, going with the explicit boolean `or` instead feels like the cleanest solution.

PR #5320, opened 2026-03-18
fix(auth): create remote users w/o password

foosel:
Password can then be set by an admin if needed.

Fixes an issue discovered by @jacopotediosi: the previous code utilized the wrong method to create a password hash, basically generating a fresh global API key whenever a new remote user was added. Thankfully that feature is not much used. On the other side, if it was more used, we would have noticed this earlier for sure.

Quick and dirty PR to be able to tag something as possibly bugfix release relevant.

### DIFF
--- a/src/octoprint/access/users.py
+++ b/src/octoprint/access/users.py
@@ -19,6 +19,8 @@ from octoprint.settings import settings as s
 from octoprint.util import atomic_write, deprecated, generate_api_key, to_bytes, yaml
 from octoprint.util import get_fully_qualified_classname as fqcn
 
+NOLOGIN_PWHASH = "nologin"
+
 password_hashers = []
 
 try:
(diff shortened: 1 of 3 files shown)


jacopotediosi:
Looks good to me.

Since this PR is security relevant (it touches authentication), and there are already unit tests in place in `tests/access/users/test_usermanager.py`, I'd suggest adding one to verify that password checks for nologin users always return `False`, testing the strings `NOLOGIN_PWHASH` (value of), `""`, and `None`.

I've coded a quick one:

```python
    @ddt.data(octoprint.access.users.NOLOGIN_PWHASH, "", None)
    def test_nologin_user_check_password_always_false(self, password):
        user = octoprint.access.users.User(
            "username",
            octoprint.access.users.NOLOGIN_PWHASH,
            True,
            permissions=[],
            apikey="apikey",
            settings={"key": "value"},
        )

        self.assertFalse(user.check_password(password))
```

PR #5321, opened 2026-03-18
feat(auth): add optional trusted auth proxy config

foosel:

If trusted auth proxies (address or range) are configured, the chain of
trusted involved reverse proxies will be checked for any the auth proxies.

Only if an auth proxy was involved in the request will the remote user
header then be evaluated.

Additional security option.

```
mitmdump --mode reverse:http://localhost:5000@5555 --modify-headers "/X-Remote-User/admin" --modify-headers "/X-Forwarded-For/127.0.0.1"
```

with

- default config: login via header
- `accessControl.trustedAuthProxies=["127.0.0.1"]`: login via header
- `accessControl.trustedAuthProxies=["192.168.1.10"]`: no login

and 

```
mitmdump --mode reverse:http://localhost:5000@5555 --modify-headers "/X-Remote-User/admin" --modify-headers "/X-Forwarded-For/192.168.1.10"
```

with

- default config: login via header
- `accessControl.trustedAuthProxies=["127.0.0.1"]`: no login
- `accessControl.trustedAuthProxies=["192.168.1.0/24"]`: login

No.

As suggested by @jacopotediosi

### DIFF
--- a/src/octoprint/schema/config/access_control.py
+++ b/src/octoprint/schema/config/access_control.py
@@ -62,6 +62,18 @@ class AccessControlConfig(BaseModel):
     **ONLY ENABLE THIS** if your OctoPrint instance is only accessible through a connection locked down through an authenticating reverse proxy!
     """
 
+    trustedAuthProxies: Optional[list[str]] = None
+    """
+    Only trust the remote user header if the request came via one of these configured (trusted!) reverse proxies.
+
+    **Unset by default, meaning this will NOT be checked!** If you plan to use an auth proxy with OctoPrint,
+    be sure to also set it's IP address (or a CIDR range covering it) here!
+
+    Please note that only your **trusted reverse proxies** as defined via `server.reverseProxy.trustedProxies` and
+    `server.reverseProxy.trustLocalhostProxies` will be used for a check, so if your trusted auth proxy address
+    isn't part of your trusted reverse proxies, this check will always fail, causing remote user headers to not work!
+    """
+
     remoteUserHeader: str = "REMOTE_USER"
     """Header used by the reverse proxy to convey the authenticated user."""
 
--- a/src/octoprint/schema/config/access_control.py
+++ b/src/octoprint/schema/config/access_control.py
@@ -66,12 +66,8 @@ class AccessControlConfig(BaseModel):
     """
     Only trust the remote user header if the request came via one of these configured (trusted!) reverse proxies.
 
-    **Unset by default, meaning this will NOT be checked!** If you plan to use an auth proxy with OctoPrint,
-    be sure to also set it's IP address (or a CIDR range covering it) here!
-
-    Please note that only your **trusted reverse proxies** as defined via `server.reverseProxy.trustedProxies` and
-    `server.reverseProxy.trustLocalhostProxies` will be used for a check, so if your trusted auth proxy address
-    isn't part of your trusted reverse proxies, this check will always fail, causing remote user headers to not work!
+    If left unset, the configured trusted reverse proxies from `server.reverseProxy.trustedProxies` and `server.reverseProxy.trustLocalhostProxies`
+    will be used. Make sure this includes your auth proxy or set it explicitly here!
     """
 
     remoteUserHeader: str = "REMOTE_USER"
--- a/src/octoprint/schema/config/access_control.py
+++ b/src/octoprint/schema/config/access_control.py
@@ -59,15 +59,36 @@ class AccessControlConfig(BaseModel):
     Whether to trust remote user headers. If you have setup authentication in front of OctoPrint and the user names you use there match OctoPrint
     accounts, by setting this to true users will be logged into OctoPrint as the user provided in the header.
 
-    **ONLY ENABLE THIS** if your OctoPrint instance is only accessible through a connection locked down through an authenticating reverse proxy!
+    Make sure to add your auth proxy address(es) to `trustedAuthProxies` as well, see below!
     """
 
-    trustedAuthProxies: Optional[list[str]] = None
+    trustedAuthProxies: list[str] = []
     """
-    Only trust the remote user header if the request came via one of these configured (trusted!) reverse proxies.
+    Only trust the remote user header if the request came via one of these configured (trusted!) reverse proxie addresses or CIDR ranges.
 
-    If left unset, the configured trusted reverse proxies from `server.reverseProxy.trustedProxies` and `server.reverseProxy.trustLocalhostProxies`
-    will be used. Make sure this includes your auth proxy or set it explicitly here!
+    Must either be included in the set of trusted reverse proxies (see `server.reverseProxy.trustedProxies` and `server.reverseProxy.trustLocalhostProxies`)
+    or directly connect to OctoPrint!
+
+    **WARNING**: If left unset, the remote user header won't ever be trusted!
+
+    Examples:
+
+    ``` yaml
+    accessControl:
+      trustedAuthProxies:
+        - 127.0.0.0/8
+        - ::1
+    ```
+
+    to trust localhost.
+
+    ``` yaml
+    accessControl:
+      trustedAuthProxies:
+        - 192.168.1.128
+    ```
+
+    to trust 192.168.1.128
     """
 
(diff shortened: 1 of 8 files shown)


jacopotediosi:
The purpose of this feature should be to improve the security of an OctoPrint setup with Single Sign-On, by allowing explicit configuration of the Trusted Authentication Server IP and ignoring remote authentication headers if they originate from a different IP.

In the current state of this PR, however, I can see that only the `X-Forwarded-For` header is checked, and the actual source IP of the request is not validated.

This introduces two issues:

1. An attacker able to reach OctoPrint directly can set the authentication server's address in the `X-Forwarded-For` header and authenticate using remote authentication headers, defeating the whole purpose of this feature I mentioned above.
\
Steps to reproduce:
\
OctoPrint config:
    ```yaml
    accessControl:
      remoteUserHeader: X-Remote-User
      trustRemoteUser: true
      trustedAuthProxies: 192.168.1.1
    server:
      reverseProxy:
        trustedProxies:
        - 192.168.1.1
    ```
    Attacker (their IP is `127.0.0.1`, but they can still log in by spoofing `X-Forwarded-For: 192.168.1.1`):
    ```bash
    curl http://127.0.0.1:5000/login -H "X-Forwarded-For: 192.168.1.1" -H "X-Remote-User: admin" 
    <!doctype html>
    <html lang=en>
    <title>Redirecting...</title>
    <h1>Redirecting...</h1>
    <p>You should be redirected automatically to the target URL: <a href="/">/</a>. If not, click the link.
    ```

2. In the typical scenario where the authentication server sits directly in front of OctoPrint with no intermediate reverse proxies, its IP is not present in the `X-Forwarded-For` header (which instead contains the end user's IP), so it would not be considered trusted and the SSO login would not work.

Finally, by default `trustedAuthProxies` is `None`, meaning no check is performed on the authentication server's IP. This goes against secure-by-default design. I understand this choice was made to avoid breaking existing SSO setups, but we could instead take advantage of the fact that version 2.0.0 will already introduce breaking changes and address this as well.

---

Personal note:

Command to run mitmproxy without forcing the `X-Forwarded-For` header and without hitting CORS/CSRF errors when visiting `localhost:5555`:
```bash
mitmdump --mode reverse:http://localhost:5000@5555 --modify-headers "/X-Remote-User/admin" --set keep_host_header=true
```

foosel:
You are right, that's not going to work with the current implementation, both due to 1 and 2.

However, we have a general problem here.

> In the typical scenario where the authentication server sits directly in front of OctoPrint with no intermediate reverse proxies

This most likely is NOT going to be the typical scenario, and also the reason why I *cannot* just use the original client ip of the request as that might not be the auth proxy but a child of it.

Imagine a setup with a typical OctoPi instance and some authentication proxy in front:

```
client --> [ auth proxy ] --> [ haproxy --> OctoPrint ]
```

The remote IP of the request here will be `127.0.0.1` (or `::1`).

The auth proxy's IP will be sitting in the FF header, as will the actual client ip.

The auth proxy is the instance adding the remote user header. There's no way to see that however, the header doesn't keep that information. All I have is the knowledge of the original requesting ip (effectively localhost) and the first non trusted client IP from the FF header (the real client). I can hope that if the request comes over the remote proxy chain, the auth proxy would have removed any kind of spoofed remote user headers from the incoming request. 

The only kind of trust I could establish here is basically

``` python
trusted_ff = get_trusted_proxies_in_ff(request)
if (auth_proxy in trusted_ff and request_ip in trusted_proxies)
    or (request_ip == auth_proxy):
   trust_remote_user_header()
```

I'm currently not sure though if changing the logic that way will prevent any kind of spoofing 🤔 It's all quite delicate.

---

About the default value of `None`, I see your point, but even going with 2.0.0 I do *not* want to open the floodgates now of breaking stuff left and right. I think we might be able to achieve the same by going with a default of falling back to all trusted proxies here?

jacopotediosi:
> Imagine a setup with a typical OctoPi instance and some authentication proxy in front:
> 
> ```
> client --> [ auth proxy ] --> [ haproxy --> OctoPrint ]
> ```
> 
> The remote IP of the request here will be `127.0.0.1` (or `::1`).
> 
> The auth proxy's IP will be sitting in the FF header, as will the actual client ip.
> 
> The auth proxy is the instance adding the remote user header. There's no way to see that however, the header doesn't keep that information. All I have is the knowledge of the original requesting ip (effectively localhost) and the first non trusted client IP from the FF header (the real client). I can hope that if the request comes over the remote proxy chain, the auth proxy would have removed any kind of spoofed remote user headers from the incoming request.
> 
> The only kind of trust I could establish here is basically
> 
> ```python
> trusted_ff = get_trusted_proxies_in_ff(request)
> if (auth_proxy in trusted_ff and request_ip in trusted_proxies)
>     or (request_ip == auth_proxy):
>    trust_remote_user_header()
> ```
> 
> I'm currently not sure though if changing the logic that way will prevent any kind of spoofing 🤔 It's all quite delicate.

This looks like it could work, but it's important that `get_trusted_proxies_in_ff()` does chain-walking right-to-left and stops at the first untrusted address, like `get_http_client_ip()` already does. The `get_trusted_forwarded_for_addresses()` in this PR doesn't do that, leaving open the case where an attacker bypasses the auth proxy and reaches a trusted intermediate  proxy (e.g. haproxy) directly, spoofing the auth proxy's IP in the FF header.

---

> About the default value of `None`, I see your point, but even going with 2.0.0 I do _not_ want to open the floodgates now of breaking stuff left and right. I think we might be able to achieve the same by going with a default of falling back to all trusted proxies here?

I think that whoever configured SSO before this PR didn't add the authentication server to the trusted proxies (it wasn't specified in the docs to do so), so the default you're suggesting would also break their setup.

foosel:
> it's important that `get_trusted_proxies_in_ff()` does chain-walking right-to-left and stops at the first untrusted address, like `get_http_client_ip()` already does. The `get_trusted_forwarded_for_addresses()` in this PR doesn't do that, leaving open the case where an attacker bypasses the auth proxy and reaches a trusted intermediate proxy (e.g. haproxy) directly, spoofing the auth proxy's IP in the FF header.

That's been fixed and the discussed logic implemented.

> I think that whoever configured SSO before this PR didn't add the authentication server to the trusted proxies (it wasn't specified in the docs to do so), so the default you're suggesting would also break their setup.

Given that an auth proxy (usually) will be a reverse proxy on the LAN and thus should also NOT be the seen client address ("List of trusted proxy servers for which to ignore the IP address when trying to determine the connecting client's IP address"), I'd definitely consider that scenario a full blown configuration error.

But given that I just confused myself repeatedly with this setting, I guess we will go with the trusted auth proxies defaulting to an empty list and thus this feature requiring some manual configuration to get working.

Something to think about though: With this now in place we could deprecate/migrate the `trustRemoteUser` setting: if it's false we go with the default empty list, if it's true we set the trusted proxy list?

New installations will then start with safer defaults (you have to explicitly add your trusted auth proxy to the list to enable the feature), existing installations will hopefully continue to work like before in most cases. That plus a big fat heads-up in the changelog maybe?

---

Side note:

Turns out caddy is also a nice tool for testing this stuff:

```
caddy reverse-proxy --from :5555 --to :5000 --header-up "X-Remote-User: admin"
```

jacopotediosi:
Just reviewed again, now it looks good to me.

> Something to think about though: With this now in place we could deprecate/migrate the `trustRemoteUser` setting: if it's false we go with the default empty list, if it's true we set the trusted proxy list?
> 
> New installations will then start with safer defaults (you have to explicitly add your trusted auth proxy to the list to enable the feature), existing installations will hopefully continue to work like before in most cases. That plus a big fat heads-up in the changelog maybe?
> 

I agree about the migration to keep that previous setups working.

foosel:
Ok, removed `trustRemoteUser` and added a migration to the settings manager implementing the above logic.

jacopotediosi:
In `get_user_for_remote_user_header` you still use it:
```python
    if not s.getBoolean(["accessControl", "trustRemoteUser"]):
        return None
```

foosel:
Turns out, not only there.

jacopotediosi:
Everything now looks good to me :+1:

foosel:
Just also added a bunch of unit tests

PR #5337, opened 2026-03-24
fix(serial): move serial logging from logging to serial_connector

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

The setting to enable/disable `serial.log` was broken because it pointed to the old path `["serial", "log"]` instead of the new `serial_connector`.

This PR moves both the serial log initialization and the `serial.log is enabled` warning from the `logging` plugin to the `serial_connector`, and fixes the setting paths.

Manually tested (enabling/disabling `serial.log` and restarting OctoPrint. Repeated this procedure multiple times and in various combinations).
The `serial.log is enabled` warning is visually the same and in the same place in the navbar as before.
The warning setting should migrate correctly but I didn't really tested it.

No


foosel (changes requested):
Hm, no, as it is now that tightens the binding between the internal stuff and the serial plugin, when I actually want to reduce it.

And it would also cause hard to detect bugs with other connectors.

I think that's rather a case of having to somehow move that stuff into the serial connector itself (especially the log), or figure out another solution. I'll have to think about it. Leaving this open as a reminder (and/or in case you come up with a better solution before I do 😅)

jacopotediosi:
> I think that's rather a case of having to somehow move that stuff into the serial connector itself (especially the log), or figure out another solution.

Force pushed and updated the first post. Now this PR moves serial log initialization and serial log warning from `logging` plugin to `serial_connector`, as you asked.

Please let me know what you think about.

---

I still need to fix the following line that uses the old path, but I guess we'll do that in a separate PR. To be honest, I don't even know what that setting is and I might need help with that.

https://github.com/OctoPrint/OctoPrint/blob/2988b6c8b5ca2fb193ec7bdd6b499f180e638994/src/octoprint/printer/estimation.py#L43

foosel (approved):
Wanted to fully move anything serial.log related out of the core code into the plugin, but was pretty sure it would be annoying so did want to do it myself. Good that I did, found a bug on the used log handler that's also been in there for ages, and it cost me hours to figure out what the actual issue was. Or rather, how to solve it (I'm not 100% sure what the *actual* issue was, but switching from `self.foo` to `type(self).foo` for accessing a class level parameter solved things finally, whereas `self.__class__.foo` for some reason didn't and at some point I need to do a deep dive into the logging stuff in python to figure out what's going on there). 

Anyhow... I had to refactor some bits on my end to get the `serial.log` stuff out of core, so now I need a review from *you* 😅

Your changes up until mine are fine!

foosel:
Great, for some reason GitHub doesn't allow me to request a review from you officially, probably because it's your own PR.

src/octoprint/plugins/serial_connector/__init__.py
```diff
@@ -86,6 +120,18 @@ def on_settings_migrate(self, target, current):
                         ["plugins", "serial_connector"], config, force=True
                     )
 
+            # TODO: migration should be in version 3
```
foosel: Oh, I forgot about that one, but I'll take care of that myself.

foosel: Done... Unit tests will have to wait until #5340 is merged

jacopotediosi:
Your changes look good. I just added a few commits to fix some additional things I was forgetting:
- fix a docstring in the printer interface that still mentioned `serial.log` instead of `connector's log`
- fix settings APIs that still referenced the `SERIAL` logger, which no longer exists with that name
- add migration for the order/disabling of the `seriallog` component in the navbar, just in case someone configured it
    - regarding this point, during the migration, we could also remove the `logWarning` setting entirely and use it instead to decide whether to disable the `seriallog` component in the navbar, what do you think @foosel?

foosel:
> * regarding this point, during the migration, we could also remove the `logWarning` setting entirely and use it instead to decide whether to disable the `seriallog` component in the navbar, what do you think @foosel?

Am I right in assuming we don't have that setting exposed anywhere on the UI? I couldn't find it, so I'm assuming that was always just a `config.yaml` only setting, so migrating it here to just use the existing "disabled bits of the UI" concept would be fine.

I also found another thing that still needs to be done... we currently no longer produce logging output in `octoprint.log` when `serial.log` is switched on/off. It's not strictly needed, but it does help to match up bits and pieces in both logs, and also allows to enable `serial.log` during an ongoing connection (which in the current implementation no longer works).

So I'll quickly take care of implementing that again. And then I'll do a review of your changes, and you'll please do a review of my changes to get the log output again and enable during ongoing connection, and then we can hopefully finally merge this ^^

foosel:
> we currently no longer produce logging output in `octoprint.log` when `serial.log` is switched on/off. It's not strictly needed, but it does help to match up bits and pieces in both logs, and also allows to enable `serial.log` during an ongoing connection (which in the current implementation no longer works).

Strike that. We didn't log to `octoprint.log`, but we did allow switching while the connection is active. So that's what I'm going to still fix.

jacopotediosi:
> Am I right in assuming we don't have that setting exposed anywhere on the UI?

We don't. Disabling that warning was implemented in c528f64, which mentioned only config.yaml.

foosel (approved):
From my side, this is good to go now. Would love a second pair of eyes on my last two changes though.

jacopotediosi:
Two minor corrections, and...

Some improvements about the log warning migration tests:
- Unified the three test cases into a single parametrized one
- Added a test case: warning disabled and component disabled
- Stricter checks on the exact number of `settings.global_set()` calls for all test cases

foosel (approved):

foosel:
Merged manually as 2fb7d55673a97a7b1cc24b34520a1e86dd7040d3 (because GitHub is having issues again and the merge button above is broken)

PR #5340, opened 2026-03-25
Fix `autoconnect` related backwards compatibility and UX issues

foosel:

Solves the backwards compatibility and UX issues regarding the `autoconnect` settings.

1. `autoconnect` has been added back to `/api/connection` (under `options`)
2. Settings schema of the serial connector no longer contains `port`, `baudrate` and `autoconnect`, the initial settings migration has been adjusted to move those to `printerConnection.[autoconnect|preferred.parameters]` as needed. Unit tests have been added to fully test the migration.
3. A new settings page has been added that allows to *see* the current default connection settings and change the `autoconnect` setting. The behaviour of the auto connect checkbox on the connection sidebar has been changed to switch back to off once the connection command has been triggered, as has the "save as default" checkbox. Small wording changes.

1. curl
2. unit tests
3. manual testing

No

image

#### Further notes

Would love a review from you, @jacopotediosi!

### DIFF
--- a/src/octoprint/plugins/serial_connector/serial_comm.py
+++ b/src/octoprint/plugins/serial_connector/serial_comm.py
@@ -218,10 +218,9 @@ def serialList():
     else:
         candidates = []
         try:
-            with os.scandir("/dev") as it:
-                for entry in it:
-                    if regex_serial_devices.match(entry.name):
-                        candidates.append(entry.path)
+            for entry in os.scandir("/dev"):
+                if regex_serial_devices.match(entry.name):
+                    candidates.append(entry.path)
         except Exception:
             _logger.exception("Could not scan /dev for serial ports on the system")
 
--- a/src/octoprint/plugins/serial_connector/serial_comm.py
+++ b/src/octoprint/plugins/serial_connector/serial_comm.py
@@ -189,7 +189,7 @@ PORT_AUTO = "AUTO"
 BAUDRATE_AUTO = 0
 
 STANDARD_BAUDRATES = [115200, 250000, 230400, 57600, 38400, 19200, 9600]
-
+# standard baudrates, sorted by likelihood
 
 SDFileData = namedtuple("SDFileData", ["name", "size", "timestamp", "longname"])
 
@@ -269,8 +269,7 @@ def serialList():
 
 def baudrateList(candidates=None):
     if candidates is None:
-        # sorted by likelihood
-        candidates = STANDARD_BAUDRATES
+        candidates = STANDARD_BAUDRATES[:]
 
     # additional baudrates prepended, sorted descending
     additionalBaudrates = settings().get(
--- a/src/octoprint/plugins/serial_connector/serial_comm.py
+++ b/src/octoprint/plugins/serial_connector/serial_comm.py
@@ -188,6 +188,8 @@ MINIMAL_SD_TIMESTAMP = 1212012000
 PORT_AUTO = "AUTO"
 BAUDRATE_AUTO = 0
 
+STANDARD_BAUDRATES = [115200, 250000, 230400, 57600, 38400, 19200, 9600]
+
 
 SDFileData = namedtuple("SDFileData", ["name", "size", "timestamp", "longname"])
 
@@ -251,11 +253,16 @@ def serialList():
                 filter(lambda x: not fnmatch.fnmatch(x, pattern), candidates)
             )
 
-    # last used port = first to try, move to start
-    prev = settings().get(["plugins", "serial_connector", "port"])
-    if prev in candidates:
-        candidates.remove(prev)
-        candidates.insert(0, prev)
+    # preferred port = first to try, move to start
+    preferred_connector = settings().get(["printerConnection", "preferred", "connector"])
+    if preferred_connector == "serial":
+        preferred_params = settings().get(
+            ["printerConnection", "preferred", "parameters"]
+        )
+        preferred = preferred_params.get("port")
+        if preferred in candidates:
+            candidates.remove(preferred)
+            candidates.insert(0, preferred)
 
     return candidates
 
@@ -263,7 +270,7 @@ def serialList():
 def baudrateList(candidates=None):
     if candidates is None:
         # sorted by likelihood
-        candidates = [115200, 250000, 230400, 57600, 38400, 19200, 9600]
+        candidates = STANDARD_BAUDRATES
 
     # additional baudrates prepended, sorted descending
     additionalBaudrates = settings().get(
@@ -288,11 +295,16 @@ def baudrateList(candidates=None):
             except ValueError:
                 pass
 
-    # last used baudrate = first to try, move to start
-    prev = settings().getInt(["plugins", "serial_connector", "baudrate"])
-    if prev in candidates:
-        candidates.remove(prev)
-        candidates.insert(0, prev)
+    # preferred baudrate = first to try, move to start
+    preferred_connector = settings().get(["printerConnection", "preferred", "connector"])
+    if preferred_connector == "serial":
+        preferred_params = settings().get(
+            ["printerConnection", "preferred", "parameters"]
+        )
+        preferred = preferred_params.get("baudrate")
+        if preferred in candidates:
+            candidates.remove(preferred)
+            candidates.insert(0, preferred)
 
     return candidates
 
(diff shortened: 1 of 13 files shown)


jacopotediosi:
In my opinion, it seems counterintuitive that the `Auto-connect on server startup` checkbox resets to unchecked, because from a UX standpoint I would expect it to always be bound to the current setting.

Additionally, I found a bug: if I start the server with auto-connect enabled, click the `Disconnect` button, check the `Set as default connection settings` checkbox, leave the `Auto-connect on server startup` checkbox unchecked, and press `Connect`, the API request is sent without the `autoconnect: False` parameter, so it doesn't get saved. It is correctly sent and saved if I check and then uncheck the checkbox, but not if I don't touch it at all.

foosel:
> In my opinion, it seems counterintuitive that the `Auto-connect on server startup` checkbox resets to unchecked, because from a UX standpoint I would expect it to always be bound to the current setting.

I decided against that due to that making it easier to accidentally overwrite your current autoconnect settings with a different connection. This checkbox is now supposed to only allow changing the `autoconnect` setting on connect, same as like the `save` checkbox is only actually doing something during connect. The fact that its side effect happens to be saving a boolean to the settings that could also be shown as a checkbox (and is elsewhere) is just down to us basically setting a flag alongside the connection request. Maybe I should add a hint about where to find that stuff in the settings?

I originally wanted to set it checked if autoconnect is true and the currently selected settings match the preferred ones, but then I ran into an issue where I'd basically have to constantly fetch all parameters from the currently selected connector (it might look like a trivial template, but we are actually looking at some difficult Javascript magic to pull off support for plugins to inject their own form into the connection panel here, and that's not through bindings...). Hence the decision to turn it into a connection flag vs "the current setting is".

Good point on the bug though, and also the `autoconnect` checkbox should probably only work at all/be enabled if `save` is also checked.

jacopotediosi:
`port` and `baudrate` settings from `serial_connector` are still used in the following lines:

https://github.com/OctoPrint/OctoPrint/blob/dc4b9dd2426fb67f3c77f7ab505b7698d70f2574/src/octoprint/plugins/serial_connector/serial_comm.py#L256

https://github.com/OctoPrint/OctoPrint/blob/dc4b9dd2426fb67f3c77f7ab505b7698d70f2574/src/octoprint/plugins/serial_connector/serial_comm.py#L293

jacopotediosi:
> I originally wanted to set it checked if autoconnect is true and the currently selected settings match the preferred ones, but then I ran into an issue where I'd basically have to constantly fetch all parameters from the currently selected connector (it might look like a trivial template, but we are actually looking at some difficult Javascript magic to pull off support for plugins to inject their own form into the connection panel here, and that's not through bindings...). Hence the decision to turn it into a connection flag vs "the current setting is".

Got it.

> the `autoconnect` checkbox should probably only work at all/be enabled if `save` is also checked.

I think this also helps reduce questions about why the `autoconnect` checkbox doesn't reflect the current state. Having to check `save` to enable the `autoconnect` checkbox helps the user understand they are entering a new configuration flow, and that the `autoconnect` checkbox may not reflect the current state, but rather the choice they want to make upon saving.

foosel:
Have fixed both points I hope. Also found a logic bug in the server still (`autoconnect` was saved even if `save` was `false` - would normally say this is bugfix related but it's been like this ever since the start, it can wait until 2.0.0).

src/octoprint/plugins/serial_connector/serial_comm.py
```diff
@@ -252,19 +253,24 @@ def serialList():
...
+            candidates.insert(0, preferred)
 
     return candidates
 
 
 def baudrateList(candidates=None):
     if candidates is None:
         # sorted by likelihood
-        candidates = [115200, 250000, 230400, 57600, 38400, 19200, 9600]
+        candidates = STANDARD_BAUDRATES
```
jacopotediosi: Here `STANDARD_BAUDRATES` is copied by reference. Subsequent modifications to `candidates` also alter it. In particular, I think this causes `STANDARD_BAUDRATES` to grow unboundedly on each call to `baudrateList()` when `additionalBaudrates` are configured.

jacopotediosi: Also, why not make `STANDARD_BAUDRATES` a `set`? I don't expect it to contain duplicates under any circumstances (?)

src/octoprint/static/js/app/viewmodels/connection.js
```diff
@@ -148,7 +206,7 @@ $(function () {
                 connectorParameters,
                 response.current,
                 {connector: self.lastConnector, parameters: self.lastConnectorParameters},
-                response.options.preferredConnector
+                preferredConnector
```
jacopotediosi: Why this change? `preferredConnector` is a string, but `self.onConnectionDataReceived()` in `SerialConnectorViewModel` still expects an object in `response.options.preferredConnector` format. Either this change is accidental, or it needs to be changed there too.

src/octoprint/static/js/app/viewmodels/connection.js
jacopotediosi, line 57: `self.autoconnect` is initialized as `undefined`, so if the user enables autoconnection, reloads the page, disconnects the printer, checks the `Set as default connection settings` checkbox, leaves the `Auto-connect on server startup` checkbox untouched (thus leaving the observable `self.autoconnect` at `undefined`), and clicks `Connect`, the autoconnection setting is not saved and remains enabled instead of being disabled.
Either we should initialize the observable as `false`, or add `self.autoconnect(false)` in `self.fromResponse()`.

tests/plugins/serial_connector/test_settings_migration.py
```diff
@@ -0,0 +1,368 @@
...
+    "current_version, config, expected",
+    [
+        (None, {"blacklistedPorts": []}, {"blocklistedPorts": []}),
+        (1, {"blacklistedPorts": []}, {"blocklistedPorts": []}),
+        (None, {"blacklistedBaudrates": []}, {"blocklistedBaudrates": []}),
+        (1, {"blacklistedBaudrates": []}, {"blocklistedBaudrates": []}),
+    ],
+)
+def test_migration_blocklists(plugin, current_version, config, expected):
+    """Tests migration of blocklisted{ports|Baudrates} to blacklisted{Ports|Baudrates}"""
```
jacopotediosi: This is the opposite?
Should be:
```
"""Tests migration of blacklisted{Ports|Baudrates} to blocklisted{ports|Baudrates}"""
```

src/octoprint/static/js/app/viewmodels/connection.js
```diff
@@ -95,7 +97,61 @@ $(function () {
...
+                response.options.preferredConnector.parameters;
+            const preferredConnectorData = connectors.find(
+                (c) => c.connector === preferredConnector
+            );
+
+            if (preferredConnectorData) {
+                self.preferredConnectorName(preferredConnectorData.name);
+            } else {
+                // fallback
+                self.preferredConnectorName(`<code>${preferredConnector}</code>`);
```
jacopotediosi, line 114: Template binds `preferredConnectorName` as text, so it should not contain HTML tags:
```html
<span data-bind="text: preferredConnectorName"></span>
```

One of the two points is wrong.

jacopotediosi (changes requested):

src/octoprint/plugins/serial_connector/serial_comm.py
foosel: `set`s are unordered, and we a need a order here. OctoPrint shows the most likely baudrates on top (and also tests those first during auto connect) and if you have a preferred baudrate, that gets fully on the top.

Good catch on the reference though, didn't think of that when I quickly moved the defaults out to be able to use them in the unit tests.

tests/plugins/serial_connector/test_settings_migration.py
foosel: Yep.

src/octoprint/static/js/app/viewmodels/connection.js
foosel: Accident 😅

src/octoprint/static/js/app/viewmodels/connection.js
foosel: Should now behave the same as `saveSettings` and get reset to `false` in `fromResponse`

foosel:
Should have solved all of the above... CI is still running through, should hopefully be green, if so consider another review requested here.

Given that GitHub never understood that you gave one, I don't want to request another one officially, not that you'll end up at *two* requested ones you never can solve 😅

jacopotediosi (approved):
All looks good to me

PR #5341, opened 2026-03-25
feat: track job's plate & add plate param to thumbnail api

foosel:

This PR introduces a new `platehint` parameter on the thumbnail related methods of filemanager, storages and printer connectors that allows to specify the plate for which to fetch the thumbnail, adds a `plate` parameter to the print job that also gets pushed to the UI through the current job data and can then be used there to fetch the plate specific thumbnail in the printer/job state panel.

I haven't yet adjusted any connectors, so for now I only tested whether thumbnails still work and the printer state attaches the expected `plate` query parameter.

No

This is related to OctoPrint/OctoPrint-BambuConnector#2

OctoPrint/OctoPrint-BambuConnector#2

#### Further notes

Would love some feedback on this from you, @jneilliii, especially whether it would actually solve the issue in question - you are currently deeper in the plugin than me.

The idea would be to set the `plate` parameter when creating the print job in the connector, and adjusting `get_thumbnail` and `read_thumbnail` to support plate specific ones, and then things should work automatically.

### DIFF
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -1834,6 +1834,7 @@ class Printer(PrinterMixin, ConnectedPrinterListenerMixin):
                             size=None,
                             date=None,
                             upload=None,
+                            plate=None,
                         ),
                         estimatedPrintTime=None,
                         filament=None,
@@ -1873,6 +1874,7 @@ class Printer(PrinterMixin, ConnectedPrinterListenerMixin):
                         if job.date
                         else None,
                         upload=isinstance(job, UploadJob),
+                        plate=job.plate,
                     ),
                     estimatedPrintTime=estimatedPrintTime,
                     filament=filament,
--- a/src/octoprint/printer/standard.py
+++ b/src/octoprint/printer/standard.py
@@ -1834,11 +1834,11 @@ class Printer(PrinterMixin, ConnectedPrinterListenerMixin):
                             size=None,
                             date=None,
                             upload=None,
-                            plate=None,
                         ),
                         estimatedPrintTime=None,
                         filament=None,
                         user=None,
+                        plate=None,
                     )
                 )
                 return
@@ -1874,11 +1874,11 @@ class Printer(PrinterMixin, ConnectedPrinterListenerMixin):
                         if job.date
                         else None,
                         upload=isinstance(job, UploadJob),
-                        plate=job.plate,
                     ),
                     estimatedPrintTime=estimatedPrintTime,
                     filament=filament,
                     user=user,
+                    plate=job.plate,
                 )
             )
             self._selected_job = job
--- a/src/octoprint/filemanager/__init__.py
+++ b/src/octoprint/filemanager/__init__.py
@@ -1385,13 +1385,19 @@ class FileManager:
     def has_thumbnail(self, location, path) -> bool:
         return self._storage(location).has_thumbnail(path)
 
-    def get_thumbnail(self, location, path, sizehint=None) -> Optional[StorageThumbnail]:
-        return self._storage(location).get_thumbnail(path, sizehint=sizehint)
+    def get_thumbnail(
+        self, location, path, platehint: int = None, sizehint: str = None
+    ) -> Optional[StorageThumbnail]:
+        return self._storage(location).get_thumbnail(
+            path, platehint=platehint, sizehint=sizehint
+        )
 
     def read_thumbnail(
-        self, location, path, sizehint=None
+        self, location, path, platehint: int = None, sizehint: str = None
     ) -> tuple[StorageThumbnail, IO]:
-        return self._storage(location).read_thumbnail(path, sizehint=sizehint)
+        return self._storage(location).read_thumbnail(
+            path, platehint=platehint, sizehint=sizehint
+        )
 
     def refresh_thumbnails(
         self, location, path, force: bool = False, recursive: bool = False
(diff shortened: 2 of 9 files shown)


src/octoprint/printer/standard.py
```diff
@@ -1834,6 +1834,7 @@ def _set_job_data(self, job: PrintJob, user: str = None, data: dict = None):
                             size=None,
                             date=None,
                             upload=None,
+                            plate=None,
```
jacopotediosi: `printerstate.js` [expects](https://github.com/OctoPrint/OctoPrint/pull/5341/changes#diff-0681847b29de47c1a2d9ec99139a7fb603d8ce7abb1dcd8bbc06c473f16e001aR476) `plate` to be outside `file`:

```js
self.plate(data.plate);
```

src/octoprint/printer/standard.py
```diff
@@ -1873,6 +1874,7 @@ def _set_job_data(self, job: PrintJob, user: str = None, data: dict = None):
                         if job.date
                         else None,
                         upload=isinstance(job, UploadJob),
+                        plate=job.plate,
```
jacopotediosi: `printerstate.js` [expects](https://github.com/OctoPrint/OctoPrint/pull/5341/changes#diff-0681847b29de47c1a2d9ec99139a7fb603d8ce7abb1dcd8bbc06c473f16e001aR476) `plate` to be outside `file`:

```js
self.plate(data.plate);
```

jacopotediosi (changes requested):

foosel:
Issue should be solved.

Proposed usage in bambu connector specifically would be something like this:

```diff
diff --git a/octoprint_bambu_connector/connector.py b/octoprint_bambu_connector/connector.py
index 6221bdd..fb54340 100644
--- a/octoprint_bambu_connector/connector.py
+++ b/octoprint_bambu_connector/connector.py
@@ -510,6 +510,16 @@ class ConnectedBambuPrinter(
     def supports_job(self, job: PrintJob) -> bool:
         return job.storage == FileDestinations.PRINTER
 
+    def create_job(self, path: str, owner: str = None, params: dict = None) -> PrintJob:
+        plate = 1
+        if params:
+            plate = params.pop("plate_num", 1)
+
+        job = super().create_job(path, owner=owner, params=params)
+        job.plate = plate
+
+        return job
+
     def start_print(
         self, pos=None, user=None, tags=None, params: dict = None, *args, **kwargs
     ):
@@ -541,14 +551,13 @@ class ConnectedBambuPrinter(
         perform_bed_leveling = fetch_param("perform_bed_leveling", converter=bool)
         perform_flow_cali = fetch_param("perform_flow_cali", converter=bool)
         enable_timelapse = fetch_param("enable_timelapse", converter=bool)
-        plate_number = fetch_param("plate_number", converter=int)
 
         self.set_state(ConnectedPrinterState.STARTING)
 
         # TODO: deal with ams_mapping, for now will default to what is set in sliced file
         self._client.print_3mf_file(
             name=path,
-            plate=plate_number,
+            plate=self.active_job.plate,
             bed=PlateType.AUTO,  # Always assume the sliced gcode file has this set correctly
             use_ams=use_ams,
             ams_mapping="",
```

Needs some testing though. First want to know Jim's opinion on the whole approach.

jacopotediosi (approved):

jneilliii:
With these changes I believe it would work fine for when starting a print from OctoPrint UI, but I'm not sure it will pick up the plate when you send from slicer or directly on the printer's panel would it? I'm just assuming that based on the override of create_job not being fired in those circumstances. Based on my initial tests with Bambu Connector, that plate querystring argument isn't changing for me in the UI. 

image

It's still showing the correct thumbnail in state panel because I haven't swapped out the download_thumbnail code yet. The above screenshot was starting the print from the printer's panel and choosing plate 2.

jneilliii:
> but I'm not sure it will pick up the plate when you send from slicer or directly on the printer's panel would it?

this is a non-issue. I realized shortly after posting I just need to update the PrintJob data with that plate information. After doing that, the plate querystring is getting updated in the UI as expected.

jneilliii:
Fully functional best I can tell from this branch on bambu connector, would appreciate non X1C verification. 

https://github.com/OctoPrint/OctoPrint-BambuConnector/tree/fix/thumbnail_updates

jneilliii (approved):
This looks totally fine for me conceptually when starting print jobs from OctoPrint, but

foosel:
> would appreciate non X1C verification.

Will check on tuesday!

foosel:
Created a two plate print job, uploaded it to the A1, started print of plate 2 (still ongoing, is actually the correct one). Thumbnail showed the right plate. Disconnected, reconnected, plate 1 is shown. Checking the data I get from the printer there's no info at all about the plate number. The filename is just `<project>.gcode.3mf`, so no chance to match it out of the name there. Fun thing: Display of the A1 also shows the thumbnail auf plate 1. So at least things are matching up? 😅 Not sure how to proceed here though, as we simply don't seem to get the necessary data from the printer at all.

However, the short time before disconnect/reconnect I actually saw the right thumbnail, so the PR itself works. It's just the Bambustuff that's causing headaches here.

So, if there's nothing going against it apart from the usual implementation problems against this proprietary stuff, I'd like to merge this...

jneilliii (approved):
I'm good with the merge. Curious what bpm version you had in your venv? There were some fixes in one of the newer releases related to plate detection IIRC. I went ahead and bumped the version dependency to the latest one in the plugin.

foosel:
> Curious what bpm version you had in your venv?

That was already 1.0.2 on my end:

```
$ pip freeze | grep bambu
bambu-printer-manager==1.0.2
```

I'm suspecting a general issue with the A1 mini firmware here, as its own display shows the wrong thumbnail (plate 1) when printing plate 2. But we should continue that on the plugin repo.

PR #5353, opened 2026-03-31
chore!: fix/remove long standing TODOs & FIXMEs

foosel:

As we are preparing for release of 2.0.0, it's time to take a hard look at all the `TODO`s and `FIXME`s still in OctoPrint's source, get rid of long standing deprecations, implement stuff that's still missing after the hard refactoring and get rid of some stuff that's actually outdated. This is what this PR is going to do.

Hopefully the CI will catch everything that I don't see during development.

No


src/octoprint/access/users.py
```diff
@@ -1251,20 +992,6 @@ def get_id(self):
...
-
-    @property
-    def is_authenticated(self):
-        return FlaskLoginMethodReplacedByBooleanProperty("is_authenticated", lambda: True)
-
-    @property
-    def is_active(self):
-        return FlaskLoginMethodReplacedByBooleanProperty(
-            "is_active", lambda: self._active
-        )
```
jacopotediosi, line 1266: Removing the `is_active` property causes inactive users to always appear active, since the `User` class inherits from `UserMixin`, where `is_active()` always returns `True`.

We should probably keep it as:
```py
@property
def is_active(self):
    return self._active
```

src/octoprint/util/commandline.py
```diff
@@ -38,24 +37,21 @@ def clean_ansi(line: Union[str, bytes]) -> Union[str, bytes]:
     a future version of OctoPrint. A warning will be logged.
```
jacopotediosi: This docstring still states that the function supports an input of `bytes`

jacopotediosi: At least the "a warning will be logged" statement should be removed, as it is no longer true.

src/octoprint/server/api/files.py
jacopotediosi: Same problem you found in `_isBusy`, here there is still a hardcoded `FileDestinations.LOCAL` instead of `target`.

src/octoprint/access/users.py
jacopotediosi: Yep, confirmed, I just tested and was indeed able to login using an inactive user.

src/octoprint/static/js/app/viewmodels/loginstate.js
```diff
@@ -334,14 +334,12 @@ $(function () {
                 self.username(data.name);
                 self.currentUser(data);
 
-                // TODO: deprecated, remove in 1.5.0
-                self.isUser(data.user);
-                self.isAdmin(data.admin);
+                self.isUser(data.groups.includes("users"));
+                self.isAdmin(data.groups.includes("users"));
```
jacopotediosi: This one should be `admins`?

src/octoprint/static/js/app/viewmodels/connection.js
```diff
@@ -209,6 +254,23 @@ $(function () {
...
 
+            // add listeners to inputs to detect changes & reevaluate connection availability
+
+            const event = "input.revalidate";
+            self.doOnAllInputs((element) => {
+                $(element)
+                    .off(event)
+                    .on(event, () => {
+                        if (self.selectedConnector() === connector) {
+                            self.reevaluateConnectionParameters(connector);
```
jacopotediosi, line 269: `connector` doesn't exist here (?)
Or am I missing some JavaScript magic?

src/octoprint/plugins/backup/__init__.py
```diff
@@ -1156,13 +1156,8 @@ def add_to_zip(source, target, ignored=None):
                             )
 
             except Exception as exc:  # noqa: F841
```
jacopotediosi: You can remove `# noqa: F841`, since `exc` is now used

jacopotediosi (changes requested):

jacopotediosi:
In addition to the review above, I'm reporting that the following points in the frontend still access `admin` even if `User.as_dict()` doesn't contain it anymore:

https://github.com/OctoPrint/OctoPrint/blob/5847190ae8949d6722e1710ae8a6447577a7725c/src/octoprint/static/js/app/client/access.js#L133

https://github.com/OctoPrint/OctoPrint/blob/5847190ae8949d6722e1710ae8a6447577a7725c/src/octoprint/static/js/app/client/access.js#L163

https://github.com/OctoPrint/OctoPrint/blob/5847190ae8949d6722e1710ae8a6447577a7725c/src/octoprint/static/js/app/viewmodels/access.js#L475

src/octoprint/access/users.py
foosel: Ah shit, yeah...

foosel:
Took care of all the points, but while doing so unearthed a whole bunch of missed deprecated stuff. So now there's a massive amount of new commits added and I've converted this back to draft status to see what happens in CI.

src/octoprint/static/js/app/client/base.js
```diff
@@ -485,55 +472,53 @@
...
-        newVar,
+        deprecated,
+        replacement,
         getter,
-        setter
+        setter,
+        deprecatedVersion,
+        removalVersion
     ) {
         Object.defineProperty(object, oldVar, {
```
jacopotediosi: `oldvar` no longer exists here

src/octoprint/plugins/appkeys/static/clientjs/appkeys.js
```diff
@@ -40,12 +40,15 @@
...
-    };
+    OctoPrintAppKeysClient.prototype.revokeKey = OctoPrintClient.deprecated(
+        "OctoPrint.plugins.appkeys.revokeKey",
+        "OctoPrint.plugins.appkeys.revokeKeyForApp",
+        function (key, opts) {
+            return this.base.simpleApiCommand("appkeys", "revoke", {key: key}, opts);
+        },
+        "1.10.0",
+        "3.0.0"
+    );
```
jacopotediosi: Missing `// TODO remove in 3.0.0`

src/octoprint/static/js/app/client/printer.js
```diff
@@ -291,29 +293,34 @@
...
+        OctoPrintPrinterClient.prototype.releaseStorage,
+        "2.0.0",
+        "3.0.0"
+    ); // TODO remove in 3.0.0
+
+    OctoPrintPrinterClient.prototype.refreshSd = OctoPrintClient.deprecated(
+        "OctoPrintClient.printer.refreshSd",
+        "OctoPrintClient.files.listForLocation",
+        function (opts) {
+            this.issueStorageCommand("refresh", {}, opts);
```
jacopotediosi: Missing `return`

src/octoprint/static/js/app/viewmodels/slicing.js
```diff
@@ -16,7 +16,14 @@ $(function () {
         self.defaultProfile = undefined;
 
         self.destinationFilename = ko.observable();
-        self.gcodeFilename = self.destinationFilename; // TODO: for backwards compatibility, mark deprecated ASAP
+
+        self.gcodeFilename = OctoPrintClient.deprecated(
+            "SlicingViewModel.gcodeFileName",
```
jacopotediosi: Should be `SlicingViewModel.gcodeFilename` (`n` is lowercase)

src/octoprint/static/js/app/viewmodels/files.js
```diff
@@ -666,25 +644,8 @@ $(function () {
...
-                    focus = [{location: arguments[2], path: arguments[1]}];
-                } else {
-                    focus = [{location: "local", path: arguments[1]}];
-                }
-                if (arguments.length > 3) {
-                    switchToPath = arguments[3] || undefined;
-                }
-            }
+            const focus = params.focus || [];
+            const switchToPath = params.switchToPath || undefined;
```
jacopotediosi: Unguarded access to `params` , which could be `undefined`. I suggest:
```js
const focus = params ? params.focus || [] : []; 
const switchToPath = params ? params.switchToPath : undefined;
```

src/octoprint/printer/profile.py
```diff
@@ -198,7 +198,6 @@ def values(cls):
 
 
 BedTypes = BedFormFactor
```
jacopotediosi: What about removing the alias entirely? It's not used anywhere in the core, nor by any plugin in the repository.

jacopotediosi (changes requested):

jacopotediosi:
In addition to the review above, I conducted a thorough research on all deprecations in the current state of this PR. Below are my findings.

### Leftovers:

- [x] Remove the documentation file [`docs/jsclientlib/logs.rst`](https://github.com/OctoPrint/OctoPrint/blob/fix/remove-long-deprecated-stuff/docs/jsclientlib/logs.rst). `OctoPrintClient.logs` has already been removed in this PR by commit fdf41b76417f7ccac29751f50c2c29b8217e2dbf

- [x] Remove `usersViewModel` from the viewmodels documented in https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/docs/plugins/viewmodels.rst?plain=1#L146-L147. It was removed in this PR by commit c4a6be0baef2bf84b40704ee45a2a9e89274fdf1.

- [x] Remove `admin_permission` and `user_permission`:
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/server/__init__.py#L116-L126
Fix `timelapse.py` which still uses `admin_permission`, replacing it with `Permissions.ADMIN.can()`:
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/server/api/timelapse.py#L336

- [x] Remove the plugin hook `octoprint.users.factory`:
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/server/__init__.py#L975-L979
And the following reference in a docstring:
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/cli/user.py#L28-L35


### Additional suggestions:

- [x] Add `since=2.0.0` to the deprecation of the `blacklisted` plugin property, to improve traceability:
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/plugin/core.py#L828

- [x] Consider removing the `Settings._config` property, both setter and getter. It has been deprecated for 4 years, was never documented, and a warning is logged when used. The core codebase already does not use it anywhere. If opting not to remove it, at least add a `TODO` to plan its removal.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/settings/__init__.py#L1031-L1045

- [x] Consider removing `--daemon` from the CLI (the new alternative is `octoprint daemon`) and server start via `octoprint` (the new alternative is `octoprint serve`). These modes have been deprecated for over 10 years.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/cli/__init__.py#L374-L381
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/cli/__init__.py#L392-L395

- [x] The function that redirects from `/api/users` to `/api/access/users` is called `deprecated_get_users`, but it does not actually log any deprecation warning. I suggest renaming the function to `getUsers`, consistent with the rest of the file. If all redirects in [`src/octoprint/server/api/users.py`](https://github.com/OctoPrint/OctoPrint/blob/fix/remove-long-deprecated-stuff/src/octoprint/server/api/users.py) are deprecated, I suggest adding log warnings and planning their future removal by adding a `TODO`.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/server/api/users.py#L13-L15

- [x] Consider removing the following deprecated access methods in `octoprint.plugin.PluginSettings`. They have been deprecated for 10 years, in favor of the snake_case version, and a warning is logged when used. If opting not to remove them, at least add a `TODO` to plan their removal.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/plugin/__init__.py#L479-L486

- [x] Consider removing the function `octoprint.util.dict_clean()`. It has been deprecated for 10 years, in favor of `dict_sanitize()`, and a warning is logged when used. It is not used by the core or any of the plugins in the repository. If opting not to remove it, at least add a `TODO` to plan its removal.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/util/__init__.py#L640-L643

- [x] Consider removing the function `octoprint.util.json.dump()`. It has been deprecated for 4 years, in favor of `JsonEncoding.dumps()`, and a warning is logged when used. It is not used by the core or any of the plugins in the repository. If opting not to remove it, at least add a `TODO` to plan its removal.
https://github.com/OctoPrint/OctoPrint/blob/7f399a4ba193fec75e3ddc95a006350596a0e40d/src/octoprint/util/json/__init__.py#L9-L13

src/octoprint/printer/profile.py
foosel: Did check the core but wasn't sure about the plugin situation, removing it :+1:

foosel:
> Consider removing the following deprecated access methods in octoprint.plugin.PluginSettings

Decided against removing them last week, still stand by this, but now added a TODO for 3.0.0. Plan is to send the affected plugins (MANY) some PRs in the meantime.

> Remove the plugin hook octoprint.users.factory

While I could only find one plugin that uses it, it so far never logged a deprecation warning. So instead of removing it, I've now added a deprecation warning (and a "remove me" TODO).

---

After a quick grep through the plugin sources, I'm seriously worried about getting rid of `admin_permission` and `user_permission` btw. But those things have been warning about deprecation for ages now, so let's hope we can fix the broken stuff out there with some PRs. If not, this particular change can still be rolled back.

foosel:
Totally forgot to re-request your review, now done. Sorry for the noise.

src/octoprint/cli/user.py
jacopotediosi: `octoprint.users.factory` -> `octoprint.access.users.factory`

src/octoprint/server/__init__.py
```diff
@@ -110,23 +110,9 @@ def __init__(self):
...
-
-# we set user_permission to a GroupPermission with the default user group
-user_permission = octoprint.util.variable_deprecated(
-    "user_permission has been deprecated, please use individual Permissions instead",
-    since="1.4.0",
-)(groups.GroupPermission(groups.USER_GROUP))
-
 import octoprint._version  # noqa: E402
-import octoprint.access.groups as groups  # noqa: E402
+import octoprint.access.groups as groups  # noqa: E402  # noqa: E402
```
jacopotediosi: Duplicated `# noqa: E402`

src/octoprint/cli/__init__.py
jacopotediosi: This could be removed

jacopotediosi (changes requested):
There are just these three leftovers and then for me this PR can go

jacopotediosi (approved):

foosel:
Contrary to my usual approach, I have merged this one without squashing it first. I feel like the context provided by the individual commits will be quite important down the road.

PR #5354, opened 2026-04-01
fix(settings): add serial compat overlay

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

This PR adds a compat overlay to the `["serial"]` global settings path, so that any plugins that were accessing that path can still do so, causing a deprecation message to appear in the logs and giving them some time to migrate to the new paths instead of crashing or getting always `None`.

This is necessary, since several plugins were accessing that global path. To my knowledge, at least the following: `bettergrblsupport, Chituboard, FlashForge, klipper, meatpack, portlister, portretry, siocontrol, telegram`.

Please note that this PR also includes a fix for path update callbacks: when a path update callback was registered, it was not called if a subpath was updated. Without this fix, it would have been necessary, in the serial compat overlay, to subscribe callbacks for all subpaths of `["plugins", "serial_connector"]`, one by one. I believe it is instead a desired behavior that when a child is modified, the update callback is also called for all ancestors.

This PR also includes new unit tests for the serial compat overlay and for the webcam compat overlay. They can be run with the following command:

```bash
python -m pytest tests/settings/test_settings.py
```

No.

### DIFF
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -562,8 +562,8 @@ def init_serial_compat_overlay(settings):
             settings.remove_overlay("serial_compat")
             return
 
-        send_checksum = serial_connector.get("sendChecksum")
-        error_handling = serial_connector.get("errorHandling")
+        send_checksum = serial_connector.pop("sendChecksum", None)
+        error_handling = serial_connector.pop("errorHandling", None)
 
         serial_compat = {
             # Connection params (moved to printerConnection.*)
--- a/src/octoprint/__init__.py
+++ b/src/octoprint/__init__.py
@@ -551,6 +551,78 @@ def init_settings_plugin_config_migration_and_cleanup(plugin_manager):
     ]
 
 
+def init_serial_compat_overlay(settings):
+    import logging
+
+    logger = logging.getLogger(__name__)
+
+    def set_overlay():
+        serial_connector = settings.get(["plugins", "serial_connector"])
+        if serial_connector is None:
+            settings.remove_overlay("serial_compat")
+            return
+
+        send_checksum = serial_connector.get("sendChecksum")
+        error_handling = serial_connector.get("errorHandling")
+
+        serial_compat = {
+            # Connection params (moved to printerConnection.*)
+            "port": settings.get(
+                ["printerConnection", "preferred", "parameters", "port"]
+            ),
+            "baudrate": settings.get(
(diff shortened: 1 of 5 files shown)


foosel:
> Please note that this PR also includes a fix for path update callbacks: when a path update callback was registered, it was not called if a subpath was updated. Without this fix, it would have been necessary, in the serial compat overlay, to subscribe callbacks for all subpaths of `["plugins", "serial_connector"]`, one by one. I believe it is instead a desired behavior that when a child is modified, the update callback is also called for all ancestors.

I agree!

src/octoprint/__init__.py
```diff
@@ -551,6 +551,78 @@ def settings_plugin_config_migration_and_cleanup(identifier, implementation):
...
+            "disconnectOnErrors": error_handling == "disconnect",
+            "ignoreErrorsFromFirmware": error_handling == "ignore",
+            # Old names (blacklisted -> blocklisted)
+            "blacklistedPorts": serial_connector.get("blocklistedPorts"),
+            "blacklistedBaudrates": serial_connector.get("blocklistedBaudrates"),
+        }
+
+        for key, value in serial_connector.items():
+            if key not in serial_compat:
+                serial_compat[key] = value
```
foosel, line 597: Unless I'm mistaken (and I'm tired, so that's definitely an option here), this will also copy over stuff that doesn't belong in the old settings, e.g. `sendChecksum` and `errorHandling`. `pop`ing those above should solve this.

foosel (commented):
Only one issue (that might not even be one)

jacopotediosi:
Fixed your observation by`pop`ing `sendChecksum` and `errorHandling` before the loop.

Please note that any future keys added to the serial connector schema that don't belong in the old serial path would have the same issue.

Should we change our approach and build the compat overlay by explicitly listing all the keys that used to exist under the serial settings path instead?

I don't think that's necessary, since there's no reason for a plugin to try to retrieve from the old path settings that only exist in the new one.

foosel:
> Should we change our approach and build the compat overlay by explicitly listing all the keys that used to exist under the serial settings path instead?
> 
> I don't think that's necessary, since there's no reason for a plugin to try to retrieve from the old path settings that only exist in the new one.

Nah, I don't think it's necessary either. An additional key should usually not be a concern, I just didn't want to data duplication basically and start with only what should be in there.

PR #5355, opened 2026-04-09
fix(auth): fix workflows for users without a set password (e.g. created via remote header)

foosel:

Two things:
- fixes the credential recheck workflow for users without password (no recheck should be asked for in that case)
- fixes the workflow to set a new password for users without a password via the user settings (no current password input shown and also not checked on the backend)

- set reauthentication timeout to 1m
- created a remote user via header without password through reverse proxy
- verified that after two minutes I was not asked for reauthentication when trying to open the plugin repo
- set a password through the user settings
- verified that after two minutes I now WAS asked for reauthentication when trying to open the plugin repo

Also checked the response on `/api/login` with and without a set password (`has_password` and `_credentials_recheck_supported` changing accordingly).

No

This also now makes it possible to request reauthentication checks from users coming in through an auth proxy, as long as they also set a password locally. For now I decided against forcing that.

Related to #5321


foosel:
As this is security related (and I also remember you asking for that specific feature this enables), officially requesting a review from you @jacopotediosi ^^

src/octoprint/server/api/access.py
```diff
@@ -263,18 +263,25 @@ def change_password_for_user(username):
     ):
         data = request.get_json()
 
-        if "password" not in data or not data["password"]:
+        new_password = data.get("password")
+        if new_password is None:
```
jacopotediosi: You changed the check from `not data["password"]` to `new_password is None`, it means that an empty password (`""`) will be accepted (only via direct API calls, as in frontend we check that inputs are filled) and I think this is not what we want 😄

src/octoprint/static/js/recovery/recovery.js
```diff
@@ -171,6 +175,7 @@ $(function () {
         };
```
jacopotediosi: In `loginstate.js` file, you added the `self.credentialsRecheckSupported()` guard within `self.forceReauthentication()`. I don't understand why you didn't do that here too. There may be a reason, but I can't figure it out.

src/octoprint/server/api/access.py
```diff
@@ -263,18 +263,25 @@ def change_password_for_user(username):
...
-                abort(403, description="Invalid current password")
-
-        elif current_user.has_permission(Permissions.ADMIN):
-            ensure_credentials_checked_recently()
+        if current_user.has_permission(Permissions.ADMIN):
+            if not credentials_checked_recently():
+                if current_password is None:
+                    abort(403, description="Please reauthenticate with your credentials")
+                if not userManager.check_password(username, current_password):
+                    abort(403, description="Invalid current password")
```
jacopotediosi: I found that after this change, an admin can change their own password in OctoPrint Settings -> Access Control even entering an incorrect current password, as long as the credentials have been checked recently. Something is off in this logic.

jacopotediosi (changes requested):

src/octoprint/server/api/access.py
foosel: Now that you write that, yes, indeed, that's a bad idea. The thought process while coding was "but this is string and an empty string is still a string, this needs to be a None check" 🙈

src/octoprint/server/api/access.py
foosel: The problem we have here is that an admin can change everyone's password. So the thoughtprocess here was they should also be able to change their own the same way.

On the other hand, it's probably cleaner to simply disallow changing the password of the current (admin) account through Settings > Access Control and instead force the user to do this via their own user settings (and then force the password check again).

src/octoprint/static/js/recovery/recovery.js
foosel: `self.forceReauthentication` will only be called by `self.reauthenticateIfNecessary`, and only if `self.checkCredentialsSeen` returns `false`, which does check against `self.credentialsRecheckSupported` and returns `true` if it it's unsupported.

In the other case, I had to assume that maybe some other viewmodel might also be calling `self.forceReauthentication` (third party plugins), but the recovery page is isolated and thus I don't have to cover for that.

However, I'll just pull in the `self.forceReauthentication` code into `self.reauthenticateIfNecessary` to make this explicit.

foosel:
Feedback implemented, please take another look :)

src/octoprint/static/js/recovery/recovery.js
jacopotediosi, line 83: Missing `self.credentialsRecheckSupported(resp._credentials_recheck_supported);` before `self.credentialsSeen(resp._credentials_seen);`.
Without this, `restoreBackup` and `reauthenticateDownload` never trigger the reauth dialog if credentials expire while on the recovery page.

src/octoprint/templates/snippets/settings/accesscontrol/users.jinja2
jacopotediosi, line 200: This is dead code, given that now the `Current Password` input is always hidden

jacopotediosi (changes requested):

jacopotediosi:
Apart from the above, the rest looks fine to me

foosel:
Thanks for reminding me that the current password field was actually there on the settings version of that panel, I was 100% sure it wasn't and that this whole workflow was broken from the get go. That made me think about things again, and realize that there was actually still a bug in there for a long time ([7bc9e95](https://github.com/OctoPrint/OctoPrint/pull/5355/commits/7bc9e95c70237227fceaf42fed05a6bb125ab4f0)). That's now fixed, changing the password of the current user via the settings has been brought back, and the logic on the server side makes sure that if you change your own password, your current password is required, regardless of your permissions.

Now things should be fine I hope...

src/octoprint/static/js/recovery/recovery.js
foosel: Fixed.

src/octoprint/templates/snippets/settings/accesscontrol/users.jinja2
foosel: Reverted the removal of setting your password via the settings after all, so that's no longer dead code.

src/octoprint/static/js/recovery/recovery.js
```diff
@@ -78,6 +79,7 @@ $(function () {
             OctoPrint.browser.passiveLogin().done((resp) => {
                 self.username(resp.name);
                 self.permitted(_.includes(resp.needs.role, "admin"));
+                self.credentialsRecheckSupported(response._credentials_recheck_supported);
```
jacopotediosi: This should be `resp`, not `response`

jacopotediosi (changes requested):

src/octoprint/static/js/recovery/recovery.js
foosel: Not my day...

jacopotediosi (approved):

PR #5386, opened 2026-05-04
fix: fix template permission checks

jacopotediosi:
  Brainstorming ticket
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

This PR fixes a bug [reported](https://discord.com/channels/704958479194128507/705047010641838211/1500664412371943475) by a user on Discord: users with the `CONTROL` permission who don't belong to the `user` group could not send commands through the terminal tab because the input was disabled.

This happened because in `terminal.jinja2`, `loginState.isUser()` was erroneously checked instead of `loginState.hasPermissionKo(access.permissions.CONTROL)`.

I took the opportunity to check whether the same mistake had been made elsewhere and to review permission checks across all templates in general, and I found the following additional issues:
- In `webcam.jinja2` and `timelapse.jinja2`, `loginState.isAdmin()` was checked instead of `loginState.hasPermissionKo(access.permissions.SETTINGS)()` to decide which error message to show based on whether the user could change the settings. Additionally, `loginState.isAdmin` and `loginState.isUser` were used without parentheses, so they were always truthy.
- In `access.js`, `loginState.hasPermissionKo()` was used instead of `loginState.hasPermission()`, which returns a Knockout pureComputed (a function) and therefore always evaluated to true.
- In `files.jinja2`, `loginState.hasPermissionKo()` was called without the trailing parentheses `()`, always evaluating to true inside `&&` chains.

This PR fixes all these issues.

Use the terminal tab with a user who has the `CONTROL` permission but is not in the `user` group.

No

### DIFF
--- a/src/octoprint/plugins/classicwebcam/templates/classicwebcam_webcam.jinja2
+++ b/src/octoprint/plugins/classicwebcam/templates/classicwebcam_webcam.jinja2
@@ -62,15 +62,15 @@
                 <p>
                     <strong>{{ _("Webcam stream not loaded") }}</strong>
                 </p>
-                <p data-bind="visible: !loginState.isAdmin()">
+                <p data-bind="visible: !loginState.hasPermissionKo(access.permissions.SETTINGS)()">
                     <small>{{ _("It might not be configured correctly or require authentication. To have this fixed, get in touch with an administrator of this OctoPrint instance.") }}</small>
                 </p>
-                <p data-bind="visible: loginState.isAdmin">
+                <p data-bind="visible: loginState.hasPermissionKo(access.permissions.SETTINGS)">
                     <small>{% trans %}
                         It might not be configured correctly or require authentication. You can change the URL of the stream under "Settings" > "Classic Webcam" > "Stream URL". If you don't have a webcam you can also just disable webcam support there.
                     {% endtrans %}</small>
                 </p>
-                <p data-bind="visible: loginState.isUser">
+                <p>
                     <small>{{ _("Currently configured stream URL") }}: <a target="_blank"
     rel="noreferrer"
     data-bind="attr: {href: settings.streamUrl}, text: settings.streamUrl"></a></small>
--- a/src/octoprint/plugins/classicwebcam/templates/classicwebcam_webcam.jinja2
+++ b/src/octoprint/plugins/classicwebcam/templates/classicwebcam_webcam.jinja2
@@ -70,7 +70,7 @@
                         It might not be configured correctly or require authentication. You can change the URL of the stream under "Settings" > "Classic Webcam" > "Stream URL". If you don't have a webcam you can also just disable webcam support there.
                     {% endtrans %}</small>
                 </p>
-                <p>
+                <p data-bind="visible: loginState.loggedIn">
                     <small>{{ _("Currently configured stream URL") }}: <a target="_blank"
     rel="noreferrer"
     data-bind="attr: {href: settings.streamUrl}, text: settings.streamUrl"></a></small>
--- a/src/octoprint/templates/tabs/timelapse.jinja2
+++ b/src/octoprint/templates/tabs/timelapse.jinja2
@@ -142,13 +142,13 @@
         <p>
             <strong>{{ _("Timelapse not fully configured") }}</strong>
         </p>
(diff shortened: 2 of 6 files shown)


jacopotediosi:
Force pushed because e2e tests detected an error (I forgot to import `accessViewModel` in `classicwebcam.js`), should be okay now

foosel:
> * Additionally, `loginState.isAdmin` and `loginState.isUser` were used without parentheses, so they were always truthy.

Hm, I'm confused by this statement. From what I can see in the PR, these occurances where limited to evaluations in `data-bind` statements, and both are observables. Something like `visible: observable` will evaluate the observable value without `()`. `visible: !observable()` however needs them as knockout will create an on-the-fly computed observable from that.

Am I missing something here?

jacopotediosi:
> > * Additionally, `loginState.isAdmin` and `loginState.isUser` were used without parentheses, so they were always truthy.
> 
> Hm, I'm confused by this statement. From what I can see in the PR, these occurances where limited to evaluations in `data-bind` statements, and both are observables. Something like `visible: observable` will evaluate the observable value without `()`. `visible: !observable()` however needs them as knockout will create an on-the-fly computed observable from that.
> 
> Am I missing something here?

Maybe you are right. If remember correctly in some cases Knockout automatically unwraps even when without parentheses, but I always get confused. I need to check at runtime if this was the case.

foosel:
For the record, you make use of this mechanism yourself in this PR :D

``` html
<p data-bind="visible: loginState.hasPermissionKo(access.permissions.SETTINGS)">
```

Note the missing parentheses. Works, because `hasPermissionKo` generates an observable that can be used directly here. If this needs to be negated however, you need the `()`.

src/octoprint/plugins/classicwebcam/templates/classicwebcam_webcam.jinja2
```diff
@@ -62,15 +62,15 @@
...
                     <small>{{ _("It might not be configured correctly or require authentication. To have this fixed, get in touch with an administrator of this OctoPrint instance.") }}</small>
                 </p>
-                <p data-bind="visible: loginState.isAdmin">
+                <p data-bind="visible: loginState.hasPermissionKo(access.permissions.SETTINGS)">
                     <small>{% trans %}
                         It might not be configured correctly or require authentication. You can change the URL of the stream under "Settings" > "Classic Webcam" > "Stream URL". If you don't have a webcam you can also just disable webcam support there.
                     {% endtrans %}</small>
                 </p>
-                <p data-bind="visible: loginState.isUser">
+                <p>
```
foosel:
```suggestion
                <p data-bind="visible: loginState.loggedIn">
```

src/octoprint/templates/tabs/timelapse.jinja2
```diff
@@ -142,13 +142,13 @@
...
-        <p data-bind="visible: !loginState.isAdmin()">
+        <p data-bind="visible: !loginState.hasPermissionKo(access.permissions.SETTINGS)()">
             {{ _("A working snapshot setup and/or the path to FFMPEG are missing. To have this fixed, get in touch with an administrator of this OctoPrint instance.") }}
         </p>
-        <p data-bind="visible: loginState.isAdmin">
+        <p data-bind="visible: loginState.hasPermissionKo(access.permissions.SETTINGS)">
             {{ _("A working snapshot setup and/or the path to FFMPEG are missing. You can change both under \"Settings\" > \"Webcam & Timelapse\" and in the settings of your chosen webcam plugin. If you don't have a webcam or don't want to enable timelapse support you can also just disable it there.") }}
         </p>
-        <p data-bind="visible: loginState.isUser">
+        <p>
```
foosel:
```suggestion
        <p data-bind="visible: loginState.loggedIn">
```

foosel (changes requested):
Two minor changes... I think it makes sense to still hide the current config info of webcam & timelapse stuff if viewed by someone who's not at least logged in. That's a milder approach then `isUser` or a specific permission, but still makes sure that some rando passing by an OctoPrint instance that for some reason allows anonymous viewing doesn't get that kind of info.

foosel (approved):

Issue #5393, opened 2026-05-14
[2.0.0rc1] bug: unable to move files from file list in local storage


jneilliii:
### The problem

You are unable to move a file from root of local storage to sub-folder in 2.0.0rc1 in the file list, but works in upload manager. After some digging it looks like the API call from the file list is not using the same destination path. 

borken API call
<img width="330" height="118" alt="Image" src="https://github.com/user-attachments/assets/1b8db12f-1b16-4199-b82a-1598e9258492" />

working API call from upload manager
<img width="389" height="120" alt="Image" src="https://github.com/user-attachments/assets/d9b61987-8ddb-4440-b3d9-9e53edea840e" />

### Proposed solution

sanitize the destination path in the file list move command to match upload manager's? 

### Additional information

brought up by @paukstelis in Discord #dev-core channel.

### Verification

- [x] I'm a maintainer or collaborator of OctoPrint and allowed to use this issue type

foosel:
Fixed by PR #5399

PR #5399, opened 2026-05-16
fix(coreui): fix unable to move from root

jneilliii:
- [X] You have read through `CONTRIBUTING.md`
- [X] Your changes are not possible to do through a plugin and relevant
  Brainstorming ticket
- [ ] Your PR targets OctoPrint's `dev` branch
- [X] Your PR was opened from a custom branch on your repository
- [X] Your PR only contains relevant changes: no unrelated files,
  if necessary!
  with `lessc`)
- [X] You have tested your changes (please state how!) - ideally you
  have added unit tests
- [ ] You have run the existing unit tests against your changes and
  nothing broke (`pytest`)
- [X] You have run the included `pre-commit` suite against your changes
- [X] You have added yourself to the `AUTHORS.md` file :)

appends trailing slash to folder list in copy move dialog of file list to fix erroring API calls

manually tested by making the changes in locally and moving files from root to sub-folder.

No

#5393

### DIFF
--- a/src/octoprint/server/api/files.py
+++ b/src/octoprint/server/api/files.py
@@ -1352,6 +1352,11 @@ def gcodeFileCommand(storage, path=""):
                 sanitized_source = fileManager.path_in_storage(storage, path)
 
                 destination = data["destination"]
+                if fileManager.folder_exists(
+                    dst_storage, destination
+                ) and not destination.endswith("/"):
+                    destination += "/"  # ensure target folders end on /, see #5393
+
                 dst_path, dst_name = fileManager.split_path(dst_storage, destination)
                 if not dst_name:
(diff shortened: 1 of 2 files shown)


foosel:
Hm, I fear the issue is a bit more difficult than this. The underlying problem is that if the `/` is missing on a destination folder, the actual destination for the file will be the existing targeted folder, vs a file inside said folder. There's code there that's supposed to solve this, but it isn't working.

So while this PR here is a solution of the symptom, it's not fixing the actual problem. And I'd also prefer that the API isn't that problematic to use that something like this will cause problems for clients.

I'll extend this PR with the related fix on the backend, once I've figured it out, then merge.

PR #5413, opened 2026-06-06
fix(pmgr): remove duplicated hide_abandoned filter

jacopotediosi:
  Brainstorming ticket
- [ ] Your PR targets OctoPrint's `dev` branch
  if necessary!
  with `lessc`)
  have added unit tests
  nothing broke (`pytest`)

This PR removes the `hide_abandoned` filter, introduced by commit 31cc659497de07cfd5ac6a8381aa21f96fa0ca84, because it's a duplicate of the `filter_abandoned` filter which was already there:

image

No.

### DIFF
--- a/src/octoprint/plugins/pluginmanager/static/js/pluginmanager.js
+++ b/src/octoprint/plugins/pluginmanager/static/js/pluginmanager.js
@@ -138,9 +138,6 @@ $(function () {
                         plugin.attributes && plugin.attributes.includes("ai-developed")
                     );
                 },
-                hide_abandoned: (plugin) => {
-                    return !plugin.abandoned;
-                },
                 filter_installed: function (plugin) {
                     return !self.installed(plugin);
                 },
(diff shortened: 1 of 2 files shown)


foosel:
Hmmm... "Hide abandoned" feels a bit more intuitive than "Only show unabandoned". So yes, while this was doubled (no clue why I didn't see it, I did look for it because I remembered it), maybe it would make more sense to switch to the new one instead of sticking with the old one? What do you think?

jacopotediosi:
There could be various ways to interpret the logical difference between the second and third section, and the reason why they are separated. I don't have a precise idea about this.

To be honest, I find it slightly confusing also that some options are "show" and others are "hide".

"Only show uninstalled plugins" also seems weird to me, shouldn't it be "not installed"? Doesn't "uninstalled" imply that the plugin was previously installed and then removed?

jacopotediosi:
Maybe the second section is about plugin tags and the third one is about things that are instance specific (compatible, already installed)? If so, then "abandoned" should be in the second section because it's a tag.

foosel:
You raise a good point there, maybe I need to rethink those sessions (and their wordings) altogether. For now I'll merge this, but make a note to take another look at this menu.

PR #5430, opened 2026-07-07
fix(api): add aggressive path restrictions on settings api

### LINKED ISSUE #5429: Limit `SETTINGS_READ` settings paths to frontend relevant settings on the settings API endpoint
### The problem

There are several settings available on the settings API endpoint for users that have the `SETTINGS_READ` permission that are backend-only settings and that are such irrelevant to the frontend, unless the user also has the necessary permissions to set them.

Therefore, they should not be returned on the API either.

This especially concerns local paths to (re)start commands, `ffmpeg` and configured folders. For the sake of offering as little of information about the underlying system as possible, such settings should be redacted from the API response, unless the user has the related permissions to change them.

### Proposed solution

Add path permissions for settings paths irrelevant for the frontend's function, those should only be available with write permissions through the `SETTINGS` permission.

Currently identified are:

- `devel.*`
- `feature.pollWatched`
- `folder.*`
- `gcodeAnalysis.runAt`
- `server.*`
- `system.*`
- `scripts.*`
- `webcam.[bitrate|ffmpegCommandline|ffmpegPath|ffmpegVideoCodec|ffmpegThreads|watermark]`

We can use the same approach to limit `controls.*` to `CONTROL` permission, instead of keeping it as a custom check.

We should keep the leaves of these settings, and just remove the actual data, to not cause issues with existing third party clients.

### Additional information

Thanks to @MKLKCDG for the heads-up.

### Verification

- [x] I'm a maintainer or collaborator of OctoPrint and allowed to use this issue type

foosel:

Provide backend-only settings only to users with SETTINGS
permission.

Unit tests were added for the path restriction function that's now also used by the settings endpoint itself. E2E tests are green. Checks with admin & user account showed no errors in the server logs or the browser logs.

Hahaha... no

-

Closes #5429

-

#### Further notes

-


foosel:
@jacopotediosi would love a review from you

src/octoprint/server/api/settings.py
```diff
@@ -851,8 +871,6 @@ def _saveSettings(data):
     if "system" in data:
         if "actions" in data["system"]:
             s.set(["system", "actions"], data["system"]["actions"])
-        if "events" in data["system"] and not api_version_matches(">=2.0.0"):
-            s.set(["events"], data["system"]["events"])
```
jacopotediosi, line 855: Why this removal? Doesn't it break backward compatibility?

src/octoprint/server/api/settings.py
```diff
@@ -345,21 +374,12 @@ def getSettings():
...
-            "autologinLocal": s.getBoolean(["accessControl", "autologinLocal"]),
-            "autologinHeadsupAcknowledged": s.getBoolean(
-                ["accessControl", "autologinHeadsupAcknowledged"]
-            ),
-            "defaultReauthenticationTimeout": s.getInt(
-                ["accessControl", "defaultReauthenticationTimeout"]
-            ),
-        }
+    if not Permissions.ADMIN.can() or not credentials_checked_recently():
+        data.get("api", {})["key"] = None
```
jacopotediosi, line 385: This way the `accessControl.defaultReauthenticationTimeout` field disappears from the response, even for admins, but I believe it's still used by some frontend elements, e.g. `snippets/settings/server/serverReauthenticationTimeout.jinja2`

src/octoprint/plugins/serial_connector/__init__.py
```diff
@@ -63,6 +63,15 @@ def get_settings_defaults(self):
...
 
+    def get_settings_restricted_paths(self):
+        return {
+            "admin": [
+                [key]
+                for key in self.get_settings_defaults().keys()
+                if key not in ("log", "ignoreEmptyPorts")
+            ]
+        }
+
```
jacopotediosi, line 75: This way the `plugins.serial_connector` field does effectively disappear from the response for non-admins, but the same information remains accessible in the legacy `serial` block. Should we restrict that one too?

jacopotediosi (changes requested):
Below are some potential issues I found while comparing the Settings API responses before and after this PR.

Additionally, I'm not sure that restricting these fields to `admin` only via `get_settings_restricted_paths` is the right approach. Should we rely on granular permissions instead? For example, if a user holds the `CONTROL` permission but isn't an admin, why shouldn't they see the serial settings? Maybe I'm missing something here.

src/octoprint/server/api/settings.py
foosel: That stupid line needs a comment to not remove it... I fell over it while looking at these things and couldn't find any references to it in the code, and that confused me.

src/octoprint/server/api/settings.py
foosel: Yeah, `defaultReauthenticationTimeout` should have been moved to the general settings model, and at some point it also was, but then I moved some stuff around trying to make things a bit better to read and things got wrong I guess as this line got lost. Thanks!

src/octoprint/plugins/serial_connector/__init__.py
foosel: Yes, and I think instead of `admin` it should also be the more granular `Permissions.SETTINGS`.

foosel:
Have implemented the review feedback in 6744fca, and added some basic settings API tests in 7e72470 that verify the restrictions. Final commit above was to fix the api key situation on CI.

If you could take a quick look today that would be great, I need to push this out with rc4 tomorrow morning the latest.

foosel:
And sorry, I forgot to respond to your question:

> why shouldn't they see the serial settings

Because the serial settings (and all of the now restricted settings) in question are server-side only. There's no reason I can see for a user who cannot change these settings (nor see them in the UI) to be able to see them on the API. Overkill? Maybe, but it should have been like that from the get go I think, and with 2.0.0 that's a chance to fix it.

I'd even have preferred to not even leave them as empty leaves in the settings API response, but I could image some third party issues caused by that. We'll see with rc4.

jacopotediosi (approved):
LGTM

jacopotediosi:
I would just add a couple of lines of documentation, possibly also in the breaking changes page.

foosel:
Merged after adding some docs, thank you!

Issue #5454, opened 2026-09-22
[2.0.0rc5] Wrong check inside local storage leads to the possibility of a very limited path traversal in download endpoint


foosel:
### The problem

1. manually create `uploads2` in OctoPrint's base folder
2. put a file "foo.gcode" inside
3. attempt to download said file through `http://127.0.0.1:5000/downloads/files/local/../uploads2/foo.gcode`

It will succeed, it shouldn't.

### Proposed solution

Improve check inside local storage's `sanitize_path`. Instead of

``` python
        if not path.startswith(self.basefolder):
            raise ValueError(f"path not contained in base folder: {path}")
```

go with 

``` python
        if path != self.basefolder and not path.startswith(self.basefolder + os.path.sep):
            raise ValueError(f"path not contained in base folder: {path}")
```

`path` must thus be either `self.basefolder`, or start with `self.basefolder` + `/`

### Additional information

Path traversal is restricted to sharing a prefix with the actual `uploads` folder. There's usually no such path available in OctoPrint's base folder and one such thing can also only get created through external means.

Thanks to @AdminZch0 for the bug report.

### Verification

- [x] I'm a maintainer or collaborator of OctoPrint and allowed to use this issue type

PR #5455, opened 2026-09-22
fix(storage): fix sanitize_path on local storage

foosel:

The prefix check was not aggressive enough, allowing limited path traversal to a sibling path whose name starts with the base folder's name.

Added a unit test, failed before patch, succeeded after patch.

Nope

Closes #5454


foosel:
@jacopotediosi would love a quick review here. Planning to push this out with 2.0.0rc6 next week.

